PR #30, opened 2026-02-15
Add macOS Shortcuts deep-link support for start/stop recording

filippohronsky:
## Description

This PR adds Apple macOS Shortcuts integration via `stenoai://` deep links, allowing recording to be controlled from Shortcuts and automation flows.

#### `app/main.js`
- Added protocol handling for:
  - `stenoai://record/start?name=...`
  - `stenoai://record/stop`
- Added macOS `open-url` handling with startup queueing so links work on cold and warm launch
- Added dev-safe protocol registration and `argv` fallback for deep-link delivery
- Added renderer-ready handshake to prevent startup race conditions
- Added idempotent shortcut behavior (`already recording` / `already stopped`)
- Added quiet-launch behavior and Dock activate recovery
- Added shortcut notifications (respecting existing notifications setting)

#### `app/index.html`
- Added shortcut IPC listeners:
  - `shortcut-start-recording` -> reuses existing `startBtn.click()`
  - `shortcut-stop-recording` -> reuses existing `stopBtn.click()`
- Added `shortcut-renderer-ready` handshake event

#### `app/package.json`
- Added `build.protocols` registration for the `stenoai` URL scheme

#### `README.md`
- Added `macOS Shortcuts` section with setup and URL examples
- Added calendar automation documentation using **Rules – Calendar Automation** bridge
- Clearly documented that macOS Shortcuts cannot natively trigger on Calendar event start

### Why

Enable reliable macOS Shortcut and calendar-driven automation control for recording, without changing existing recording business logic.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- [x] Tested on macOS

### Manual test scenarios

1. Warm launch deep link start:
   - `stenoai://record/start?name=Daily%20Standup`
2. Warm launch deep link stop:
   - `stenoai://record/stop`
3. Cold launch via deep link with hidden-window behavior
4. Idempotent responses for repeated start/stop triggers
5. Existing `Cmd+Shift+R` global hotkey unchanged
6. Re-open from Dock after background shortcut start and verify UI shows active recording

## Additional Notes

- Recording start/stop backend logic was not modified; shortcuts are routed into existing renderer button handlers.
- No Python backend changes.
- `app/package-lock.json` intentionally not modified.
- I have read the CLA Document and I hereby sign the CLA

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,79 @@
 - **System audio capture** - Record mic + system audio simultaneously for virtual meetings with headphones
 - **macOS desktop app** with intuitive interface
 
+## macOS Shortcuts (Optional)
+
+<details>
+<summary>Expand setup and calendar automation guide</summary>
+
+StenoAI supports Apple Shortcuts via deep links using the `stenoai://` URL scheme.
+
+- Start recording: `stenoai://record/start?name=Daily%20Standup`
+- Stop recording: `stenoai://record/stop`
+
+### How to set it up
+
+1. Open the **Shortcuts** app on macOS.
+2. Create a new shortcut (for example: "Start StenoAI Recording").
+3. Add the **Open URLs** action.
+4. Use one of the URLs above.
+5. (Optional) Add a keyboard shortcut from the shortcut settings.
+
+### Calendar event naming (optional)
+
+If you want calendar-based names, resolve the event title in your Shortcut workflow and pass it as the `name` query value in the start URL.
+
+Example:
+
+`stenoai://record/start?name=Weekly%20Product%20Sync`
+
+### Calendar event start automation (via Rules bridge)
+
+macOS Shortcuts **cannot natively trigger** exactly at Calendar event start.  
+To run this automatically on event timing, a third-party automation app is required.
+
+This addon uses:
+
+- **Apple Shortcuts**: builds the `stenoai://record/start?...` action.
+- **Rules – Calendar Automation**: watches Calendar events and triggers the shortcut.
+
+#### Architecture overview
+
+1. Rules App monitors upcoming Calendar events.
+2. Rules checks the event note/body for a marker keyword (for example `stenoai`).
+3. If matched, Rules runs a Shortcut.
+4. The Shortcut gets the next event title and opens:
+   - `stenoai://record/start?name={calendar_event_title}`
+5. StenoAI receives the URL and starts recording with that name.
+
+#### Step-by-step setup
+
+1. Install **Rules – Calendar Automation** on macOS.
+2. Create a Shortcut in Apple Shortcuts (example name: `StenoAI Start From Calendar Event`).
+3. In that Shortcut, add actions in this order:
+   - `Find Calendar Events` (limit to `1`, sorted by start date ascending, upcoming only)
+   - Extract the event title from the found event
+   - `URL Encode` the title
+   - `Open URLs` with:
+     - `stenoai://record/start?name=<encoded title>`
+4. Open Rules and create a calendar-trigger rule:
+   - Source: your target calendar(s)
+   - Trigger window: event start (or preferred offset)
+   - Condition: event note contains `stenoai`
+   - Action: run Shortcut `StenoAI Start From Calendar`
+5. In your Calendar event notes, add the word `stenoai` for meetings that should auto-start recording.
+6. Test with a near-future event:
+   - create event with `stenoai` in notes,
+   - wait for trigger,
+   - confirm StenoAI starts and uses the event title as session name.
+
+#### Notes
+
+- Without Rules (or another automation bridge), this cannot be fully event-driven from Calendar start time.
+- Keep using regular manual shortcuts (`Open URLs`) for non-automated scenarios.
+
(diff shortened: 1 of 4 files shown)


ruzin (dismissed):
Thanks for the contribution, Filip! This is a well-scoped feature — routing shortcuts into the existing button handlers is a clean approach, and the cold-launch queuing + renderer-ready handshake are nice touches.

A few things that need to be addressed before merge:

### Breaking: `safeStorage` import dropped

The PR's diff changes the `require('electron')` destructure and drops `safeStorage`, which is used elsewhere in `main.js`. This would break existing functionality. The fix is straightforward — just add `Notification` to the existing import line without removing anything.

### Subprocess calls instead of existing IPC handlers

`shouldShowShortcutNotifications()` and `isBackendRecording()` both shell out via `runPythonScript()` to call `simple_recorder.py`. But `main.js` already has IPC handlers for both of these (`get-notifications` at line ~1982 and `get-status` at line ~407). Spawning a new Python subprocess for each is slower and creates a second code path to maintain. These should reuse the existing internal logic directly (or at minimum call the same handler functions).

### No input sanitization on session name

The `name` query param from the URL is passed through to the renderer with only a `.trim()`. Since this is an externally-triggered input (any app can open a `stenoai://` URL), it should have basic validation — e.g., strip path separators, limit length, restrict to alphanumeric + common punctuation.

### Commit messages

The three commits all have `\tmodified: <filename>` as their headline, which looks like pasted `git status` output. Would recommend squashing into a single commit with a descriptive message on merge.

### Minor: README placement

The macOS Shortcuts section is quite detailed for an optional/advanced feature. Consider wrapping it in a `<details>` collapse or moving to a separate doc (e.g., `docs/shortcuts.md`) and linking from the README.

---

Overall the approach is solid — the queuing, idempotent behavior, and edge case handling show good attention to detail. Just needs the above fixes and it should be good to go.

ruzin (dismissed):
Thanks for addressing all the initial feedback, Filip — everything from round one looks good. A few new issues to address:

### Critical

**1. Missing `requestSingleInstanceLock()` — duplicate instance risk**

With deep-link support, a `stenoai://` URL can now cold-launch a second Electron instance while one is already running. Two instances would compete for the same backend, audio devices, and state files. This needs the standard Electron single-instance pattern:

```js
const gotTheLock = app.requestSingleInstanceLock();
if (!gotTheLock) {
  app.quit();
} else {
  app.on('second-instance', (event, argv) => {
    const url = argv.find(a => a.startsWith(`${SHORTCUT_PROTOCOL}://`));
    if (url) handleShortcutUrl(url);
    if (mainWindow) {
      if (mainWindow.isMinimized()) mainWindow.restore();
      mainWindow.focus();
    }
  });
}
```

**2. URL logging leaks meeting names**

`console.log` statements log the full URL including `?name=1-on-1%20with%20John%20HR%20Review`. For a privacy-focused app, meeting names shouldn't end up in system logs. Log with the query string stripped:

```js
console.log(`Received shortcut URL: ${parsed.protocol}//${parsed.hostname}${parsed.pathname}`);
```

### Medium

**3. `launchedByShortcut` flag never properly reset**

This flag is set to `true` on every `open-url` event, not just cold launches. If the app is already running and visible, a `stenoai://record/stop` URL will set it to `true`. If the window is later destroyed and recreated, the next `ready-to-show` will hide the window unexpectedly.

Fix: only set `launchedByShortcut = true` when the app is not yet ready (`!app.isReady()`), and reset it after the shortcut action is dispatched.

**4. `void handleShortcutUrl()` swallows rejections**

`handleShortcutUrl` is async but called with `void`, so any rejection becomes an unhandled promise rejection (can crash Electron in strict mode). Add `.catch()`:

```js
handleShortcutUrl(incomingUrl).catch(err => {
  console.error('Error handling shortcut URL:', err.message);
});
```

**5. Unbounded `shortcutQueue`**

If the renderer crashes or fails to load, every incoming shortcut URL pushes to the queue with no limit. A misconfigured automation loop could cause memory growth. Add a cap:

```js
function enqueueShortcutAction(action) {
  if (shortcutQueue.length >= 5) {
    console.warn('Shortcut queue overflow, dropping oldest action');
    shortcutQueue.shift();
  }
  shortcutQueue.push(action);
  flushShortcutQueue();
}
```

**6. `isBackendRecording()` fragile during cold start**

During a cold launch, `handleGetStatus()` is called before the backend may be fully ready. If it fails, `isBackendRecording()` returns `false`, so a "stop" shortcut sent before the backend is ready will be told "already stopped." Consider adding a brief retry or at minimum documenting this limitation.

### Low

- `console.log` used instead of `sendDebugLog` — shortcut events are invisible in the app's debug panel, inconsistent with the rest of the codebase.
- First regex in `sanitizeShortcutSessionName` (`/[/\\:*?"<>|]/g`) is redundant — those characters are already excluded by the Unicode allowlist on the next line.
- `ensureMainWindow()` has no `app.isReady()` guard — safe today but fragile if called from new code paths.

PR #60, opened 2026-02-25
Feature/auto generate meeting titles

c9482:
# PR: Auto-generate meeting titles from transcript content

**Branch:** `feature/auto-generate-meeting-titles`
**Files changed:** 3 files, +107 / -2 lines

## Problem

When a user starts a recording without providing a name, StenoAI assigns a random hash like `Meeting-A3F82B`. This name persists even after the AI has fully summarized the meeting content, making it hard to identify meetings in the sidebar.

## Solution

After summarization completes, if the meeting name is an auto-generated hash (matches `Meeting-XXXXXX`), the AI generates a short, descriptive title from the summary content and replaces the hash. User-provided names are never overwritten.

**Example:** `Meeting-A3F82B` → `Q4 Budget Planning Review`

### `src/summarizer.py`
- Added `generate_title()` method to `OllamaSummarizer`
- Uses a focused prompt that asks the model for a max 6-word title
- Cleans up model output (strips quotes, prefixes, enforces word limit)
- Supports multi-language title generation
- Works with all AI providers (local Ollama, remote Ollama, cloud APIs)
- 30-second timeout to avoid blocking the pipeline
- Returns `None` on failure (never breaks the pipeline)

### `simple_recorder.py`
- Added `import re` to top-level imports
- Added Step 2b in `process_recording()`: after summarization, checks if `session_name` matches the `Meeting-[A-Z0-9]{6}` pattern and generates a title if so
- The generated title replaces `session_name` before the summary JSON is saved
- Wrapped in try/except — failure falls back silently to the hash name

### `app/main.js`
- Updated meeting lookup in `processNextInQueue()` to fall back to matching by audio file basename when name-based matching fails (since the name may have changed)
- Updated the same pattern in the `start-recording-ui` handler's stdout listener
- This ensures the frontend receives the correct `meetingData` in the `processing-complete` event even when the title was renamed

## Behavior

| Scenario | Before | After |
|---|---|---|
| No name provided | `Meeting-A3F82B` forever | AI generates title after summarization |
| User provides a name | Name preserved | Name preserved (no change) |
| Title generation fails | N/A | Falls back to `Meeting-A3F82B` |
| Calendar integration names meeting | Calendar name preserved | Calendar name preserved (no change) |
| Empty/short transcript | N/A | Falls back to hash (title generation returns None) |

## Testing

- [ ] Record a meeting without providing a name → verify AI-generated title appears after processing
- [ ] Record a meeting with a custom name → verify custom name is preserved
- [ ] Record with calendar integration active → verify calendar name is preserved
- [ ] Test with poor audio / empty transcript → verify graceful fallback to hash
- [ ] Test with non-English language setting → verify title generated in correct language

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -5996,7 +5996,9 @@ Session started - waiting for activity...
                 log('❌ No summary file found for reprocessing');
                 return;
             }
-            
+
+            const regenerateTitle = confirm('Also regenerate the meeting title?');
+
             // Track which meeting is being processed
             processingMeeting = selectedMeeting;
             const meetingName = selectedMeeting.session_info?.name;
@@ -6015,7 +6017,7 @@ Session started - waiting for activity...
             
             try {
                 // Call main process to reprocess the meeting
-                const result = await ipcRenderer.invoke('reprocess-meeting', summaryFile);
+                const result = await ipcRenderer.invoke('reprocess-meeting', summaryFile, regenerateTitle);
                 
                 if (result.success) {
                     log('✅ Meeting reprocessed successfully');
--- a/app/main.js
+++ b/app/main.js
@@ -1,4 +1,8 @@
 const { app, BrowserWindow, ipcMain, dialog, shell, systemPreferences, globalShortcut, safeStorage, Tray, Menu, nativeImage, Notification } = require('electron');
+
+// Prevent EPIPE crashes when stdout/stderr pipe is broken (e.g. launching terminal closed)
+process.stdout?.on('error', () => {});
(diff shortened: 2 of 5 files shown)


c9482:
All feedback addressed in e412970:

Replaced unsafe allMeetings[0] fallback with audio file basename matching
Cleaned up package-lock.json peer dependency noise
Fixed Ollama timeout to use HTTP-level ollama.Client(timeout=30) instead of model runtime option
Added title regeneration prompt on reprocess (per my earlier comment)

PR #79, opened 2026-03-12
Feature/add autodetect language

JeroenR123:
## Description

Brief description of what this PR does and why it's needed.

## Type of Change

- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested on macOS

## Additional Notes

Used AI

### DIFF
--- a/src/transcriber.py
+++ b/src/transcriber.py
@@ -293,6 +293,7 @@ class WhisperTranscriber:
                 try:
                     detection_result, _ = self.model.auto_detect_language(media=str(converted_path))
                     if detection_result and len(detection_result) >= 1:
+                        # pywhispercpp returns (language_code, probability) on auto_detect_language.
                         detected_language = detection_result[0]
                         resolved_language = detected_language
                         if len(detection_result) >= 2:


JeroenR123:
> Great work! Here is some feedback:
> 
> **1. Rebase needed** — PR #78 (Dutch language) has been merged. Since this PR includes that same commit, please rebase onto `main` to drop the duplicate and resolve any conflicts.

Done, had a little fight with the rebase. 

> **2. Typo in README** — `"Automatically detect you language"` should be `"Automatically detect your language"`.

Done

> **3. Query language fallback** — In the `query` command, the final `if language == "auto": language = "en"` fallback means "Ask Steno" queries on auto-detected non-English meetings will respond in English if the JSON didn't store `output_language`. This could be surprising for users who recorded in e.g. Dutch and expect Dutch answers.

It should never be the case that there is no output_language. This could happen when an old version is used before, and the output_language is not added. Then, the best language to back up to would be en. If you see this different, please let me know. 

> **4. Frontend settings persistence** — The HTML dropdown gets the new `"auto"` option, but there don't appear to be any JS changes to handle saving/restoring `"auto"` as the selected value. Can you verify the settings load code properly selects `"auto"` in the dropdown on app restart?

Weird, it is persistent in the app, just tried it.

> **5. `auto_detect_language` return value** — The tuple is indexed positionally (`detection_result[0]`, `detection_result[1]`). A brief comment noting the expected return shape from pywhispercpp would help future readers, since a silent API change would fall through to the fallback without any obvious indication.

Done.

PR #93, opened 2026-04-22
feat: notes generation animation, inline title editing, and UI polish

WilliamDrewett:
> **Stacked on #92** — depends on the chat interface redesign. Once #92 merges, this branch will be rebased onto `main` so the diff only shows changes below.

## Summary
- **Notes generation scanner**: Granola-style rounded rect overlay that glides smoothly down the page as notes stream in, with "Analyzing transcript" → "Generating notes" label and spinner
- **Regenerate button**: Icon spins in place instead of amber background + "Processing..." text swap; notes clear immediately on confirm
- **Inline title editing**: Click directly on meeting title to edit (Notion/Granola style), saves on blur — replaces edit button + modal
- **Fix: title save for `.md` meetings**: `update-meeting` IPC handler was calling `JSON.parse()` on YAML frontmatter files, causing silent save failures; now branches on file extension
- **Fix: `####` heading rendering** in streaming markdown renderer
- **Fix: bold-only lines** (e.g. "No topics discussed") now render as plain text instead of bold
- **Search bar**: Background colour matches sidebar/chat
- **Settings button**: Height aligned to match New Note button

## Test plan
- [x] Open a meeting, click regenerate — notes clear immediately, icon spins, scanner glides down as text streams in, scanner disappears on completion
- [x] Click meeting title, edit text, click away — title saves and updates in sidebar
- [x] Click meeting title, press Escape — reverts to original
- [x] Check `####` headings render correctly in streamed notes
- [x] Settings button and New Note button appear the same height
- [x] Search bar background matches sidebar colour

---
## Summary by cubic
Adds a smooth notes-generation scanner on stop, inline title editing, streaming chat with a Stop button (Esc also cancels), and a live "Recording" sidebar entry. Renames the auto session prefix to "Note-" and improves chat reliability, title handling, scanner performance, and overall stability.

- **New Features**
  - Notes scanner on stop; glides as notes stream; regenerate clears instantly with spinner; header status removed.
  - Streaming chat with Stop/Esc; cancels in-flight queries and appends "Stopped".
  - Inline title editing (click to edit; saves on blur; Esc reverts).
  - Live "Recording" sidebar entry with pulsing dot; safe to navigate; replaced by "Processing…" then final note.
  - Auto session prefix updated to "Note-"; auto-title works for both "Meeting-" and "Note-".
  - Chat sessions persist to disk and load on startup.

- **Bug Fixes**
  - Recording: remove ghost "Recording" item when start fails.
  - Chat: prevent duplicate streaming bubble; guard provider usage-only chunks; prompt chips submit reliably; stays expanded after stop.
  - Queries: add `query-cancel` IPC; kill orphaned procs; clean up per-query "destroyed" listeners on normal close.
  - Title/YAML: fix `.md` saves; unescape YAML on read; reject/strip newlines; escape backslashes; revert inline title only if still editing that meeting.
  - Markdown/UI: fix `####` headings and bold-only lines; match transcript/chat panel styling; minimise icon; subtle dark-mode hover fills; sidebar to 250px with aligned toggle; align Settings/New Note heights; search bar background.
  - Scanner: batch position updates via `requestAnimationFrame`; clamp for short notes.
  - Config: use `sys.frozen` to detect bundled builds and resolve data dirs correctly.

<sup>Written for commit a37cc0cdff84a2ca3076453d33cbedd81914b4bd. Summary will update on new commits.</sup>

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1094,7 +1094,7 @@
         
         /* Sidebar with meetings list */
         .sidebar {
-            width: 320px;
+            width: 250px;
             background: var(--bg-primary);
             border-right: 1px solid var(--border-subtle);
             display: flex;
@@ -1130,7 +1130,7 @@
 
         .sidebar-toggle {
             position: fixed;
-            left: 319px;
+            left: 249px;
             top: 50%;
             transform: translateY(-50%);
             width: 24px;
--- a/app/index.html
+++ b/app/index.html
@@ -8435,6 +8435,8 @@ Session started - waiting for activity...
                     hideAiQueryPopup();
                 } else if (settingsVisible) {
                     toggleSettings();
+                } else if (activeQueryId) {
+                    askBarStop.click();
                 }
             }
         });
@@ -9133,10 +9135,7 @@ Session started - waiting for activity...
         // ── Streaming Summary Rendering ──
         let streamedMarkdown = '';
         let streamingActive = false;
-
-        function escapeHtml(str) {
-            return str.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
-        }
+        let scannerRafPending = false;
 
         function renderStreamingMarkdown(md) {
             const container = document.getElementById('streaming-output');
@@ -9186,7 +9185,16 @@ Session started - waiting for activity...
             const scanner = document.getElementById('generation-scanner');
             if (scanner) {
                 scanner.style.display = 'flex';
-                scanner.style.top = (container.scrollHeight - 14) + 'px';
+                if (!scannerRafPending) {
+                    scannerRafPending = true;
+                    requestAnimationFrame(() => {
+                        scannerRafPending = false;
+                        const s = document.getElementById('generation-scanner');
+                        if (s && s.style.display !== 'none') {
+                            s.style.top = Math.max(0, container.scrollHeight - 40) + 'px';
+                        }
+                    });
+                }
             }
 
             const detailPanel = document.getElementById('meeting-detail');
@@ -9195,6 +9203,7 @@ Session started - waiting for activity...
 
         function stopStreamingRenderer() {
             streamingActive = false;
+            scannerRafPending = false;
             const scanner = document.getElementById('generation-scanner');
             if (scanner) scanner.style.display = 'none';
         }
--- a/app/index.html
+++ b/app/index.html
@@ -155,10 +155,12 @@
             background: transparent;
             border: 1px solid var(--border-subtle);
             color: var(--text-muted);
-            padding: 8px 12px;
+            padding: 6px 10px;
             border-radius: 8px;
             cursor: pointer;
-            font-size: 16px;
+            display: flex;
+            align-items: center;
+            justify-content: center;
             transition: all 0.2s ease;
         }
 
@@ -1183,7 +1185,7 @@
         }
 
         .search-input {
-            background: var(--bg-surface);
+            background: var(--bg-elevated);
             border: 1px solid var(--border-subtle);
             color: var(--text-primary);
             padding: 7px 12px 7px 30px;
@@ -1197,12 +1199,15 @@
 
         .search-input:focus {
             border-color: var(--accent-primary);
-            background: var(--bg-elevated);
-            box-shadow: 0 0 0 3px var(--accent-glow);
+            box-shadow: 0 0 0 3px rgba(129, 140, 248, 0.18);
         }
 
         .search-input::placeholder {
             color: var(--text-muted);
+        }
+
+        [data-theme="light"] .search-input {
+            background: var(--bg-deep);
         }
         
         .sidebar-header p {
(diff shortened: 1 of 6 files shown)


ruzin:
Pushed a small UX tweak on top (commit 0184f57):

- **Dropped the top-center `● Recording... 0:16` indicator** across recording/paused/processing — the bottom controls pill already conveys state + elapsed time, so the header feels calmer.
- **Scanner now appears the moment Stop is pressed** with the "Analyzing transcript" label, instead of only surfacing on the first summary chunk. That removes the ~5–30s blank gap between Stop and the first token where nothing was happening on screen. When streaming kicks in, the label flips to "Generating notes" and the scanner glides down as usual.

Feel free to revert if you'd prefer a different approach — happy to iterate.

ruzin (commented):
Took a pass over the full PR (all 8 commits). A handful of things worth addressing before merge — nothing blocking, but a couple are worth a quick fix:

## Medium

### 1. YAML title escape only handles `"`
`app/main.js:1358` — when the user inline-edits a meeting title for an `.md` file, the YAML frontmatter update does:

```js
const escapedName = updates.name.replace(/"/g, '\\"');
updatedContent = updatedContent.replace(/^title:\s*".*"$/m, `title: "${escapedName}"`);
```

That's missing two cases:
- **Newlines** (`\n` / `\r`): the regex is line-anchored (`^...$/m`) so the replace silently no-ops if the user pastes a multi-line title. User sees "saved" state but the file is untouched.
- **Backslashes** (`\`): a title ending in `\` produces `"foo\"` which breaks the YAML double-quoted string, corrupting the frontmatter. Not a security issue (user's own file) but will eat their summary on next reload.

Minimal fix:
```js
if (/[\n\r]/.test(updates.name)) {
  return { success: false, error: 'Title cannot contain newlines' };
}
const escaped = updates.name.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
```

### 2. Scanner position read forces layout per streamed chunk
`app/index.html:9189` — every `summary-chunk` event does `container.innerHTML = html` then reads `container.scrollHeight` to reposition the scanner. That's a forced layout flush + style recalc per chunk (10+/sec on fast models), on top of the innerHTML parse. Batch the position update in a rAF:

```js
if (!scannerRafPending) {
  scannerRafPending = true;
  requestAnimationFrame(() => {
    scannerRafPending = false;
    const s = document.getElementById('generation-scanner');
    if (s) s.style.top = (container.scrollHeight - 14) + 'px';
  });
}
```

### 3. Two `escapeHtml` implementations
`app/index.html:5759` uses the `textContent` trick (correct for all entities). `app/index.html:9137` uses a manual regex and is missing `'` (the single quote). Not exploitable in the current callers but a maintenance smell. Consolidate to the `textContent` version and delete the regex one.

## Low

### 4. Scanner can sit off-screen on very short notes
`app/index.html:9189` positions the scanner at `scrollHeight - 14`. Clamp it so it stays inside the visible region:

```js
const top = Math.max(0, container.scrollHeight - 40);
scanner.style.top = top + 'px';
```

### 5. Esc to cancel an in-flight streaming query
When a query is streaming, Esc should cancel it (and optionally collapse the chat). Currently only the explicit Stop button works. Small tap-saver:

```js
document.addEventListener('keydown', (e) => {
  if (e.key === 'Escape' && activeQueryId) askBarStop.click();
});
```

## False positives I ruled out

- **"Event listener leak on meeting list re-render"** — `meetingsList.innerHTML = ...` destroys the old DOM nodes, so their listeners are GC'd with them. No leak.
- **"Multi-query race"** — `submitBarQuery` already guards with `if (activeQueryId) return;` at `app/index.html:7960`, so overlapping queries can't start.

Overall the streaming UX, inline title edit, and live sidebar entry all feel solid in testing. The YAML escape is the one I'd fix before merge; the rest are follow-up polish.

ruzin (commented):
Took a pass over the full PR (all 8 commits). A handful of things worth addressing before merge — nothing blocking, but a couple are worth a quick fix.

## Medium

### 1. YAML title escape only handles `"`
`app/main.js:1358` — when the user inline-edits a meeting title for an `.md` file, the YAML frontmatter update does:

```js
const escapedName = updates.name.replace(/"/g, '\\"');
updatedContent = updatedContent.replace(/^title:\s*".*"$/m, `title: "${escapedName}"`);
```

That's missing two cases:
- **Newlines** (`\n` / `\r`): the regex is line-anchored (`^...$/m`) so the replace silently no-ops if the user pastes a multi-line title. User sees 'saved' state but the file is untouched.
- **Backslashes** (`\\`): a title ending in `\\` produces `"foo\\"` which breaks the YAML double-quoted string, corrupting the frontmatter. Not a security issue (user's own file) but will eat their summary on next reload.

Minimal fix: reject newlines in input and escape both backslashes and quotes.

### 2. Scanner position read forces layout per streamed chunk
`app/index.html:9189` — every `summary-chunk` event does `container.innerHTML = html` then reads `container.scrollHeight` to reposition the scanner. That's a forced layout flush per chunk (10+/sec on fast models). Batch the position update in a `requestAnimationFrame`.

### 3. Two `escapeHtml` implementations
`app/index.html:5759` uses the `textContent` trick (correct for all entities). `app/index.html:9137` uses a manual regex and is missing the single quote. Not exploitable in current callers but a maintenance smell — consolidate to the textContent version.

## Low

### 4. Scanner can sit off-screen on very short notes
`app/index.html:9189` positions the scanner at `scrollHeight - 14`. Clamp it so it stays inside the visible region.

### 5. Esc to cancel an in-flight streaming query
When a query is streaming, Esc should cancel it. Currently only the explicit Stop button works. Small tap-saver.

## False positives I ruled out

- **'Event listener leak on meeting list re-render'** — `meetingsList.innerHTML = ...` destroys the old DOM nodes, so their listeners are GC'd with them. No leak.
- **'Multi-query race'** — `submitBarQuery` already guards with `if (activeQueryId) return;` at `app/index.html:7960`, so overlapping queries can't start.

Overall the streaming UX, inline title edit, and live sidebar entry all feel solid in testing. The YAML escape is the one I'd fix before merge; the rest are follow-up polish.

PR #95, opened 2026-04-26
feat(renderer): replace legacy HTML renderer with React+Vite

WilliamDrewett:
## Why this change

The previous renderer was a single 9,700-line HTML/JS file with no build tooling, no type checking, and no component abstraction. Adding new features meant editing a monolithic file where a typo anywhere could silently break the whole UI. This PR replaces it with a proper React+Vite workspace.

The transition is **flag-gated** — users toggle to the new renderer in Settings, and the old renderer stays untouched and available as a fallback. This means the PR can merge and ship without disrupting anyone currently using the app.

## What's in each commit

This PR is structured as 7 self-contained commits that build on each other. Each commit is reviewable in isolation:

1. **Build wiring** — Vite workspace, electron-builder CI config, feature flag in `main.js`, `settings-store.js`, and a frozen IPC contract doc that acts as the boundary between main process and renderer.

2. **Preload + IPC client** — `preload.js` exposes an explicit allow-list of IPC channels via `contextBridge` (nothing else leaks through). `lib/ipc.ts` wraps every channel in a typed TypeScript function so channel strings never appear in component code.

3. **Data layer (hooks)** — TanStack Query hooks over every IPC call. The renderer never calls `ipc.ts` directly; it always goes through a hook. This gives automatic caching, background-refetch, and consistent loading/error states across the whole UI. Notable: `useStreamingQuery` turns a push-event channel into an async iterator for the Ask Bar.

4. **Component library** — Reusable primitives (Button, Card, Chip, Dialog, Tabs, Input, Select, Switch, Tooltip, etc.) built on shadcn/ui + Radix, styled with Tailwind to the StenoAI brand tokens. A `/dev` sandbox route lets you inspect every component without real meeting data.

5. **All screens** — Home (meeting list + folder sidebar), MeetingDetail (Summary / Transcript / Insights / Calendar tabs), FolderDetail, Settings, Setup wizard. Transcript tab supports full-text search and word-level timestamp highlighting. Settings layout mirrors the legacy panel so existing users find everything in the same place.

6. **Live recording screen** — Replaces the minimal recording overlay with a full screen: live audio wave (PCM level pushed ~10×/sec from main process), ticking duration timer, live note-taking (Zustand draft store, submitted on stop), pause/resume controls, and a Processing screen shown during transcription. Also adds `title-regeneration` and `model-install-status` backend commands used by the settings model picker.

7. **Ask Bar** — Floating AI chat panel anchored to the bottom of any meeting screen. Streams token-by-token via IPC push events. Conversation history is persisted per meeting in `chat_sessions.json` and survives restarts. Automatically pre-seeds context with the current meeting's summary and transcript. Migrates the legacy chat session format so existing users don't lose history.

## What's not in this PR

The Playwright E2E test suite (covering all routes and flows in the new renderer) is on a separate branch (`e2e-tests`) and will follow in a subsequent PR. Keeping it separate avoids the baseline screenshot images (~80 PNG/HTML files) inflating this diff.

## Test plan

- [ ] Toggle new renderer on in Settings → app reloads into new UI
- [ ] Toggle back to legacy renderer → original UI returns
- [ ] Start a recording, verify audio wave animates and timer ticks
- [ ] Stop recording, confirm Processing screen shows, then navigates to meeting detail
- [ ] Open a meeting → all four tabs load (Summary, Transcript, Insights, Calendar)
- [ ] Open Ask Bar (⌘K), ask a question, verify streaming response
- [ ] Create and rename a folder, move a meeting into it
- [ ] Dark/light theme toggle persists after restart
- [ ] Settings changes (audio device, model, storage path) persist after restart
- [ ] Cold-start test: kill all background processes, launch fresh, run full record→transcribe→summarize pipeline

---
## Summary by cubic
Replaces the legacy HTML renderer with a default `react` + `vite` UI, restores the full recording→processing flow with a bottom Ask Bar, and adds a setup wizard for local/cloud models with a per‑provider cloud model picker. Ships behind a tray toggle so users can switch back to the legacy UI if needed.

- **New Features**
  - Default React renderer with secure IPC via `preload.js` + `app/docs/ipc-contract.md`, a typed client (`renderer/src/lib/ipc.ts`), and `@tanstack/react-query`; components use `shadcn/ui`, `@radix-ui/*`, and `tailwindcss`. Includes a `/dev` sandbox and `pack:unsigned` CI build.
  - All screens rebuilt (Home, Note Detail tabs, Folders, Settings, Setup). Live Recording and Processing screens with audio wave, timer, draft notes, and a shared bottom dock. Ask Bar streams tokens and persists per‑note chat history (legacy format is migrated). Search works in Home/All notes.
  - Setup wizard with telemetry toggle and Local/Cloud summarization choice. Cloud path persists provider/key, runs an immediate API test, and skips model download; Local installs the bundled model. Cloud models are stored per provider with a one‑shot migration; Settings uses a per‑provider model picker (with “Custom…”).
  - Terminology and polish: “meeting” → “note” across UI. Default empty title is “Note” to trigger auto‑title. Sidebar/toolbar tweaks, better selection highlight, and theme toggle.

- **Bug Fixes**
  - YAML-safe frontmatter updates: title/updated_at are rewritten with proper quoting, preserving other fields and avoiding corrupted YAML and clobbered types.
  - Processing navigation: main process keys off Python’s “SAVED:<path>” marker to open the correct note after auto‑rename.
  - Ask Bar/transcript: fixed click‑outside so copy/search work; transcript header wave no longer animates. Chat sessions write atomically to v2 files, legacy read once then left untouched; corrupt files are quarantined; per‑note matching; optimistic updates roll back on save errors.
  - Security and stability: tight CSP; `openExternal` restricted to GitHub release URLs; quit dialog skipped on legacy and times out after 5s on the new UI; path validation for reveal‑meeting‑folder; removed broken postinstall; production builds don’t ship sourcemaps.
  - Debug logs captured from app start via a ring buffer with auto‑scroll in Settings → Developer.
  - Recording flow: notes list pre‑seeded on processing complete; Note Detail treats “fetching with no data” as loading; Ollama startup reuses an existing instance if `127.0.0.1:11434` is live.
  - P1/P2 fixes: cloud model migration avoids overwriting on parse/IO errors; cloud model dropdown no longer resets on selection; Ask Bar guards against rapid duplicate submits; a single processing‑complete listener is mounted at App level; route decoders handle bad `%` escapes; chips render real buttons with type="button"; typography style merging fixed; toolbar shows disabled “Processing” state; folder icon updates roll back on failure.

<sup>Written for commit 3c852327eb2211db35e43cf8fce274dddce3dee7. Summary will update on new commits.</sup>


WilliamDrewett:
### Code review

Found 2 issues:

1. **YAML frontmatter round-trip corrupts the `folders` field on every `.md` meeting rename.** When the `.md` update path reads frontmatter back, YAML parses `folders: ["id1","id2"]` as a JS array. `String(["id1","id2"])` then produces `"id1,id2"` (comma-joined, no brackets), which gets written back as `folders: "id1,id2"` — a YAML double-quoted string, not an array. On the next load, the Python parser's `startswith('[')` branch no longer matches and folder membership is silently dropped. Any user that renames an `.md`-backed meeting that belongs to a folder loses that folder association permanently and silently.

https://github.com/ruzin/stenoai/blob/ca928b3f4eddab6d158b9d9f93b5ffd4aef1b135/app/main.js#L1547-L1556

2. **6 IPC channels added but not documented in `ipc-contract.md`, violating the contract's own rule.** The doc at line 8 states: "Any PR that adds, removes, or changes an IPC channel must update this doc in the same commit." The following channels are exposed in `preload.js` but absent from the contract: `open-external`, `reveal-meeting-folder`, `regen-meeting-title`, `update-folder-icon`, `quit-dialog-response`, `show-quit-dialog`.

https://github.com/ruzin/stenoai/blob/ca928b3f4eddab6d158b9d9f93b5ffd4aef1b135/app/docs/ipc-contract.md#L7-L11

https://github.com/ruzin/stenoai/blob/ca928b3f4eddab6d158b9d9f93b5ffd4aef1b135/app/preload.js#L67-L72

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ruzin (approved):

PR #100, opened 2026-04-30
Add whisper model selection and keep recordings settings

valentinweyer:
## Summary

Adds two new user settings:

### 🎙️ Whisper Model Selection
Users can now select the Whisper transcription model size from the AI tab in settings. Models are auto-downloaded on first use via pywhispercpp.

Supported sizes: `tiny`, `base`, `small` (default), `medium`, `large`, `large-v3-turbo`

Transcription model selector

### 💾 Keep Recordings
Users can now choose to retain audio files after processing. When enabled, `.webm`/`.wav` recordings are not deleted after transcription and summarization complete.

Keep recordings toggle

## Changes
- `src/config.py` — `whisper_model` and `keep_recordings` config options with get/set methods
- `simple_recorder.py` — whisper model read from config at all `WhisperTranscriber()` instantiation sites; audio deletion guarded by `keep_recordings` flag; new CLI commands `get/set-whisper-model` and `get/set-keep-recordings`
- `app/main.js` — IPC handlers for all four new CLI commands
- `app/index.html` — transcription model dropdown in AI tab, keep recordings toggle in general settings

## Testing
CLI commands can be tested independently:
```
stenoai get-whisper-model
stenoai set-whisper-model large-v3-turbo
stenoai get-keep-recordings
stenoai set-keep-recordings True
```

---
## Summary by cubic
Adds Whisper model selection and a “Keep recordings” setting. Lets users pick accuracy vs. speed and choose whether to retain `.webm`/`.wav` files; adds a download-size hint for large models and structured error handling.

**New Features**
- Settings UI: Whisper model dropdown in AI tab (`tiny`, `base`, `small` (default), `medium`, `large`, `large-v3-turbo`); “Keep recordings” toggle in General.
- Transcription: pass chosen model to `WhisperTranscriber`; models auto-download on first use via `pywhispercpp`; toast shows expected download size for `medium`/`large`/`large-v3-turbo`.
- File retention: delete audio only when “Keep recordings” is off.
- Config: `whisper_model` and `keep_recordings` with get/set; model validated on read/write (falls back to `small`); persisted in config.
- CLI/IPCs: `get/set-whisper-model` and `get/set-keep-recordings`; handlers return structured JSON with errors and `supported_models`; UI wired to load/save via IPC.

<sup>Written for commit e160d7ae730f8c2b4b7114c9536f0bbd963f42fd. Summary will update on new commits.</sup>


ruzin:
@valentinweyer nice PR. Could you check cubic comments and I can approve?

valentinweyer:
@ruzin All 4 cubic issues have been addressed in the follow-up commit 2d8adb2:

- get_whisper_model() now validates on read and falls back to "small" if the config value is invalid
- PC handlers for get-whisper-model and get-keep-recordings now return { success: false, error } instead of silent fallbacks
- UI load functions check for result.error before updating the toggle/dropdown

Should be ready for approval.

ruzin (dismissed):

ruzin (commented):
Clean, well-scoped PR — both settings work the way you'd expect, the CLI surface is symmetric, validation lives on the Python side, and the JSON IPC pattern matches existing handlers. A few items worth addressing before merge:

### Bug — `.gitignore` adds a tracked file (medium)
`app/package-lock.json` is currently tracked in git (`git ls-files app/package-lock.json` confirms). Adding it to `.gitignore` doesn't untrack it but causes future updates to be silently ignored — which can drift the lockfile from reality and break reproducible signed-DMG builds. Either remove this `.gitignore` line or `git rm --cached app/package-lock.json` and commit that intentionally. I'd recommend the former: lockfiles should stay tracked for an app that ships to users.

### Suggestion — UX for large model first-load (low/medium)
Selecting `large` (~3 GB) or `large-v3-turbo` (~1.5 GB) triggers a `pywhispercpp` download on the next recording. There's no progress indicator — the first transcription will appear hung for minutes. At minimum, consider:
- Showing a toast warning when a non-`small` model is picked (\"will download ~3GB on next use\").
- Optionally pre-fetching via the existing `download-whisper-model` command after `set-whisper-model` succeeds, with progress surfaced like the Ollama model download flow does.

### Inconsistent error returns in IPC handlers (low)
In `app/main.js`:
```js
ipcMain.handle('get-whisper-model', async () => { ... return { success: false, error: e.message }; });
ipcMain.handle('set-whisper-model', async (...) => { ... return { success: false }; });  // no error
```
Same asymmetry on `set-keep-recordings`. Renderer code branches on `result.success` but won't have a message to show. Mirror the `get-*` handlers' `error: e.message` for consistency.

### Style nits (low)
- `simple_recorder.py:255-259` now imports `get_config` twice in adjacent blocks. Hoist a single import to the top of `transcribe_audio` (or to the module).
- `from src.config import get_config` is repeated inline in 4+ new places. The module already imports from `src` elsewhere — consider a top-level import.
- `set_whisper_model_cmd` uses an f-string with `config.SUPPORTED_WHISPER_MODELS` inside a JSON value — works, but `json.dumps({\"supported\": list(config.SUPPORTED_WHISPER_MODELS)})` is cleaner than embedding the repr.
- Two separate `DOMContentLoaded` listeners in `index.html` for the new toggle/select — fine, but the file already has a similar pattern; one combined listener would match the surrounding style.

### Test coverage
No new tests. `tests/test_config.py` already covers similar config flows; would be a small addition to cover:
- `get_whisper_model` falls back to `\"small\"` on invalid stored value.
- `set_whisper_model` rejects unknown sizes (returns False, doesn't persist).
- `get_keep_recordings` defaults to False when key absent.
- Round-trip persistence of both settings.

### Conventions
- No bare `exit()`; no shelling out to bundled binaries; user data path conventions respected.
- Settings UI mirrors the existing rows — visual consistency is good.

### Security
Nothing concerning. Boolean values are stringified safely before subprocess args; model sizes are validated against an allowlist on the Python side.

The cubic check is green and this is already approved — fix the `.gitignore` line, decide whether to address the large-model UX hint, and this is good to merge.

ruzin (approved):

PR #102, opened 2026-05-01
feat: enhance Chinese language support with variant handling and conversion

BrianNguyen291:
## Summary by cubic
Adds full Chinese variant support (Simplified and Traditional) with OpenCC-based conversion across transcripts, summaries, titles, and reports. Also ships diarised, timestamped transcripts in the UI and improves Ollama startup reliability.

- **New Features**
  - Chinese variants: UI now offers `zh-Hans`/`zh-Hant`. Map both to Whisper `zh`. Convert transcripts (incl. diarised text), streamed summaries, final summaries, generated titles, and reports via OpenCC. Fails safe — if OpenCC is missing or errors, output falls back to original text. Migrate legacy `zh` to `zh-Hans`.
  - Diarised transcripts: include [mm:ss] timestamps and interleave mic/others by start time; omit speaker tag for single-source recordings. UI parses and displays timestamps and copies `diarised_text` when present. Expose per-segment data from whisper.cpp and fall back when segments are missing.
  - Transcript/summary polish: add recording duration to transcript headers. Normalize markdown spacing in streamed and final outputs; prompt enforces spaces after headings and bullets.
  - Reliability: reuse an already running Ollama on 11434 to avoid “address already in use” and speed up retries.

- **Dependencies**
  - Added `opencc-python-reimplemented` (pinned) for Simplified/Traditional conversion; bundled OpenCC data files in packaging.

<sup>Written for commit 5e9b72ab7c30ecb2c693be17a0e86a873e5eb2d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/routes/settings/languages.ts
+++ b/app/renderer/src/routes/settings/languages.ts
@@ -3,8 +3,11 @@ import { PARAKEET_LANGUAGE_CODES } from '@/lib/transcription-languages';
 export type LangOption = { value: string; label: string };
 
 // Curated language list shown in Settings → Transcribe. Whisper supports
-// all 12 (it covers 99 languages at the model level; the dropdown is just
-// the tested curation).
+// all 13 (it covers 99 languages at the model level; the dropdown is just
+// the tested curation). Chinese is split into Simplified (zh-Hans) and
+// Traditional (zh-Hant): whisper.cpp always emits Simplified, so Traditional
+// is a post-transcription OpenCC conversion (see src/chinese.py) — both map
+// to whisper's "zh" for the actual ASR call.
 export const LANGUAGES_WHISPER: LangOption[] = [
   { value: 'auto', label: 'Auto (detect)' },
   { value: 'en', label: 'English' },
@@ -14,7 +17,8 @@ export const LANGUAGES_WHISPER: LangOption[] = [
   { value: 'nl', label: 'Dutch' },
   { value: 'pt', label: 'Portuguese' },
   { value: 'ja', label: 'Japanese' },
-  { value: 'zh', label: 'Chinese' },
+  { value: 'zh-Hans', label: 'Chinese (Simplified)' },
+  { value: 'zh-Hant', label: 'Chinese (Traditional)' },
   { value: 'ko', label: 'Korean' },
   { value: 'hi', label: 'Hindi' },
   { value: 'ar', label: 'Arabic' },
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,3 +40,8 @@ pydantic>=2.5.0
 python-dateutil>=2.8.0
 openai>=1.0.0
 anthropic>=0.18.0
+# OpenCC (pure-Python reimplementation) converts Chinese Simplified ↔ Traditional
+# for the zh-Hans / zh-Hant language options. Cross-platform, no native deps.
+# PINNED to an exact version — this ships inside the signed/notarised binary, and
+# its dictionary JSON is bundled via collect_data_files('opencc') in stenoai.spec.
+opencc-python-reimplemented==0.1.7
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -99,6 +99,29 @@ def resolve_output_language(
     return "en"
 
 
+def _apply_chinese_variant(text: Optional[str]) -> Optional[str]:
+    """Convert ``text`` to the user's selected Chinese script, if any.
+
+    whisper.cpp and Parakeet (and the summariser LLM) emit Simplified for
+    Chinese; a user who picked Traditional (``zh-Hant``) gets an s2t pass here.
+    A thin, never-throwing hook: non-Chinese languages, a missing OpenCC, or a
+    conversion error all fall through to the input unchanged so transcription
+    is never broken by variant conversion.
+    """
+    if not text:
+        return text
+    try:
+        from src.config import get_config
+        from src.chinese import apply_variant
+
+        variant = get_config().get_chinese_variant()
+        if variant:
+            return apply_variant(text, variant)
+    except Exception as e:
+        logger.warning(f"Chinese variant conversion skipped: {e}")
+    return text
+
+
 def resolve_persisted_output_language(
     session_info: dict,
     transcript_text: Optional[str],
@@ -464,9 +487,12 @@ Summary output language: {config.get_language_name(output_language)}
 
         # Get configured language
         configured_language = config.get_language()
+        # ASR only knows "zh" for Chinese; the Simplified/Traditional choice is
+        # applied as a post-transcription conversion, not an ASR mode.
+        asr_language = config.get_whisper_language()
 
         # Transcribe with diarisation support (stereo → [You]/[Others])
-        transcript_result = self.transcriber.transcribe_diarised(audio_path, language=configured_language)
+        transcript_result = self.transcriber.transcribe_diarised(audio_path, language=asr_language)
 
         # A transcription crash (e.g. an OOM on a long file) is not silence:
         # propagate the flag and skip writing a normal transcript file so the
@@ -502,6 +528,10 @@ Summary output language: {config.get_language_name(output_language)}
         if isinstance(transcript_result, dict):
             is_diarised = transcript_result.get("is_diarised", False)
             diarised_text = transcript_result.get("diarised_text")
+
+        # Convert the transcript to the selected Chinese script (no-op otherwise).
+        transcript_text = _apply_chinese_variant(transcript_text)
+        diarised_text = _apply_chinese_variant(diarised_text)
 
         output_language = self._resolve_output_language(
             configured_language, detected_language, transcript_text=diarised_text or transcript_text
@@ -736,7 +766,7 @@ Summary output language: {config.get_language_name(output_language)}
             err_msg = str(e).replace('\n', ' ').replace('\r', ' ')
             print(f"STREAM_ERROR:{err_msg}", flush=True)
             raise
-        streamed_md = ''.join(streamed_chunks)
+        streamed_md = _apply_chinese_variant(''.join(streamed_chunks)) or ''
 
         print("STREAM_COMPLETE", flush=True)
 
@@ -748,6 +778,7 @@ Summary output language: {config.get_language_name(output_language)}
             generated_title = self.summarizer.generate_title(
                 streamed_md, transcript_text, language=output_language
             )
+            generated_title = _apply_chinese_variant(generated_title)
             if generated_title:
                 session_name = generated_title
                 print(f"TITLE:{session_name}", flush=True)
@@ -1282,7 +1313,7 @@ def process_streaming(audio_file, name, notes, live_transcript, append_to):
             print(f"STREAM_ERROR:{err_msg}", flush=True)
             sys.exit(1)
 
-        streamed_md = ''.join(streamed_chunks)
+        streamed_md = _apply_chinese_variant(''.join(streamed_chunks)) or ''
 
         print("STREAM_COMPLETE", flush=True)
 
@@ -1293,6 +1324,7 @@ def process_streaming(audio_file, name, notes, live_transcript, append_to):
                 generated_title = recorder.summarizer.generate_title(
                     streamed_md, transcript_text, language=output_language
                 )
+                generated_title = _apply_chinese_variant(generated_title)
                 if generated_title:
                     session_name = generated_title
                     print(f"TITLE:{session_name}", flush=True)
@@ -2943,7 +2975,7 @@ def reprocess(summary_file, regenerate_title, retranscribe):
             print(f"STREAM_ERROR:{err_msg}", flush=True)
             sys.exit(1)
 
-        streamed_md = ''.join(streamed_chunks)
+        streamed_md = _apply_chinese_variant(''.join(streamed_chunks)) or ''
 
         # Regenerate the title when explicitly forced OR when the note still has
         # an auto/placeholder name. The latter is the common case now that the
@@ -2963,6 +2995,7 @@ def reprocess(summary_file, regenerate_title, retranscribe):
             generated_title = recorder.summarizer.generate_title(
                 streamed_md, transcript, language=output_language
             )
+            generated_title = _apply_chinese_variant(generated_title)
             if generated_title:
                 session_name = generated_title
                 existing_data["session_info"]["name"] = generated_title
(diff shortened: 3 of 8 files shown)


ruzin (commented):
Thanks for stacking these together — the Ollama-reuse path and diarisation rework both look good. A few things worth addressing before merge:

### Bug — `normalize_markdown` corrupts bold/italic at line start (medium)
`src/summarizer.py:23` uses `^([-*])(?=\S)`:
```
**bold**  →  * *bold**
*italic*  →  * italic*
```
Any line starting with markdown emphasis becomes a malformed bullet. Fix with a negative lookahead:
```python
text = re.sub(r'(?m)^([-*])(?=\S)(?![-*])', r'\1 ', text)
```
Worth a unit test on `**bold**`, `*italic*`, `- item`, `*item`, `1. item` to lock behavior.

### OpenCC packaging — verify in DMG (medium)
`opencc-python-reimplemented` ships JSON dictionary files. PyInstaller doesn't auto-collect them; if `stenoai.spec` doesn't include the package's data, runtime conversion will silently fall back to the input (the warning in `_get_converter` masks the failure). Please cold-test a packaged DMG with a Traditional-Chinese recording — not just `npm start`. If conversion fails, add a `collect_data_files('opencc')` entry to the spec.

### Silent failure swallow (low)
`simple_recorder.py:391` wraps the summary variant conversion in bare `try/except: pass`. Prefer:
```python
except Exception as e:
    logger.warning(f\"Chinese variant conversion failed: {e}\")
```

### Behavior change — `is_diarised` semantics (low, worth confirming)
Previously `is_diarised = bool(diarised_text)`. Now:
```python
is_diarised = bool(diarised_text) and not single_source
```
Single-channel recordings with timestamped output now report `is_diarised=False`. Worth a quick grep for `is_diarised` to confirm no downstream consumer (UI, summarizer prompts, file naming) relies on the old semantics.

### Style nits
- `from src.summarizer import normalize_markdown` is inlined 3× in `simple_recorder.py` — hoist to module-level.
- `get_language()` silently rewrites `\"zh\"` → `\"zh-Hans\"` on every read but never persists. A one-shot migration in `Config.__init__` would be cleaner.

### Test coverage
No new tests for non-trivial logic. Recommend:
- `tests/test_chinese.py` — `apply_variant` with simplified/traditional/None; ImportError path.
- `tests/test_summarizer.py` — `normalize_markdown` including the bold/italic regression above.
- `tests/test_config.py` — `zh` → `zh-Hans` migration, `get_whisper_language`, `get_chinese_variant`.
- `tests/test_transcriber.py` — diarised interleaving (single-source, both-channels, segment-less fallback).

Splitting into two PRs (Chinese support / diarisation+Ollama-reuse) would have eased review, but the changes are coherent enough as-is.

ruzin:
@BrianNguyen291 great PR, left comments!

ruzin:
@BrianNguyen291 do you wanna address the feedback or should I?

BrianNguyen291:
Sr, i was busy these days, i will fix it soon, then notify u, thank you

BrianNguyen291:
@ruzin later can u help me check? i added tests. please let's me know any needed . thanks

ruzin:
@BrianNguyen291 will merge today

ruzin:
@BrianNguyen291 hey brian, sorry, give me a couple more days, ive been busy getting some other features out.

ruzin:
@BrianNguyen291 sorry we have a collaborator who looked at this and we will be merging this

Optic00:
Non-blocking heads-up on the summariser input, since this is about to land:

The diarised path feeds `diarised_text` to the summariser — `text_for_summary = diarised_text or transcript_text` (`simple_recorder.py`, in the streaming + record + reprocess paths). With this PR each turn in `diarised_text` is now prefixed `[mm:ss] [You]/[Others] …`, so from this change on the **summary model sees the `[mm:ss]` stamps in its prompt**, where before it saw only the speaker labels. Reprocessing an existing saved transcript picks them up too.

Probably harmless (the timestamps might even help the model with ordering), but it's a change to the model's deterministic input on every diarised meeting, so flagging in case it wasn't intended. If you'd rather keep the model input byte-identical to today, a one-line guard at the summariser chokepoint does it — strip a leading `^\[\d+:\d{2}\]\s*` (MULTILINE) before building the prompt, leaving the on-disk/displayed transcript and the speaker labels untouched. Happy to PR that as a follow-up if useful; either way it's not a blocker for merging this.

Optic00:
Nice work on the `[mm:ss]` diarised prefixes — the chronological turn ordering reads really well.

One behavioural question on the summariser side, not a blocker: with this change the `[mm:ss]` prefix becomes part of `diarised_text`, which is what gets persisted as the saved transcript (`saved_transcript = diarised_text if diarised_text else transcript_text`) and then fed verbatim into the summariser prompt (`_create_markdown_prompt` / the structured prompt). The summariser changes here are markdown-normalisation only, so they don't strip the prefix.

Net effect: from this PR on, the summary model sees `[mm:ss] [You] …` on every turn for every diarised meeting — and **reprocessing previously-saved transcripts** changes too, since those files now carry the stamps. That's a change to the model input that was `[You] …` before.

Is that intended (giving the model time context) or an unintended side effect? If the latter, a tiny guard keeps the model input byte-for-byte identical to today while preserving the stamps for display/storage:

```python
_TURN_TIMESTAMP_RE = re.compile(r"^\[\d+:\d{2}\]\s*", re.MULTILINE)

def _strip_turn_timestamps(transcript):
    """Drop per-turn [mm:ss] display prefixes before summarising; keep [You]/[Others]."""
    return _TURN_TIMESTAMP_RE.sub("", transcript) if transcript else transcript
```

…called at the top of `_create_markdown_prompt` (and the structured prompt builder). Happy to send the guard + a small unit test as a follow-up PR after this merges if that's the preference — didn't want to pre-empt the design call.

Optic00:
Heads-up @BrianNguyen291: after this weekend's v0.5.8 merge batch the branch now conflicts with `main`. The two big ones, so the rebase doesn't come as a surprise:

- `Settings.tsx` was split into per-tab modules under `app/renderer/src/routes/settings/` (#289) - the settings changes here need to move into the matching tab file (the transcription/language bits into `TranscriptionTab.tsx`).
- `simple_recorder.py` / `src/transcriber.py` changed substantially with the transcript/summary decoupling (#276).

ruzin has signalled he'd like to get this merged, so a rebase should be the last blocker. Happy to help with the conflict resolution if useful.

Optic00:
Picking this back up so we can finally land it. I rebased the branch onto current `main` locally and want to flag what I found before pushing anything, because it changes what this PR should become.

The branch carries two independent things, and they've aged very differently against `main`:

**1. The diarised `[mm:ss]` + `[You]`/`[Others]` timestamps are now redundant.** `main` grew its own timestamped diarisation independently (`src/transcriber.py:1283` builds `[mm:ss] [Speaker]` turns via `transcribe_diarised`), plus per-segment speaker-bleed correction (`_drop_per_segment_bleed`), honest-failure handling on a dead channel (#207), and a summariser that strips those markers back out before prompting (`summarizer._strip_leading_timestamps`). So every conflict on this rebase is `main`'s newer version of that same feature against this branch's older one. Taking the branch side would regress all of the above, so I'd drop those hunks and keep `main`.

**2. The Chinese Simplified/Traditional variant support is the real, still-missing value.** Nothing on `main` does this, and it's cleanly separable:

- `src/chinese.py` (OpenCC s2t/t2s, graceful passthrough if the lib is missing)
- `Config.get_whisper_language()` + `get_chinese_variant()`, and the `zh` -> `zh-Hans` legacy migration
- mapping `zh-Hans`/`zh-Hant` -> `zh` for whisper.cpp (it only knows `zh`), then converting the transcript and the LLM summary to the chosen variant
- `normalize_markdown()` for CJK output that comes back without ASCII spaces around headings/bullets
- the language-picker split (Simplified / Traditional) and `tests/test_chinese.py`

So my proposal is to reduce this PR to just the Chinese-variant slice, re-applied on top of today's `main` rather than force-merged through the stale diarisation conflicts. One detail I'd carry over from `main`: `get_language()` now defaults to `"auto"` (#281), not `"en"`, and I'll keep that. There's also one thing to verify before it ships: `opencc-python-reimplemented` bundling in the PyInstaller build (the packaged app needs its conversion dictionaries), so I'll confirm the DMG/installer picks those up.

One honest caveat on my side: I can verify the plumbing (the whisper mapping, the markdown normalization, the config migration are all deterministic and unit-testable) and OpenCC's conversion is a mature, trusted library, but I don't read Chinese, so I can't judge that the end-to-end output actually reads right in either script. @BrianNguyen291 you can - so before this merges I'd ask you to run a real Chinese recording through it in both Simplified and Traditional and confirm the transcript and summary come out correct. You'd stay the natural owner for any follow-up a Chinese user reports, since neither @ruzin nor I could triage that.

@BrianNguyen291 since maintainer-edit is on, I can push the reworked slice straight to this branch and keep you as the author, so your Chinese support is what lands. @ruzin does reducing it to that scope work for you, and are you happy carrying a feature whose language none of the current maintainers read? If you're both good with it I'll push the rebased version for Brian to validate.

Optic00 (changes requested):
## Verdict

Request changes. The Chinese-variant slice is a genuinely useful addition and the earlier review points were mostly addressed, but the PR no longer applies to `main` (conflicting), roughly half of it duplicates diarisation/timestamp work that has since landed on `main`, and the summary-side Chinese conversion is dead code that never runs.

## Summary

This PR adds Simplified/Traditional Chinese support (`zh-Hans`/`zh-Hant` in the UI, both mapped to whisper `zh`, OpenCC conversion of transcripts and summaries, a one-shot `zh` -> `zh-Hans` migration) and bundles the `opencc-python-reimplemented` dependency. It also carries a diarisation rework (per-segment `[MM:SS]` timestamps, interleaving, single-source handling) and a `normalize_markdown` pass. Since the branch was opened, `main` has independently gained the diarisation-timestamp feature (#272, #284), so those parts now conflict and are largely redundant.

## Findings

### Blocker

None.

### Important

- **Chinese summary/title conversion never runs** — `simple_recorder.py:388`. The block is guarded by `isinstance(summary_result, dict)`, but `summarize_transcript()` returns a `MeetingTranscript` object, not a dict (the very next lines read it via `getattr(summary_result, 'overview', ...)`). The keys it tries (`summary`, `title`, `text`) are not `MeetingTranscript` fields either. On top of that, the streaming summary paths (`simple_recorder.py:587`, `:921`, `:1560`) — which is what the app actually uses — never call `apply_variant`. Net effect: only the transcript is converted; Traditional-Chinese summaries and titles are not enforced, which contradicts the PR description. Conversion should run on the streamed markdown (and/or the extracted fields) rather than on a dict that never exists.

- **Diarisation/timestamp hunks are already on `main`** — `src/transcriber.py:800-817`, `app/renderer/src/components/TranscriptPanel.tsx:173`, `app/renderer/src/components/AskBar.tsx:34`. `main` already has `is_diarised = bool(mic_segments) and bool(system_segments)` and emits `[MM:SS] [You] ...` via `_format_timestamp` (with `H:MM:SS` past an hour), plus renderer timestamp parsing. This PR reintroduces a narrower version: `_fmt_ts` only formats `MM:SS`, so a recording past 99:59 emits three-digit minutes (`[100:00]`) that the renderer regex (`\d{2}`) can't match, dropping the timestamp row. Please rebase and drop these hunks in favour of the `main` implementation.

- **`normalize_markdown` still corrupts some valid markdown** — `src/summarizer.py:33-34`. The `**bold**` / `*italic*` (whole-line) regression from the earlier review is fixed, but inline emphasis followed by text still breaks: `*italic* rest` -> `* italic* rest`, `--flag` -> `- -flag`, `#######Title` -> `###### #Title`. It also rewrites inside fenced code blocks. Worth either a more targeted rule (only treat as a bullet when the whole line is emphasis, and skip code fences) or accepting the limitation explicitly. The added test only covers the whole-line cases.

### Nit

- **Live streamed summary is shown un-normalized** — `simple_recorder.py:584-587`. Chunks are emitted to the Processing UI before `normalize_markdown` runs, so only the saved/parsed markdown is normalized; the live view still shows the malformed spacing.

- **Scope** — the PR mixes Chinese support, diarisation timestamps, and markdown normalization under a "Chinese language support" title. Now that the diarisation half is on `main`, extracting a rebased Chinese-only PR would be clean and easy to review.

- **OpenCC packaging needs a real DMG check** — `stenoai.spec:86`. `collect_data_files('opencc')` targets the right package (`opencc-python-reimplemented` imports as `opencc`), so the code looks correct, but the dictionaries only fail at runtime. A cold packaged-DMG test with a Traditional-Chinese recording is the only way to confirm the JSON dicts are bundled (the `_get_converter` warning otherwise silently falls back to the input).

### Question

- Does the transcript-side variant conversion in `simple_recorder.py:273` cover the parakeet path too, or only the whisper diarised path? The conversion runs on `transcript_result` regardless of engine, which looks right, but worth confirming the parakeet dict shape carries the same `text`/`diarised_text` keys.

### Praise

- Good, focused dependency choice: `opencc-python-reimplemented` is pure-Python with JSON dictionaries, no native build and no heavy runtime, so it fits the project's lean-dependency posture and is cross-platform clean.
- The `zh` -> `zh-Hans` migration in `Config.__init__` and the `get_whisper_language` / `get_chinese_variant` split are clean, and the earlier review points (bold-fix, `collect_data_files`, `logger.warning` instead of a bare `except: pass`) were addressed.

ruzin:
Rebased onto current `main` and focused this down to just the Chinese Simplified/Traditional support (the unique, valuable part). The other original changes were dropped because `main` has since implemented them independently: diarised transcript timestamps, markdown-spacing normalization, and the reuse-already-running-Ollama fix.

What's here now: OpenCC converter (`src/chinese.py`), `zh-Hans`/`zh-Hant` language options + legacy `zh` migration, Simplified→Traditional conversion of the transcript + summary, and the Settings language picker. The OpenCC dep is pinned (`opencc-python-reimplemented==0.1.7`, pure-Python, ~+0.5 MB to the DMG) and there's a unit test that runs a real s2t conversion (also confirms it works on Python 3.11+). Thanks @BrianNguyen291 — kept you as co-author on every commit.

PR #105, opened 2026-05-08
System audio: real diarisation via CoreAudio Tap + Granola-style transcript UI

ruzin:
## Summary
- Make system-audio recording actually work end-to-end on macOS 14.4+: switch the loopback path from ScreenCaptureKit (which was silently returning empty right channels) to CoreAudio Process Tap, fix the diarisation pipeline so it survives Electron's stripped-PATH subprocess environment, and add automatic speaker-bleed detection so no-headphones recordings don't show fake duplicate bubbles.
- Replace the flat "[YOU] / [OTHERS]" pill rendering with Granola-style chat bubbles — green right-aligned for the mic, grey left-aligned for system audio, plain full-width lines when diarisation isn't viable.
- Several supporting fixes: bundle-aware ffmpeg resolution (no need to ship ffprobe), corrected RMS threshold and sampling for headphones-mode quiet mic input, electron-builder 26 notarize config schema fix.

## What's in this PR
**CoreAudio Tap as the only loopback backend**
- `app/main.js`: `initMain({ forceCoreAudioTap: true })`. The `electron-audio-loopback` library always forces *one* of the two Chromium feature flags; the previous default was `MacSckSystemAudioLoopbackOverride` which on macOS 14.4+ returns a stereo stream whose right channel is digital silence. Switching to `MacCatapSystemAudioLoopbackCapture` was the original intent of this branch but had not actually been wired.
- New IPC `get-system-audio-support` reports `{ supported, osVersion, screenPermission }`. The system-audio toggle in Settings is disabled on macOS <14.4 with a "requires macOS 14.4+" hint, and `useSystemAudioCapture` short-circuits on unsupported OS so older Macs never reach `getDisplayMedia`.
- Startup log of macOS version + screen-recording permission status so silent-tap regressions don't hide.

**Diarisation pipeline robustness (`src/transcriber.py`)**
- ffmpeg path resolution mirrors `simple_recorder.py`'s bundled-first lookup. Electron-spawned subprocesses don't inherit the user's shell PATH, so a bare `ffmpeg` string was failing under hardened-runtime.
- Replaced the `ffprobe` channel-count call with `ffmpeg -i FILE -f null -` stderr parsing — avoids needing to bundle a second binary just for channel detection.
- RMS gate: lowered from 0.005 (-46 dB) to 0.0003 (-70 dB). Headphones-mode mic captures audio at much lower amplitude than speakers-mode, and the old threshold was failing legitimate quiet speech.
- RMS gate: switched from "average over first 5 seconds" to "max over 1-second sliding windows across the whole file" — recordings where speech starts mid-stream were being classified as silent.
- Jaccard-based speaker-bleed detection: when mic and system transcripts overlap above ~0.6 (the no-headphones case where the mic captures both the user and the system audio echoing through speakers), collapse to mic-only. Avoids fake duplicate `[You]`/`[Others]` bubbles for the same content.

**Chat-bubble transcript UI (`app/renderer/src/components/TranscriptPanel.tsx`)**
- Granola-style: green right-aligned bubble for `[You]`, grey left-aligned for `[Others]`, no text labels (alignment + colour carry the cue), 78% max-width.
- Non-diarised content (mono recordings or bleed-collapsed) falls back to plain full-width lines — the chat metaphor isn't forced when there's only one effective speaker.

**Build pipeline**
- `app/package.json`: `mac.notarize` from `{ teamId: ... }` object to boolean `true`. The Electron 31 → 42 / electron-builder 24 → 26 bump introduced a schema change; the old shape was failing validation, breaking any `npm run build`. CI release credentials are already set via env vars (`APPLE_ID`, `APPLE_APP_SPECIFIC_PASSWORD`, `APPLE_TEAM_ID`) in `.github/workflows/build-release.yml`.

## Test plan
- [x] Install the new DMG in `/Applications`, kill any prior StenoAI process, launch fresh
- [x] First recording attempt prompts for **Audio Capture** (CoreAudio Tap) permission in addition to Screen Recording — accept both, then quit + relaunch (CoreAudio taps need a fresh session under the new permission)
- [x] **No headphones, mic + YouTube clip**: transcript renders as flat plain text (Jaccard collapse fired), summary references both your speech and the clip content
- [x] **Headphones on, mic + YouTube clip**: transcript renders with green right-aligned bubbles for your speech and grey left-aligned bubbles for system audio
- [x] System Audio toggle in Settings: visible enabled on macOS 14.4+, disabled with "requires macOS 14.4+" hint on older test VM
- [x] Mic-only recordings (system audio toggle off) still produce a single flat transcript and continue to work
- [x] Verify the packaged DMG flow per CLAUDE.md — dev mode (`npm start`) does not apply the `extendInfo` Info.plist keys, so CoreAudio Tap testing must be done from the installed DMG

---
## Summary by cubic
System audio recording now works end to end on macOS 14.4+ via CoreAudio Process Tap with real mic/system diarisation and a Granola‑style transcript. Also adds streamed markdown summaries, a neutral paper+ink palette, an update toast for downloaded releases, and fixes for short‑clip silence detection and capture lifecycle races.

- **New Features**
  - CoreAudio Tap loopback is default; a support‑probe IPC gates macOS 14.4+ and startup logs capture OS version and Screen Recording permission.
  - Default ON: `system_audio_enabled` now defaults to true; `main.js` gates to CoreAudio Tap only on macOS 14.4+ and falls back to the Python recorder on older macOS/non‑darwin. Settings show a “requires macOS 14.4+” hint and disable the toggle when unsupported.
  - Renderer‑driven dual capture: mix mic (L) + system (R) in Web Audio, record WebM/Opus via `write-system-audio-blob`, queue for processing; skip the Python `record` process.
  - Diarisation: per‑segment timings with chronological interleave and speaker‑bleed collapse (no‑headphones → mic‑only). `TranscriptPanel` shows right‑aligned green bubbles for you and left‑aligned grey for others; mono falls back to full‑width lines.
  - Processing: stream markdown via `react-markdown` with 50ms buffered updates and a FLIP‑smoothed scanner bar; `processing-complete` includes `meetingData` for auto‑navigation.
  - Updates: top‑right toast announces a downloaded update with a Restart button to install immediately.
  - Brand/UI & deps: switch to paper+ink palette and set `--accent-primary` to foreground ink; add `.stream-markdown` styles. Bump `electron` → `42`, `electron-builder` → `26`, `electron-audio-loopback` → `1.0.6`; add `react-markdown`.

- **Bug Fixes**
  - Media tools: resolve `ffmpeg` from bundled/known paths with a thread‑safe cache; replace `ffprobe` with `ffmpeg -i -t 0` stderr parsing for fast channel/duration detection.
  - Energy gating: lower RMS threshold, scan 1s windows across the whole file, and correctly handle sub‑1s clips so quiet or late speech is detected.
  - Recording lifecycle: prevent double starts, make stop idempotent, cancel in‑flight capture when stopping, and stop/clean up when support or the toggle flips off; keep tray/elapsed accurate.
  - Processing UX: reduce markdown re‑parses; only enable the scanner bar’s compositor layer during animation.
  - Tests: add coverage for Jaccard bleed detection, ffmpeg stderr parsers, RMS scanning (including short clips), and fix mocks with numeric segment timings.

<sup>Written for commit 502245bd45cf3b02b3c6a942fd2839765304c6ca. Summary will update on new commits.</sup>


ruzin:
Addressing the cubic-dev-ai review feedback in `0070bb2`:

**Fixed**
- `src/transcriber.py` — short audio (<1s) classified as silent. The 1-second sliding window's `pos + window <= n_frames` guard never let the loop enter when `n_frames < window`. Added a short-clip path that scans the whole file as a single window, with new tests for loud and silent sub-second clips.
- `useSystemAudioCapture.ts:76` (P1) — cancellation guard inside `startCapture`. Each `await` (getUserMedia, enableLoopbackAudio, getDisplayMedia) now compares against a `startTokenRef` bumped by `stopCapture`; if the token moved, we stop the tracks we just acquired and bail before installing them. No more pending starts completing after stop.
- `useSystemAudioCapture.ts:50` / `:34` (P1/P2) — the `if (!enabled) return` early-out at the top of the effect bypassed `stopCapture` when `enabled` flipped false during an active recording (e.g., support resolving false under the loading-state-permissive `?? true` fallback, or the user toggling system audio off mid-recording). Now the stop path fires explicitly when `!enabled && activeRef.current` before the no-op return.

**Already fixed in earlier review-feedback commit `b519168`**
- `src/transcriber.py:468` — `-t 0` was added to the channel-detection ffmpeg invocation so it parses the input header and exits without decoding any audio frames.

**Not a regression — already aligned**
- `Settings.tsx:676` vs recorder logic (`?? true` vs `?? false`) — both `Settings.tsx` (`checked` fallback) and `useSystemAudioCapture.ts` (`enabled` fallback) use `?? true` since `b519168`. The flagged misalignment was based on an earlier diff snapshot.

35/35 tests pass, renderer typechecks, ruff clean. Ready for re-review.

PR #106, opened 2026-05-09
feat: add whisper model selector and keep recordings to React UI

valentinweyer:
## Summary

Completes the whisper model picker and keep recordings toggle in the new React+Vite renderer introduced in v0.3.0. The backend CLI and config were merged in #100 but the frontend was not ported to the new renderer.

### Changes

- `app/renderer/src/lib/ipc.ts` — add `GetWhisperModelResponse`, `GetKeepRecordingsResponse` types and IPC definitions
- `app/renderer/src/hooks/useSettings.ts` — add `useWhisperModelSetting`, `useSetWhisperModel`, `useKeepRecordingsSetting`, `useSetKeepRecordings` hooks
- `app/renderer/src/routes/Settings.tsx` — Keep recordings toggle in General tab, `WhisperModelSelector` dropdown in AI tab
- `app/main.js` — wrap get/set handlers in `{ success: true }` envelope to match the `unwrap()` contract
- `app/preload.js` — expose all four new methods on the `stenoai` bridge

### Testing

Both settings load correctly on app start and persist across restarts. The whisper model dropdown shows all 6 sizes with descriptions. The keep recordings toggle reflects the saved state.

---
## Summary by cubic
Adds a Transcribe tab with a Whisper model picker (download with progress) and a “Keep recordings” toggle. Adds a 30‑minute Whisper download timeout with guaranteed completion; hardens downloads with per‑process temp files and keeps the model list single‑sourced.

- New Features
  - UI: Transcribe tab with Language, Keep recordings, and Whisper model cards (Small, Large V3 Turbo) that download from HuggingFace with progress and auto-select on completion.
  - Hooks/Bridge: `whisperModels.{list,set,pull}` plus progress/complete events; keep recordings get/set; existing settings handlers return `{ success: true }` and are exposed on `stenoai`.
  - Backend: `list-whisper-models` and `pull-whisper-model`; new `src/whisper_models.py` registry/downloader; config migration maps legacy values (`large` → `large-v3-turbo`, others → `small`); `SUPPORTED_WHISPER_MODELS` now derives from the registry to stay in sync.

- Bug Fixes
  - Audio: System audio now saves in recordings/; on processing failure, source audio is deleted when Keep recordings is off.
  - UI: Fixed FLIP animation measurement in Processing to prevent bar jitter; unified size/default formatting across model lists.
  - Whisper: 30‑minute timeout with SIGTERM→SIGKILL escalation and a force‑settle path to always emit `whisper-pull-complete`; per‑process `.part` files prevent concurrent download clobbering; unit tests cover the downloader (including mid‑stream failure cleanup) and config migration.

<sup>Written for commit bd7494082a98eb93675bceb79ae4953b9c330367. Summary will update on new commits. <a href="https://cubic.dev/pr/ruzin/stenoai/pull/106?utm_source=github">Review in cubic</a></sup>


ruzin:
Thanks @valentinweyer will review and this will go in mid week.

ruzin:
@valentinweyer feel free to join the discord btw!

allaire:
Was looking for this setting, looking forward to get this in the next update, thanks!

ruzin:
Hey @valentinweyer — thanks for the original PR. I rebased onto current `main` and extended the scope while testing, so the branch now has 4 commits. Quick map so reviewers (cc @cubic-dev-ai) know what to look at:

**1. `babd576` — your original** (rebased, no functional change)
Your `feat: add whisper model selector and keep recordings to React UI`, replayed on top of `main` (was based on pre-v0.3.0).

**2. `53c9134` — `fix(processing)`: scanner-bar FLIP fix**
Independent bug surfaced while testing — the analyzing/generating bar was oscillating up/down during summary streaming. Root cause was `getBoundingClientRect()` reading a position contaminated by an in-flight transform, so `lastTopRef` captured the wrong baseline and the next delta flipped sign. Clearing `transform: none` before measuring gives the true layout position.

**3. `0a072f5` — `feat(transcribe)`: Transcribe tab + Whisper model cards**
While testing your dropdown I hit a hard crash picking `large`: pywhispercpp's `AVAILABLE_MODELS` doesn't include the bare `large` alias, so `Model('large')` → `download_model` returns `None` → `whisper_init_from_file(None)` segfaults whisper.cpp natively. Process exits with signal-kill and the UI shows a generic \"Couldn't process this recording.\"

To fix this properly + match the Ollama-model UX (which already does download cards), I:
- Added `src/whisper_models.py` (model registry + HF downloader with progress callback).
- Added `list-whisper-models` / `pull-whisper-model` CLI commands modeled on the existing `pull-model`.
- Added `whisperModels.{list,set,pull}` IPC + `whisper-pull-progress` / `whisper-pull-complete` events.
- Added `useWhisperModels` / `useSetWhisperModel` / `usePullWhisperModel` hooks (same pattern as `usePullModel`).
- Replaced your dropdown with a `WhisperModelList` of `<ModelCard>`s, in a new **Transcribe** tab (containing Language, Keep recordings, and Model). The AI tab is now just provider + summarisation model.
- Curated the list down to **Whisper Small** (default) and **Whisper Large V3 Turbo** — both multilingual, both cover every language in the dropdown + autodetect. Turbo at half the size of large-v3 is near-identical accuracy.
- Added a one-shot `_migrate_whisper_model` so existing configs with `large` / `medium` / etc. land on a supported tier on next load (no crash).

**4. `91d2d9a` — `fix(recordings)`: sysaudio → `recordings/` + orphan cleanup**
Two related cleanups around audio file management:
- Renderer sysaudio capture was writing to `system_audio_tmp/` while the mic path used `recordings/`. With `keep_recordings=ON` files would persist in the temp folder indefinitely while users naturally looked in `recordings/`. Aligned both paths onto `recordings/`.
- Python's cleanup only ran in the success branch — a processing crash left the .webm/.wav orphan even when `keep_recordings=OFF`. `processNextInQueue`'s catch handler now re-reads the setting and unlinks on failure when keep is off.

All tested in dev mode (pyinstaller + electron) against system audio + Whisper download flow. Force-push was authorized (`maintainerCanModify: true`) — sorry for overwriting your original SHA.

ruzin (dismissed):

ruzin:
@cubic-dev-ai good catch on the unsignaled-kill hang. Fixed in 64fa759:

- **T+0**: SIGTERM (graceful)
- **T+5s**: SIGKILL (kernel can't block this)
- **T+15s**: force-settle the Promise + emit `whisper-pull-complete` even if `close` never fires (zombie / uninterruptible wait edge case)

All four exit paths (close, error, timeout-force-settle, escalation) now route through a single `finishOnce()` gate keyed on a `settled` boolean, so multiple races can't double-resolve or double-send the IPC event.

Please re-review.

ruzin (dismissed):

ruzin:
@cubic-dev-ai round 2 addressed in ab3cb73:

**P1 race on `.part` file** → per-process unique temp name (`.part-<pid>-<random>`). Whoever renames first wins; the loser's tmp gets cleaned by its own exception path. Atomic rename on the same filesystem makes the second rename a safe overwrite with identical bytes.

**P2 dual registry source of truth** → `Config.SUPPORTED_WHISPER_MODELS` is now derived from `whisper_models.SUPPORTED_WHISPER_MODELS.keys()` at import. Adding a model in one place can no longer skip the validation list.

**P2 duplicated list rendering** → extracted `formatModelSize()` and `isDefaultModel()` helpers, used by both `ModelList` and `WhisperModelList`. This fixes the divergence you pointed out (MB/GB conditional vs always-GB; description regex vs displayName regex). Kept the list components separate because `ModelList` has surface area Whisper doesn't need (deprecated section toggle, remote/local provider split) — merging would force one component to be both rather than dedupe. Open to a full merge if you'd prefer.

Please re-review.

ruzin:
@cubic-dev-ai round 3 — one fix, one deferred with rationale:

**P2 test bug (`test_cleans_up_part_file_on_network_failure`)** — fixed in bd74940. You're right that mocking `requests.get` to raise immediately short-circuited before any `.part` file existed; the cleanup assertion was vacuous. Rewrote as a realistic mid-stream interruption: connection succeeds, headers arrive, `iter_content` yields a few real chunks (so the with-open block actually creates and writes to `.part`), then raises `ConnectionResetError`. The except handler now has a real file to clean up and the assertion meaningfully verifies that path.

**P2 duplicate pull-handler lifecycle (`pull-model` vs `pull-whisper-model`)** — deferred. The duplication is pre-existing: Ollama's `pull-model` has been shipping for many releases; this PR's `pull-whisper-model` is a parallel handler (and adds the timeout / SIGTERM→SIGKILL / force-settle safety machinery on top of the common skeleton). A proper dedup means refactoring `pull-model` to consume a shared helper, which is real refactoring risk against working production code for an automated-reviewer style win. This PR is already 6 commits deep on cubic feedback; another structural refactor of an adjacent handler is better as its own focused follow-up with its own testing surface than bundled here. Happy to track it as a separate issue/PR if @ruzin agrees.

ruzin (approved):

PR #177, opened 2026-06-12
Local Ollama: context cap + num_ctx so long-meeting summaries cover the whole meeting

ruzin:
Split out of #176 per review discussion.

### Problem
Long transcripts silently overrun small local models' default 2–4k context. On overflow Ollama drops the **oldest** tokens — including the prompt instructions at the head — so summaries covered only part of the meeting or came back malformed. Affects the default `llama3.2:3b` on every long meeting.

### Change (local Ollama only — cloud/remote/adapter byte-identical, tested)
- `num_ctx=8192` passed to the local Ollama chat calls (summary streaming + non-streaming + the Ask/chat query path).
- Transcripts over ~24k chars are capped to head 60% + tail 40% joined by a neutral in-prompt gap marker — head+tail so the meeting's close and action items survive.
- A user-facing note ("this meeting was longer than the local AI model's context window…") is appended **deterministically** to the streamed summary when capping occurred — never trusted to the model to echo — and folded into the parsed `summary` field by both md parsers so it survives into the meeting view, copy, and share. `generate_title` strips it from its context.

### Verified at runtime (on the original branch)
A 33k-char transcript reprocessed on local Ollama produced action items from the transcript **tail** ("send the revised budget by Friday") — proving the capping preserves the close — plus the note in the stream, the saved `.md`, and the parsed `list-meetings` summary. Short transcripts get no note. 23 new tests.

### Known limitation / planned follow-up (the reason this was split out)
The 8192/24k sizing is fixed per *provider*, not per *model*. Registry models with large native context (qwen3.5:9b, deepseek-r1:14b, gpt-oss:20b — all 128k-native) get clamped and could see the truncation note on meetings they'd handle natively. Planned fix before merge: a `num_ctx` field per `SUPPORTED_MODELS` registry entry (8192 for the 3–4B models, 32768 for 9B+ — which covers any transcript the 2-hour recording cap can produce), with the char cap derived from it.

---
## Summary by cubic
Make long local-model summaries and Ask answers cover the whole meeting by using an 8k context window and head+tail transcript capping, with a clear truncation note that persists in the summary. Cloud/remote/adapter providers are unchanged.

- **New Features**
  - Pass `options: { num_ctx: 8192 }` to local Ollama chat calls (summary and Ask; streaming and non-streaming).
  - Cap transcripts over ~24k chars to head 60% + tail 40% with a neutral gap marker across all prompt builders.
  - Append a deterministic truncation note at the end of the streamed summary when capping occurs; fold it into the parsed `summary` from both streamed and saved markdown; `generate_title` strips it; renderer removes the leading `>` for clean display.

- **Bug Fixes**
  - Prevent local summaries from dropping instructions and end-of-meeting details due to small default context windows.

<sup>Written for commit 40fa4230ab2e763e32e4e993dc7e8638d51e8fef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/177?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/routes/MeetingDetail.tsx
+++ b/app/renderer/src/routes/MeetingDetail.tsx
@@ -778,7 +778,9 @@ function parseMarkdownBlocks(text: string): MarkdownBlock[] {
       flushBullets();
     } else {
       flushBullets();
-      blocks.push({ type: 'paragraph', text: line });
+      // Strip blockquote markers (e.g. the local-model truncation note,
+      // "> Note: …") rather than rendering a literal "> " prefix.
+      blocks.push({ type: 'paragraph', text: line.replace(/^>\s+/, '') });
     }
   }
   flushBullets();
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -108,6 +108,33 @@ def _start_summary_heartbeat(label: str = "summarize", interval_s: int = 60, max
     return stop
 
 
+# The deterministic truncation note the local summariser appends to the
+# streamed markdown (src.summarizer.LOCAL_TRUNCATION_USER_NOTE). Detected by
+# prefix rather than importing the constant so the lightweight parse/list
+# paths don't pull in the summarizer module (ollama client etc.).
+_TRUNCATION_NOTE_PREFIX = '> Note: this meeting was longer than the local AI model'
+
+
+def _extract_truncation_note(text: str) -> Optional[str]:
+    """Return the truncation note's content (sans blockquote marker) if the
+    markdown contains one, else None.
+
+    The note trails the last model section in the streamed markdown, so the
+    section parsers drop it from their structured fields; callers fold it
+    into the ``summary`` field instead so it survives into the meeting view,
+    copy, and share — not just the transient streaming view.
+    """
+    if not text:
+        return None
+    for line in text.split('\n'):
+        stripped = line.strip()
+        if stripped.startswith(_TRUNCATION_NOTE_PREFIX):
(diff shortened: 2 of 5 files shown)


ruzin:
Thanks @cubic-dev-ai — both P2s addressed in `40fa423`:

**P2 (simple_recorder.py — note detection scanned full body):** Valid. `_parse_meeting_markdown` now scans only the summary sections, excluding `## Transcript` (`note_search = '\n'.join(v for k,v in sections.items() if k != 'transcript')`). This is broader than the suggested `action items`-only scan on purpose: the note can trail a different last section if a model omits Action Items, so excluding just the arbitrary transcript is the robust fix. Added a regression test (`test_parse_meeting_markdown_ignores_note_text_in_transcript`) proving note-wording in the transcript is no longer folded.

**P2 (summarizer.py — fixed num_ctx/char cap, not model-adaptive):** This is a deliberate trade-off, now documented in the constants block. Sizing num_ctx up to a large-context local model's capacity scales KV-cache RAM linearly and would reintroduce the OOM that #175 just fixed, on the low-end machines that run small local models. 8k is a RAM-safe conservative floor; capability-aware sizing (min(model_ctx, RAM-derived ceiling)) needs reliable host-RAM awareness and is a tracked follow-up.

PR #187, opened 2026-06-13
fix: make note deletion feel instant (optimistic cache removal)

Optic00:
## Description

**Problem.** Deleting a note felt broken: you'd click Delete, nothing visibly happened for ~20–30 s, then the note disappeared — and clicking Delete again in the meantime hit the backend with an already-deleted note. The deletion itself is fast (a direct `fs` unlink in the main process), but `useDeleteMeeting` then invalidates the list, which re-runs `list-meetings` — a multi-second cold start of the bundled backend. The deleted note lingered on screen until that refetch landed, looking like nothing happened.

**Change.** Remove the row from the react-query cache optimistically in `onMutate` (with a snapshot for rollback), roll back in `onError`, and reconcile via the background refetch in `onSettled`. The note now disappears immediately on click, and because it's already gone from the list there's no second (failing) delete.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested on macOS

## Additional Notes

Single-file change to `useDeleteMeeting`. The same cold-start-subprocess pattern affects other mutations (reprocess, rename) — those could get the same optimistic treatment in a follow-up if useful.

---
## Summary by cubic
Make note deletion feel instant by removing the note from the `@tanstack/react-query` cache immediately and reconciling via a background refetch. This stops the lingering item and avoids a second (failing) delete during backend cold starts.

- **Bug Fixes**
  - Optimistic removal in `useDeleteMeeting` with `onMutate` + `setQueryData` (no snapshot); cancel in-flight list queries to prevent races.
  - Roll back by re-inserting only the failed meeting in `onError`; reconcile via `onSettled` invalidation.

<sup>Written for commit dca1f392ce46ed39a5f615b533c38f825941615c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/hooks/useMeetings.ts
+++ b/app/renderer/src/hooks/useMeetings.ts
@@ -146,7 +146,36 @@ export function useDeleteMeeting() {
   const qc = useQueryClient();
   return useMutation({
     mutationFn: async (meeting: Meeting) => unwrap(await ipc().meetings.delete(meeting)),
-    onSuccess: () => qc.invalidateQueries({ queryKey: meetingsKeys.all }),
+    // Remove the row from the list immediately. The file deletion itself is
+    // fast (a direct fs unlink in the main process), but the follow-up list
+    // refresh shells out to the bundled backend — a multi-second cold start —
+    // so without this the deleted note lingers on screen, looking like
+    // nothing happened, until the refetch lands. Optimistic removal also
+    // prevents a second delete on the same note (which fails as "already
+    // deleted").
+    onMutate: async (meeting) => {
+      await qc.cancelQueries({ queryKey: meetingsKeys.list() });
+      qc.setQueryData<Meeting[] | undefined>(meetingsKeys.list(), (old) =>
+        old?.filter(
+          (m) => m.session_info.summary_file !== meeting.session_info.summary_file,
+        ),
+      );
+    },
+    onError: (_err, meeting) => {
+      // Re-insert only the meeting that failed to delete, rather than restoring
+      // a whole pre-delete snapshot — a snapshot would clobber a concurrent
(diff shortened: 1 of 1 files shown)


Optic00:
Good catch — fixed. `onError` now re-inserts only the meeting that failed to delete instead of restoring the full pre-delete snapshot, so a concurrent in-flight delete's optimistic removal is no longer clobbered; the `onSettled` refetch reconciles order.

ruzin:
Thanks for this @Optic00 — and sorry to redirect it. We dug into the premise and measured the post-delete `list-meetings` refresh at **~0.4s with ~400 notes** (not the multi-second cold start the rationale assumed), so the optimistic-removal layer is hiding a sub-second blip rather than fixing a felt problem. It also keeps the `onSettled: invalidateQueries`, so the backend re-scan still runs.

The real performance win is to **skip the redundant full re-scan on delete** entirely (remove the one item from cache, don't refetch all notes). I've captured that in #189 — that's where the perf work should go.

Separately, the bigger UX gap is that delete is permanent (it unlinks the summary, transcript **and audio**) behind only a confirm dialog — an Undo / soft-delete would be the higher-value change.

Closing this in favour of #189. Genuinely appreciate the contribution — please do take a swing at #189 if you're up for it.

PR #205, opened 2026-06-15
fix(cloud): show current OpenAI chat models in the picker

### LINKED ISSUE #198: In "Chat" I can only select gpt4o which is an odd default to have and I'd like to be able to select a newer model
Image

Optic00:
Closes #198.

## Problem
The "Test connection" path (`test-cloud-api`) returned `models.data[:10]` straight from OpenAI's `/models` endpoint. That list comes back in an arbitrary order and mixes chat models in with embeddings, speech (whisper/tts), image (dall-e/gpt-image), realtime/audio and moderation models. The hard 10-cap then crowded those non-chat entries into the dropdown and pushed newer chat models (gpt-4.1, the `o` reasoning series, dated gpt-4o variants) off the end — so the Settings model picker often offered little beyond `gpt-4o`.

## Change (backend only)
In the OpenAI/custom branch of `test_cloud_api`:
- **Sort newest-first** (`created` desc) so current models lead the dropdown.
- **For the `openai` provider, keep only chat/reasoning models** — `gpt-*`, the `o<n>` reasoning series, `chatgpt-*`, minus the embeddings/audio/image/moderation/legacy families. `gpt-` and `o\d` are prefix/pattern matches, so future releases (gpt-5, o5, …) surface without another code change.
- **Custom OpenAI-compatible endpoints keep every model** (their naming is unknown) and just get the sort + a wider cap (25 vs 10).

Anthropic (already recent-first) and Bedrock (curated list) are untouched.

## Note
The dropdown already has a "Custom…" free-text escape hatch, so a determined user could always type any model id — but it wasn't discoverable, and the default list is what most people see. This fixes the default list.

## Tests
`tests/test_cloud_model_filter.py` pins the keep/drop decisions and the future-model behaviour of the filter (pure function, no network). Green locally.

---
## Summary by cubic
Fixes the Settings model picker to show current OpenAI chat models instead of 10 arbitrary models. Sorts `/models` newest-first, filters to chat/reasoning families, keeps `*-search-preview`, and drops Responses-only tiers (`*-pro`, `*-deep-research`) so only models that work with `chat.completions` appear. Closes #198.

- **Bug Fixes**
  - Sort by `created` (newest first) and raise the cap from 10 to 25.
  - For `openai`, keep only chat/reasoning models; exclude embeddings/whisper/tts/audio/realtime/moderation/dall-e/image/transcribe/codex; keep `*-search-preview`; drop `*-pro` and `*-deep-research`.
  - For custom OpenAI-compatible endpoints, no filter; apply sort + 25 cap.
  - Add unit tests pinning kept/dropped models, including `*-search-preview` and dated `*-pro`/`*-deep-research` snapshots; Anthropic and Bedrock unchanged.

<sup>Written for commit 9343f32f86a870f26f144d9677f4b2492627b80c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -3890,9 +3890,12 @@ def test_remote_ollama(url):
 # Model families OpenAI's /models endpoint mixes in alongside chat models —
 # embeddings, speech, image and moderation. Excluded so the Settings model
 # picker only offers models that actually answer chat completions (#198).
+# NB: no "search" marker — the *-search-preview models are chat-completion
+# models (web-search-grounded), and "search" is also a substring of
+# "deep-research", so excluding it would drop real chat/reasoning models.
 _OPENAI_NON_CHAT_MARKERS = (
     "embedding", "whisper", "tts", "audio", "realtime",
-    "moderation", "dall-e", "image", "transcribe", "search", "codex",
+    "moderation", "dall-e", "image", "transcribe", "codex",
 )
 
 
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -3887,6 +3887,32 @@ def test_remote_ollama(url):
         print(json.dumps({"success": False, "error": str(e)}))
 
 
+# Model families OpenAI's /models endpoint mixes in alongside chat models —
+# embeddings, speech, image and moderation. Excluded so the Settings model
+# picker only offers models that actually answer chat completions (#198).
+_OPENAI_NON_CHAT_MARKERS = (
+    "embedding", "whisper", "tts", "audio", "realtime",
+    "moderation", "dall-e", "image", "transcribe", "search", "codex",
+)
+
+
+def _is_openai_chat_model(model_id: str) -> bool:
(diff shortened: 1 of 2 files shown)


Optic00:
Tightened the filter: `gpt-5-pro` and the `*-deep-research` tiers are served **only** via the Responses API, not `chat.completions` (which is what Steno calls), so offering them in the picker would 400 at request time. They now drop (a `-pro` segment match + a `deep-research` marker, dated snapshots included) on top of the existing non-chat-family filter; `*-search-preview` chat models are unaffected. Tests cover the new rejections.

PR #206, opened 2026-06-15
feat(ollama): reuse an already-installed supported model instead of re-pulling the default

### LINKED ISSUE #123: feat(ollama): use already-installed local models instead of re-downloading the default
## Background

Reported by @ajitesh1828 in #111. The user had Ollama already installed with at least one model (\`llama3.2\`) pulled. Steno detected the running Ollama, but still issued a fresh \`POST /api/pull {name: "llama3.2:3b"}\` for its hardcoded default tag — downloading a duplicate variant. On their second machine where pulls were restricted, the pull failed and Steno didn't fall back to the model they already had, leaving the app broken even though there was a usable model sitting right there.

The Remote-mode workaround documented at #111#issuecomment-4489785956 lets them point Steno at their own Ollama and pick from installed models manually — but that requires them to know about the distinction and flip a setting. The default-Local-mode flow should "just work" with whatever's already installed.

## Proposed behaviour

In \`src/ollama_manager.py\` / \`simple_recorder.py\`, before issuing a pull for the default model:

1. List installed models on the connected Ollama (\`GET /api/tags\` — Steno already does this elsewhere).
2. If any model in Steno's supported list is already present, use it instead of pulling. Prefer the configured default if present, then fall through the supported list by quality.
3. Only pull when there's genuinely no compatible model installed.

This makes the default Local path no-network-needed for anyone who already has a usable Ollama setup.

## Stretch — first-run AI picker

If we want to go further: a first-run AI setup screen that:

- Detects existing Ollama on 11434.
- Shows its installed models.
- Lets the user pick one (or pull a recommended default if none usable).
- Collapses the Local/Remote distinction in this case — once you're talking HTTP to 11434 the code path is the same.

The Remote-mode option stays useful for cross-machine setups (Ollama on a different host on the LAN).

## Acceptance

- Fresh install + existing Ollama with \`llama3.2:3b\` (or any other supported tag) → no \`/api/pull\` call.
- Fresh install + existing Ollama with no supported model → falls back to current pull behaviour.
- Pull failure (e.g. restricted network) + at least one supported model already installed → uses the installed one, app works.
- No regression for users with no Ollama installed (bundled-Ollama startup + default pull still works).

Optic00:
## What

First-run setup unconditionally issued `POST /api/pull {name: "llama3.2:3b"}` even when the connected Ollama already had a usable supported model. That re-downloads a multi-GB model that's already on disk, and — as reported in #123 — leaves the app broken when the pull is restricted (e.g. locked-down network) despite a perfectly usable model sitting right there.

This makes the default **Local** path no-network-needed for anyone who already has a supported Ollama model installed.

## How

Before pulling, the backend lists installed models (`GET /api/tags` via the `ollama` HTTP API — no shelling to the binary) and matches them against the supported registry, which stays the single source of truth in `config.SUPPORTED_MODELS`. On a hit it sets that model active and **skips the pull entirely**; it only downloads when nothing supported is installed.

Preference order: the configured model → the packaged default (`llama3.2:3b`) → registry order, with deprecated entries chosen only as a last resort (a live model always beats a retired one).

- **`simple_recorder.py`** — `pick_installed_supported_model()` (pure, testable picker) + a `resolve-setup-model` CLI command that reports the already-installed model as JSON.
- **`app/main.js`** — `setup-ollama-and-model` calls it before the pull. Best-effort: any failure (Ollama not ready, parse error) falls through to the existing download, so the worst case is today's behaviour.

## Acceptance criteria (from #123)

- ✅ Existing Ollama with `llama3.2:3b` (or any supported tag) → no `/api/pull`.
- ✅ Existing Ollama with no supported model → falls back to the current pull.
- ✅ Pull-restricted network + a supported model already installed → uses the installed one.
- ✅ No regression for users with no Ollama installed — the bundled-Ollama startup + default pull path is unchanged.

## Testing

- **Unit** (`tests/test_setup_model_reuse.py`, 7 cases): default preference, configured-model preference, registry fall-through, deprecated-as-last-resort, empty → `None`, blank-id handling.
- **e2e T2** (`setup-model-reuse.t2.spec.ts`, model-free): drives `setup.ollamaAndModel()` against the mock Ollama (which lists `llama3.2:3b`) and asserts `pullCalls() === 0` and `config.model === 'llama3.2:3b'`. Verified as a true regression test — it fails when the `main.js` change is reverted. Skips loudly where the bundled `bin/ollama` is absent (CI downloads it), mirroring the existing `@pipeline` skip convention.
- Full backend suite (154 tests) green.

Closes #123

---
## Summary by cubic
First-run setup now reuses an already-installed supported `ollama` model and skips pulling the default. It also fails fast if saving the reused model as active fails, preventing a false success (covers #123).

- **New Features**
  - `app/main.js`: `setup-ollama-and-model` uses `resolve-setup-model` to detect an installed supported model, set it active, and skip `/api/pull` (falls back to pull on any error).
  - `simple_recorder.py`: added `pick_installed_supported_model()` and a `resolve-setup-model` CLI using the `ollama` HTTP API (`/api/tags`), not the binary.
  - Selection order: configured model → packaged default (`gemma4:e2b-it-qat`) → supported registry; deprecated only as a last resort.
  - Tests: unit coverage for the picker; T2 e2e asserts no `/api/pull`; `mock-ollama` now exposes `pullCalls()` and the spec reflects the new default.

- **Bug Fixes**
  - Setup now reports failure if persisting the reused model fails: `set-model` exits non-zero on config-write failure, and the handler checks exit code plus structured output before reporting success.
  - Tests: regression for `set-model` exit codes (non-zero on failure, zero on success).

<sup>Written for commit 96a71a18a9c4b6ad163607acd890a0596cfac6a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -3985,7 +3985,30 @@ ipcMain.handle('setup-ollama-and-model', async () => {
       }
       sendDebugLog('Warning: Ollama may not be fully ready, attempting pull anyway...');
     }
-    
+
+    // #123: don't re-download the default if the connected Ollama already has a
+    // supported model. The backend matches installed models against the supported
+    // registry (single source of truth in config.py); on a hit we set it active
+    // and skip the pull entirely, so the default Local path needs no network for
+    // anyone with a usable Ollama already set up. Best-effort: any failure here
+    // falls through to the normal download below.
+    try {
+      const resolvedRaw = await runPythonScript('simple_recorder.py', ['resolve-setup-model'], true);
+      const resolved = JSON.parse(resolvedRaw.trim());
+      if (resolved && resolved.installed) {
+        sendDebugLog(`Found already-installed model "${resolved.installed}" — skipping download`);
+        try {
+          await runPythonScript('simple_recorder.py', ['set-model', resolved.installed], true);
+        } catch (e) {
+          // Non-fatal -- config write is best-effort, same as the pull path.
+        }
+        trackEvent('setup_completed', { step: 'ollama_existing_model' });
+        return { success: true, message: `Using already-installed model ${resolved.installed}` };
+      }
+    } catch (e) {
+      sendDebugLog(`Could not check for installed models, proceeding to download: ${e.message}`);
+    }
+
     sendDebugLog('Downloading AI model (this may take several minutes)...');
     sendDebugLog(`POST http://127.0.0.1:11434/api/pull {name: "${DEFAULT_AI_MODEL}"}`);
 
--- a/app/main.js
+++ b/app/main.js
@@ -3997,10 +3997,22 @@ ipcMain.handle('setup-ollama-and-model', async () => {
       const resolved = JSON.parse(resolvedRaw.trim());
       if (resolved && resolved.installed) {
         sendDebugLog(`Found already-installed model "${resolved.installed}" — skipping download`);
+        // Persist it as the active model, and only report success once that
+        // write actually succeeded. set-model now exits non-zero on a
+        // config-write failure (runPythonScript rejects) AND prints
+        // success:false, so we check both: swallowing the failure here would
+        // have setup claim success with no active model saved.
         try {
-          await runPythonScript('simple_recorder.py', ['set-model', resolved.installed], true);
+          const setRaw = await runPythonScript('simple_recorder.py', ['set-model', resolved.installed], true);
+          // set-model prints a human line before the JSON, so grab the last
+          // JSON-looking line rather than parsing the whole stdout.
+          const jsonLine = setRaw.trim().split('\n').reverse().find((l) => l.trim().startsWith('{'));
+          const setRes = jsonLine ? JSON.parse(jsonLine) : null;
+          if (!setRes || setRes.success !== true) {
+            return { success: false, error: (setRes && setRes.error) || 'Failed to save the selected model.' };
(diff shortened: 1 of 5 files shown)


Optic00:
Fixed a silent-failure path: `set-model` printed `success:false` on a config-write failure but exited **0**, and the reuse path in `setup-ollama-and-model` swallowed the call — so setup reported success while no active model was persisted. `set-model` now `sys.exit(1)`s on a write failure, and the reuse path checks the structured result **and** the exit code, returning `success:false` instead. Added a regression for the exit code. (The happy-path T2 still skips locally without `bin/ollama`; it runs in CI.)

ruzin (approved):
Reviewed: rebased onto current main, re-ran full CI (green), code-review clean, verified locally. Closes the linked issue.

PR #210, opened 2026-06-17
fix(ci): bundle arm64 ffmpeg in the arm64 mac build (fixes #209)

### LINKED ISSUE #209: Bundled ffmpeg is x86_64 in the arm64 macOS build → transcription fails with `OSError: [Errno 86] Bad CPU type` on Apple Silicon without Rosetta 2
**Summary**

The arm64 macOS build ships an **x86_64** `ffmpeg`. On Apple Silicon Macs **without Rosetta 2**, every recording fails to process — the transcription backend crashes the moment it shells out to the bundled ffmpeg:

```
OSError: [Errno 86] Bad CPU type in executable:
'/Applications/Steno.app/Contents/Resources/stenoai/_internal/ffmpeg'
```

The bundled `ffmpeg` is x86_64 while the rest of the bundle is correctly arm64, so this is specifically the bundled ffmpeg being the wrong architecture — a packaging/CI defect where the arm64 mac build embeds an Intel ffmpeg.

**Rosetta is what hides this.** On Apple Silicon machines that have Rosetta 2 installed, the x86_64 ffmpeg is emulated transparently and everything works — which is why most M-series users (and the working-ffmpeg case in #207) never see it. It only **hard-fails** with `Errno 86 Bad CPU type` on Apple Silicon machines **without** Rosetta. So the bug has likely been in the build for a while, masked by Rosetta on nearly every dev machine.

In the UI this surfaces as the recording getting stuck on **"Processing" / "analyzing transcript" forever**, then **"Couldn't process this recording."** Multiple users hit this (reported by two people in Discord), all on Apple Silicon without Rosetta.

**This is distinct from #207.** #207 is a non-reproducible empty-transcript transient on a machine where ffmpeg works fine ("Bundled ffmpeg decodes it fully"). This bug is the opposite: ffmpeg never runs at all, and it is **100% deterministic** — every attempt crashes identically before any transcription happens.

**What happens (from a user log, Steno latest / auto-updater "up to date", Apple M2)**

Three consecutive recordings, three identical failures:

```
HEARTBEAT:transcribe:start
...
INFO - whisper.cpp model loaded (threads=6)
Traceback (most recent call last):
  File "simple_recorder.py", line 376, in transcribe_audio
  File "src/transcriber.py", line 438, in __init__
  File "src/transcriber.py", line 530, in _ensure_ffmpeg_in_path
  File "subprocess.py", line 1955, in _execute_child
OSError: [Errno 86] Bad CPU type in executable:
'/Applications/Steno.app/Contents/Resources/stenoai/_internal/ffmpeg'
[PYI-521:ERROR] Failed to execute script 'simple_recorder' due to unhandled exception!
📝 Transcribing: sysaudio-...-Note.webm
Preserved audio after processing failure: .../recordings/sysaudio-...-Note.webm
```

Observations from the log:

- **Only ffmpeg is the wrong arch.** The same process loads `ggml-large-v3-turbo` and initialises Metal on `Apple M2` successfully (`ggml_metal_init: found device: Apple M2`, `whisper_backend_init_gpu: using MTL0 backend`). The PyInstaller sidecar itself is arm64. The crash is isolated to the `ffmpeg` subprocess exec.
- **On the latest release.** `Auto-updater: up to date` — the currently shipping arm64 build contains the broken ffmpeg; this is not a stale/mismatched install.
- **Deterministic.** 3/3 recordings failed identically. Not a transient.
- **Audio is preserved** (`Preserved audio after processing failure`), so affected users can reprocess once a fixed build ships — no data loss.
- Engine was `whisper` here, but the ffmpeg path is engine-independent, so Parakeet users on the same build are affected too.

**Root cause (confirmed)**

Confirmed directly from the official release artifact — pulled `_internal/ffmpeg` straight out of `Steno-0.5.4-arm64-mac.zip`:

```
Steno.app/Contents/Resources/stenoai/_internal/ffmpeg : Mach-O 64-bit executable x86_64
Steno.app/Contents/MacOS/Steno                        : Mach-O 64-bit executable arm64
_internal/_pywhispercpp….so                           : Mach-O 64-bit bundle arm64
```

So the arm64 release bundles an Intel-only ffmpeg (not a fat binary), while everything around it is arm64. Likely an asset-download / lipo / per-arch packaging step in CI that fetches or copies the Intel ffmpeg into the arm64 build. (Context: per-arch mac packaging has been touched before — e.g. #110, and bundled-ffmpeg resolution in #35 / #52.)

**Verification**

Anyone hitting this can confirm it on their own machine in a few seconds. Open Terminal and paste:

```bash
# 1) Are you on Apple Silicon? -> expect: arm64
uname -m

# 2) What architecture is the bundled ffmpeg? -> the bug is: x86_64
FF="/Applications/Steno.app/Contents/Resources/stenoai/_internal/ffmpeg"
file "$FF"
lipo -archs "$FF" 2>/dev/null || echo "(single-arch binary)"

# 3) Does it even run on this machine? -> the bug: "Bad CPU type in executable"
"$FF" -version >/dev/null 2>&1 && echo "ffmpeg runs OK" || echo "ffmpeg FAILS to exec on this CPU"
```

Note on step 3: if `uname -m` says `arm64` but step 3 still prints `ffmpeg runs OK`, you have **Rosetta 2 installed** and it's emulating the x86_64 binary — the build is still buggy, Rosetta is just hiding it (`softwareupdate --install-rosetta` present, `pgrep oahd` running). The build is confirmed broken whenever `file` / `lipo -archs` report `x86_64` regardless of step 3. The hard `Bad CPU type` crash hits only on machines without Rosetta.

Temporary workaround for affected users (until a fixed build ships):
- **Simplest:** install Rosetta 2 — `softwareupdate --install-rosetta --agree-to-license`. The existing x86_64 ffmpeg then runs under emulation; survives auto-updates; no touching the signed bundle.
- **Native:** replace the bundled binary with an arm64 ffmpeg (`brew install ffmpeg`, copy `$(which ffmpeg)` over the path above). Touches the signed app bundle and is overwritten by the next auto-update — stopgap only.

In CI, assert the bundled ffmpeg's arch matches the build target before packaging (e.g. `lipo -archs` must contain the target arch), so an arch mismatch fails the build instead of shipping.

**Proposed fix**

1. Fix packaging so the arm64 build bundles an arm64 (or universal) ffmpeg; add the CI arch-assertion above as a guard.
2. **Preflight the ffmpeg check before loading the model.** `_ensure_ffmpeg_in_path` (`src/transcriber.py:530`) currently runs *after* the ~1.6 GB whisper model is loaded (`whisper.cpp model loaded` prints immediately before the traceback). Validate ffmpeg first so a bad/missing binary fails fast instead of after a full model load on every attempt.
3. **Surface a real error.** Catch the exec failure and show an actionable message ("bundled ffmpeg is the wrong architecture / not runnable") instead of an indefinite "Processing" → generic "Couldn't process this recording", which sent users down the mic/permissions path. (Persistent logs from #208 would also have made this self-diagnosable.)

**Environment**

macOS, Apple Silicon, Steno v0.5.4 (auto-updater up to date), engine=whisper (`large-v3-turbo`). Hard crash on M2 without Rosetta 2; the same x86_64 ffmpeg from the same build runs fine on an M4 with Rosetta. Reproducible across recordings; second reporter on the same symptom.

Optic00:
## What

The arm64 macOS build bundles an **x86_64** `ffmpeg`. On Apple Silicon **without Rosetta 2**, the transcription backend crashes the moment it execs ffmpeg:

```
OSError: [Errno 86] Bad CPU type in executable:
'/Applications/Steno.app/Contents/Resources/stenoai/_internal/ffmpeg'
```

So every recording fails before transcription starts ("Processing" forever → "Couldn't process this recording"). Fixes #209.

## Root cause

`scripts/download-ollama.sh` selected the ffmpeg download by **OS (`uname -s`), not architecture**, and the Darwin branch pulled `evermeet.cx`, which publishes **x86_64-only** mac builds. The release job runs on the `macos-14` arm64 runner and PyInstaller builds an arm64 backend, but this step dropped an Intel ffmpeg into `bin/`, and `stenoai.spec` placed it into the bundle with no arch check.

It slipped through because **Rosetta 2 masks it** — present on GitHub's macOS runners and basically every dev machine, so the x86_64 ffmpeg runs there. It only hard-fails on Apple Silicon without Rosetta. (The `@pipeline` e2e also uses whisper in CI, not the arch-sensitive path.)

Confirmed by pulling `_internal/ffmpeg` straight out of the official `Steno-0.5.4-arm64-mac.zip`: `Mach-O 64-bit executable x86_64`, while the Electron main binary and the whisper `.so` in the same bundle are arm64.

## Change

- **`scripts/download-ollama.sh`**: the mac build is Apple-Silicon-only (arm64) since v0.4.0, so fetch the **arm64** static ffmpeg (same version, 7.1.1) from osxexperts and refuse any other arch, instead of the Intel-only evermeet build.
- **`.github/workflows/build-release.yml`**: assert `file bin/ffmpeg` is arm64 after download, mirroring the existing mic-monitor arch guard — an arch mismatch now fails the build instead of shipping.

## Verification

Locally ran the patched Darwin branch on arm64: downloads → `Mach-O 64-bit executable arm64`, `ffmpeg version 7.1.1`, the CI assertion passes. The previously-bundled x86_64 binary reproduced `Errno 86` on an M2 without Rosetta; this arm64 binary runs natively (no Rosetta needed).

Workaround for users on an affected build until this ships: `softwareupdate --install-rosetta --agree-to-license`. Preserved recordings can then be reprocessed.

---
## Summary by cubic
Bundles the correct arm64 `ffmpeg` in the macOS arm64 build to prevent crashes on Apple Silicon without Rosetta. Adds CI and build-time validation to prevent shipping a mismatched or corrupt binary. Fixes #209.

- **Bug Fixes**
  - macOS: enforce arm64 builds and download the arm64 static `ffmpeg` 7.1.1 from osxexperts; refuse non-arm64.
  - macOS: after extraction, assert the binary is Mach-O arm64 and run `ffmpeg -version` with pipefail; require 7.1.x to catch corrupt/wrong downloads.
  - CI: verify `bin/ffmpeg` is arm64 in `build-release.yml` to fail fast on arch mismatches.

<sup>Written for commit e737b30e266011d469927ab6b3ed9d4d362abd89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -100,6 +100,11 @@ jobs:
       run: |
         chmod +x scripts/download-ollama.sh
         ./scripts/download-ollama.sh
+        # Fail loudly if the bundled ffmpeg isn't arm64 — an x86_64 ffmpeg in the
+        # arm64 build crashes on Apple Silicon without Rosetta (#209). Mirrors the
+        # mic-monitor arch guard below.
+        file bin/ffmpeg
+        file bin/ffmpeg | grep -q "arm64"
 
     - name: Build mic-monitor helper (Swift)
       run: |
--- a/scripts/download-ollama.sh
+++ b/scripts/download-ollama.sh
@@ -11,14 +11,25 @@ BIN_DIR="$(cd "$(dirname "$0")/.." && pwd)/bin"
 echo "=== Downloading ffmpeg ==="
 case "$(uname -s)" in
     Darwin)
-        FFMPEG_URL="https://evermeet.cx/ffmpeg/ffmpeg-7.1.1.zip"
+        # ffmpeg must match the BUILD architecture, not just the OS. evermeet.cx
+        # (the old URL) ships x86_64-only mac builds, so it bundled an Intel ffmpeg
+        # into the arm64 release — which crashes on Apple Silicon without Rosetta
+        # (#209). The mac build is Apple-Silicon only (arm64) since v0.4.0, so use
+        # osxexperts' arm64 static build (same 7.1.1 as before) and refuse any
+        # other arch rather than silently shipping a mismatch.
+        if [ "$(uname -m)" != "arm64" ]; then
+            echo "macOS build is arm64-only; unsupported arch: $(uname -m)" >&2
+            exit 1
+        fi
+        FFMPEG_URL="https://www.osxexperts.net/ffmpeg711arm.zip"
         mkdir -p "$BIN_DIR"
         curl -L "$FFMPEG_URL" -o "$BIN_DIR/ffmpeg.zip"
         cd "$BIN_DIR"
-        unzip -o ffmpeg.zip
+        # Extract only the binary; skip the __MACOSX resource-fork junk in the zip.
+        unzip -o ffmpeg.zip ffmpeg
         rm ffmpeg.zip
         chmod +x ffmpeg
-        echo "ffmpeg downloaded"
+        echo "ffmpeg 7.1.1 (arm64) downloaded"
         cd - > /dev/null
         ;;
     Linux)
--- a/scripts/download-ollama.sh
+++ b/scripts/download-ollama.sh
@@ -29,11 +29,21 @@ case "$(uname -s)" in
         unzip -o ffmpeg.zip ffmpeg
         rm ffmpeg.zip
         chmod +x ffmpeg
-        # Validate the binary actually runs and is the expected major version —
-        # catches a truncated/corrupt download or a wrong-format extract before it
-        # gets bundled (complements the `file ... arm64` arch guard in CI). set -e
-        # makes a non-zero exit here fail the build loudly.
-        if ! ./ffmpeg -version | grep -q "ffmpeg version 7.1"; then
+        # Validate the binary before bundling it. Two distinct failure modes:
+        #  1. Wrong architecture. An x86_64 ffmpeg runs fine HERE under Rosetta and
+        #     would sail through the -version check, then crash on a Rosetta-less
+        #     user machine (#209). Assert the Mach-O is arm64 so the script
+        #     self-enforces the arch rather than leaning only on the external CI
+        #     `file ... arm64` guard.
+        #  2. Truncated/corrupt download or wrong-format extract. Run -version and
+        #     pin the major; pipefail (scoped to a subshell) makes a non-zero
+        #     ffmpeg exit fail loudly instead of being masked by grep's exit.
+        # set -e turns either non-zero exit into a loud build failure.
+        if ! file ./ffmpeg | grep -q "arm64"; then
+            echo "Downloaded ffmpeg is not an arm64 binary: $(file ./ffmpeg)" >&2
+            exit 1
+        fi
+        if ! ( set -o pipefail; ./ffmpeg -version | grep -q "ffmpeg version 7.1" ); then
             echo "Downloaded ffmpeg failed -version or is not 7.1.x" >&2
             exit 1
         fi
--- a/scripts/download-ollama.sh
+++ b/scripts/download-ollama.sh
@@ -29,6 +29,14 @@ case "$(uname -s)" in
         unzip -o ffmpeg.zip ffmpeg
         rm ffmpeg.zip
         chmod +x ffmpeg
+        # Validate the binary actually runs and is the expected major version —
+        # catches a truncated/corrupt download or a wrong-format extract before it
+        # gets bundled (complements the `file ... arm64` arch guard in CI). set -e
+        # makes a non-zero exit here fail the build loudly.
+        if ! ./ffmpeg -version | grep -q "ffmpeg version 7.1"; then
+            echo "Downloaded ffmpeg failed -version or is not 7.1.x" >&2
+            exit 1
+        fi
         echo "ffmpeg 7.1.1 (arm64) downloaded"
         cd - > /dev/null
         ;;


Optic00:
Small optional hardening: after extracting the arm64 ffmpeg, the script now runs it with `-version` and asserts the `7.1` major before bundling, so a truncated/corrupt download or wrong-format extract fails the build loudly (complements the arch guard). The #209 arch fix is unchanged. I deliberately did **not** pin a SHA-256 of the zip — that would hard-wire a trust-on-first-use hash from a third-party host into the release pipeline (and break CI on any upstream rebuild); happy to add it if you want to anchor a hash you've verified.

Optic00:
Addressed both P2s identified by cubic in the ffmpeg validation (\`e737b30\`).

1. **Arch not self-enforced.** cubic noted an x86_64 ffmpeg runs under Rosetta on the CI host and would pass the \`-version\` check, only to crash on a Rosetta-less Apple Silicon machine (#209). Added a \`file ./ffmpeg | grep -q arm64\` assertion so the script self-enforces the architecture rather than relying solely on the external CI \`file ... arm64\` guard.
2. **grep masking the pipeline exit.** Wrapped the version check in a \`( set -o pipefail; ./ffmpeg -version | grep -q ... )\` subshell so a non-zero \`ffmpeg\` exit fails loudly instead of being swallowed by grep's status. Scoped to the subshell so the rest of the script keeps its existing semantics.

Both verified locally against the real bundled arm64 binary (arch + version checks pass; \`bash -n\` clean).

ruzin (approved):

PR #236, opened 2026-06-21
fix: agree on the data dir in dev so import dedup works (#233)

### LINKED ISSUE #233: Dev-mode data-dir mismatch: dev JS uses repo dirs while the spawned frozen backend uses ~/Library (breaks import-collision dedup in npm start)
## Summary

In **dev (`npm start`)**, importing a file with the same basename as an existing note **overwrites** that note instead of de-duplicating to `<stem>-1`. The shipped (packaged) app is **not** affected — this is a dev-harness data-directory mismatch.

## Root cause

There are two parallel dev/prod data-dir splits that are supposed to mirror each other:

- **JS** (`app/main.js` `resolveRecordingsDir()` / `getUserDataDir()`): dev (`app.isPackaged === false`) → `REPO/recordings` + `REPO/output`; packaged → `~/Library/Application Support/stenoai/{recordings,output}`.
- **Python** (`src/config.get_data_dirs()` / `get_user_data_dir()`): run from source → `REPO/...`; **frozen PyInstaller bundle** → `~/Library/.../...`.

The dev app runs **dev-JS (repo dirs)** but, per `getBackendPath()`'s `dist/stenoai/stenoai` fallback, spawns the **frozen bundle** as its backend — and the frozen bundle resolves dirs as **prod** (`~/Library/...`). So in dev the two layers disagree:

- `copyImportIntoRecordings()` (PR #185) computes its collision dir as `dirname(resolveRecordingsDir())/output` = `REPO/output`, checks `noteExists()` there (always empty), and never bumps.
- The bundled Python writes the summary to `~/Library/.../output/<stem>_summary.md` — overwriting the previous import.

Evidence from a real repro: after two imports of `steno-import-test.wav`, `REPO/output` is empty (where JS looked) while `~/Library/.../output/steno-import-test_summary.md` is a single, overwritten file; `dist/stenoai/stenoai status` lists the `~/Library` recordings, but `get_data_dirs()` from source returns `REPO/...`.

## Why production is unaffected

In the packaged app, JS takes the `app.isPackaged` branch → `getUserDataDir()/output` = `~/Library/.../output`, exactly where the frozen bundle writes. JS and Python agree, so the PR #185 collision guard correctly bumps a same-basename re-import to `<stem>-1` and never overwrites.

## Scope / impact

- **User-facing:** none in the shipped DMG. Affects contributors testing imports (and any path-agreement-dependent feature) via `npm start`.
- Surfaced by PR #185 (audio import), which is the first feature to require JS and the backend to agree on the data dir at runtime.

## Suggested fixes (later)

1. Make the dev app's JS and its spawned backend agree on the data dir. Options: have dev-JS honor `getUserDataDir()` (so it doesn't diverge from the frozen bundle it spawns), or run the backend from source in dev so both use repo dirs, or set `STENOAI_USER_DATA_DIR` for the dev launch so both layers honor it.
2. **e2e blind spot:** `e2e/specs/audio-import-collision.t2.spec.ts` seeds the "existing note" directly in the JS-computed output dir and only asserts the JS guard bumps — it never drives the full Python round-trip, so it cannot catch a JS↔Python directory disagreement. Add a regression test that imports a file, lets the real backend write the summary, then re-imports the same basename and asserts a `<stem>-1` note exists and the first is intact.

## Not a blocker for

PR #185 — the feature is correct in the shipped app. This issue tracks the dev-harness divergence + the e2e gap for a follow-up.

Optic00:
Fixes #233.

## Problem
In dev (`npm start`) the JS layer and the spawned **frozen** Python backend disagree on the data dir. Both JS recordings-dir resolvers (`resolveRecordingsDir` and the `get-recordings-dir` handler) hardcoded `REPO/recordings`, while the frozen bundle writes notes under `~/Library` (`is_bundled()`). So `copyImportIntoRecordings`'s collision check (#185) consulted `REPO/output` — always empty in dev — and a same-basename re-import silently overwrote the first note instead of bumping to `<stem>-1`. Production was unaffected (packaged JS already used `getUserDataDir()`).

## Fix
Route both resolvers through `getUserDataDir()` for the non-custom case, so JS agrees with the backend's `get_data_dirs()` in every mode:

| Mode | recordings/output root | Backend | Agree? |
|---|---|---|---|
| Dev, no custom storage | `~/Library/.../stenoai` | `~/Library` (is_bundled) | yes |
| Dev, custom storage | `CUSTOM` | `CUSTOM` | yes |
| E2E (override = temp) | `temp` | `temp` | yes + isolated |
| Packaged | `~/Library` | `~/Library` | yes (unchanged) |

Also guards custom storage with `!STENOAI_USER_DATA_DIR` so the e2e isolation override beats a configured `storage_path`, mirroring `get_data_dirs()` precedence (config.py). **No Python change, no new env var.**

Behavioral note: in dev, imports now copy into `~/Library/.../stenoai/recordings` (where the frozen backend already writes), instead of the repo-root scratch. `config.json`/credentials are unaffected (already resolve to `~/Library` in dev).

## Tests
- **New `@pipeline` round-trip regression** (`audio-import-collision-roundtrip.t2`): drives the real backend twice with the same basename and asserts the re-import is deduped to `<stem>-1` with the first note byte-for-byte intact. Verified RED on the pre-fix resolvers (the second note never appears → first is overwritten), green after.
- The existing model-free collision specs become hermetic now that `getDir()` honors the override (was a documented "pre-existing quirk"). All green; `save-notes-location`/`settings-roundtrip` (custom storage) green too.

## Out of scope (spotted while here, not fixed)
- `process-system-audio-recording` looks for a notes file under `getBackendCwd()/_internal/output` (main.js ~6184) while notes are written to the canonical output dir — system-audio processing can miss a user notes file. Unverified; possibly a separate bug.
- `setup-system-check` still has the packaged/dev `REPO` split (main.js ~3300), pre-creating empty repo-root dirs in dev. Runtime-harmless but could use `getUserDataDir()`.

---
## Summary by cubic
Align dev data directories between the JS app and the frozen Python backend so import dedup works. Prevents same-basename re-imports from overwriting the first note in dev and `@pipeline` runs.

- **Bug Fixes**
  - Both recordings-dir resolvers now use getUserDataDir() for non-custom paths, matching the backend’s get_data_dirs() in dev, e2e, and packaged modes.
  - Custom storage is ignored when STENOAI_USER_DATA_DIR is set, mirroring backend precedence and ensuring e2e isolation.
  - Dev imports now copy into the OS user data dir (e.g., ~/Library/.../stenoai/recordings), not the repo root.
  - Added `@pipeline` round-trip regression spec to verify dedup bumps to "<stem>-1" and keeps the first note intact. No Python changes or new env vars.

<sup>Written for commit ea36b0a7a0704d7f161d14f0e72c86cfeddb5cda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -636,14 +636,19 @@ function getAllowedBaseDirs() {
 // handler that can't `await runPythonScript`.
 function resolveRecordingsDir() {
   let dir;
-  if (_cachedCustomStoragePath) {
+  if (_cachedCustomStoragePath && !process.env.STENOAI_USER_DATA_DIR) {
     dir = path.join(_cachedCustomStoragePath, 'recordings');
-  } else if (app.isPackaged) {
-    // getUserDataDir() (not a macOS literal) so the packaged path is correct on
-    // Windows (%APPDATA%/stenoai) too; identical on macOS.
-    dir = path.join(getUserDataDir(), 'recordings');
   } else {
-    dir = path.join(__dirname, '..', 'recordings');
+    // getUserDataDir() resolves the per-OS data dir when packaged (Windows
+    // %APPDATA%/stenoai; identical on macOS) AND honors STENOAI_USER_DATA_DIR
+    // (the e2e temp dir) — so JS agrees with the backend's get_data_dirs() in
+    // every mode. The old hardcoded REPO/recordings dev branch ignored both,
+    // so the dev app (and e2e) disagreed with the frozen backend, which writes
+    // notes under ~/Library even in dev — that mismatch broke import-collision
+    // dedup in `npm start` (#233). The STENOAI_USER_DATA_DIR guard above mirrors
+    // get_data_dirs()'s precedence (config.py): the e2e isolation override beats
+    // a configured custom storage path, so a test can never escape the temp dir.
+    dir = path.join(getUserDataDir(), 'recordings');
   }
   fs.mkdirSync(dir, { recursive: true });
   return dir;
@@ -5990,13 +5995,17 @@ ipcMain.handle('get-recordings-dir', async () => {
     const jsonData = JSON.parse(result.trim());
 
     let recordingsDir;
-    if (jsonData.storage_path) {
+    if (jsonData.storage_path && !process.env.STENOAI_USER_DATA_DIR) {
       recordingsDir = path.join(jsonData.storage_path, 'recordings');
-    } else if (app.isPackaged) {
-      // getUserDataDir() for the correct Windows path (%APPDATA%); identical on macOS.
-      recordingsDir = path.join(getUserDataDir(), 'recordings');
     } else {
-      recordingsDir = path.join(__dirname, '..', 'recordings');
+      // getUserDataDir() resolves the per-OS data dir when packaged (Windows
+      // %APPDATA%; identical on macOS) AND honors STENOAI_USER_DATA_DIR (the
+      // e2e temp dir). The old hardcoded REPO/recordings dev branch disagreed
+      // with the frozen backend (which writes under ~/Library even in dev),
+      // breaking import-collision dedup in `npm start` (#233). The override
+      // guard mirrors get_data_dirs() precedence (config.py): e2e isolation
+      // beats custom storage. Mirrors resolveRecordingsDir().
+      recordingsDir = path.join(getUserDataDir(), 'recordings');
     }
 
     // Ensure directory exists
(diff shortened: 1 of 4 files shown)


ruzin (approved):
Batch review: core fix + regression test verified correct, RED-before/GREEN-after confirmed, no cross-platform or packaged-path regression. Merging.

Issue #188, opened 2026-06-13
Map-reduce summarization for long meetings (no-loss alternative to head/tail capping)


Optic00:
## Context

PR #177 fixes the long-transcript context-overflow problem by setting `num_ctx=8192` and, for transcripts over ~24k chars, **capping to head 60% + tail 40%** with a truncation note. That's a solid, RAM-conscious fix — but as the PR itself notes, it's a compromise: **the middle of a long meeting is dropped.** For a 2-hour meeting where topics aren't strictly head/tail (e.g. a decision discussed an hour in), the capped summary can miss real content.

This issue proposes a **no-loss** path for the long case, complementary to #177 rather than replacing it.

## Proposal: adaptive single-pass → map-reduce

Keep the simple, high-quality path for meetings that fit, and switch to chunked map-reduce only when they don't:

- **Fits in a RAM-safe context** (short/medium): one pass, as today. Best quality, simplest.
- **Too long**: **map-reduce** instead of capping:
  1. **Map** — split the transcript into fixed, slightly-overlapping windows (e.g. ~10–15 min each) and summarize each window. Critically, extract **structured** elements per window (key points / decisions / action items as lists), not prose — so the meeting essentials are preserved even mid-transcript.
  2. **Reduce** — feed all window results into one final pass that **dedupes and merges** them into the final summary.

### Why map-reduce (and not topic segmentation or pre-cleaning)

- **It handles non-linear meetings.** If a topic recurs in window 1 and window 5, both windows capture it and the reduce step — which sees all window summaries together — merges them into one coherent point. Topics aren't fragmented.
- **Bounded RAM.** Each call uses a small per-window context, so `num_ctx` can stay modest (in the spirit of #177's 8192) regardless of total meeting length. This also sidesteps the per-model context concern #177 raises as its planned follow-up.
- **Scales to any length** — a 4-hour meeting works the same way, where a single big context eventually can't.
- **Topic-segmentation** (split by detected topic) is more fragile: boundary detection is hard and non-linear meetings break it.
- **Pre-cleaning/condensing** the transcript first is lossy (it drops content — the opposite of the goal) and, if done with an LLM, faces the same context limit.

### Pairs naturally with a RAM/length heads-up

When a long transcript meets a large configured model on a memory-constrained machine, warn and/or suggest a lighter model — the map-reduce path keeps each call small, so the warning becomes "this will take a few passes" rather than "this may fail / lose content."

## Relationship to #177

[...]

PR #246, opened 2026-06-22
fix: surface context-overflow as a clear error instead of silent empty summary

### LINKED ISSUE #245: Long meetings silently fail with "No summary available" (formatting instructions truncated by model context cap)
## Summary

Meetings that produce transcripts exceeding the model's context window result in **"No summary available"** in the UI. The transcript fills the context window, the formatting instructions at the front of the prompt are truncated, and the model emits the raw transcript instead of a structured summary. The failure is silent — no error is shown, a title is successfully generated, and a broken `_summary.md` is written to disk.

## Environment

- **stenoai:** v0.5.5
- **Hardware:** MacBook Air M1, 16 GB RAM
- **Model:** `llama3.2:3b` · **Provider:** Ollama · **`num_ctx`:** 8,192 tokens (entire request — instructions + transcript combined)
- **Meeting:** 57 min · **Transcript:** 60,372 chars ≈ ~15,000 tokens (≈ 2× the context cap)
- **Scope:** Affects existing installs still using `llama3.2:3b`. New installs since ~2026-06-19 (PR #218 / v0.5.5) default to Gemma 4 (32k context) and are largely unaffected.

## Steps to Reproduce

1. Use `llama3.2:3b` as the summarization model (Ollama).
2. Record a meeting long enough that `transcript_tokens + instruction_tokens > 8192` (roughly 50+ min for an active meeting).
3. Let processing complete.
4. Open the meeting in the UI.

## Causal Chain

```
Context overflow (15k tokens > 8k cap)
  → Ollama truncates from the front of the prompt
  → Formatting instructions + ## Summary template are cut
  → Model sees only the raw transcript tail, no output format guidance
  → Model emits a partial fragment + echoes the transcript under ## Transcript
  → _parse_meeting_markdown() finds no ## Summary section → summary = ""
  → UI renders "No summary available for this meeting."
```

## Evidence

**Saved `_summary.md`** (truncated for brevity):
```
---
title: "Discussing Proposed Changes to Team Structure"
date: "2026-06-22T10:35:49.281155"
duration_seconds: 3431
language: "en"
is_diarised: true
---

Das

## Transcript

[You] Guten Morgen. So.
[Others] So, moin.
...
```

"Das" (German: "The/That") is the fragment the model began before running out of context. It then copied the raw transcript under its own `## Transcript` header.

**Log (relevant excerpt):**
```
[08:33:48] INFO  - Transcript: 60372 chars
[08:33:48] INFO  - Using configured model: llama3.2:3b
[08:33:48] INFO  - Starting streaming summary with local model: llama3.2:3b
[08:34:48] HEARTBEAT:summarize:1            ← only 1 heartbeat (60 s interval, slow prompt-eval)
[08:35:47] HTTP Request: POST .../api/chat "HTTP/1.1 200 OK"
[08:35:47] STREAM_COMPLETE
[08:35:49] INFO  - Generated meeting title: Discussing Proposed Changes to Team Structure
[08:35:49] SAVED:/Users/.../output/sysaudio-...-Note_summary.md
```

The title was generated correctly (second LLM call on a short input), masking the failure of the first call.

## Workaround

Switch the summarization model to Gemma 4 or Qwen (32k context), then reprocess the meeting via the UI. The existing transcript will be re-summarized with the larger context window.

## Proposed Fixes

**a) Pre-flight token budgeting (short-term)**
Estimate `transcript_tokens + instruction_tokens` against the active model's `num_ctx` before the call. If it won't fit, fail visibly with an actionable message ("Meeting too long for current model — switch to a larger-context model or wait for map-reduce support") instead of producing a broken summary.

**b) Validation before save (guardrail — improves failure visibility, does not solve the token limit)**
After generation, check the output for the expected `## Summary` header. If absent, do **not** persist as success — set an explicit error state. Prevents silent/empty summaries from being stored as if they succeeded. This is error-visibility hardening, not a content fix.

## Rejected Alternative

> **"Reverse the prompt layout: transcript first, instructions last"**

This preserves the formatting instructions by silently truncating the *beginning* of the meeting instead. Output looks correct but is a summary of only the final minutes. Silent wrong output is worse than a visible failure.

## Long-term Fix

Map-reduce summarization: split into overlapping windows → summarize each → merge. Handles transcripts of arbitrary length regardless of context cap. Proposed in **#188**.

## Affected Code

| File | Location | Role |
|------|----------|------|
| `src/summarizer.py` | `resolve_num_ctx` | Hard-codes `num_ctx=8192` for `llama3.2:3b` |
| `src/summarizer.py` | `_create_markdown_prompt` | Instructions-first layout causes front-truncation loss |
| `simple_recorder.py` | save path | Saves output unconditionally without checking for `## Summary` |
| `simple_recorder.py` | `_parse_meeting_markdown` | `sections.get('summary', '')` → empty without header |
| `app/renderer/src/routes/MeetingDetail.tsx` | ~line 642 | Renders "No summary available" on empty summary |

## Acceptance Criteria

- [ ] Context overflow is detected and surfaced to the user before or immediately after the LLM call
- [ ] No silent empty/broken summary is persisted as a success state
- [ ] Regression test: transcript exceeding model context cap triggers a clear error, not "No summary available"
- [ ] Map-reduce path tracked separately in #188

Optic00:
## Problem

Closes #245.

When a transcript exceeds the model's context window (reproducible: `llama3.2:3b` at 8,192 tokens, 57-min meeting → ~15,000 transcript tokens), Ollama silently truncates the formatting instructions from the front of the prompt. The model echoes the raw transcript without a `## Summary` header. `_parse_meeting_markdown` reads this as an empty summary and the UI shows **"No summary available"** — while a hallucinated title is generated and the broken file is saved as a success.

## What this PR does

Two guards, no protocol changes, no new dependencies:

**1. Pre-flight token check** (`src/summarizer.py`, `summarize_transcript_streaming`)
Before the Ollama API call: estimates `len(transcript) / 4` against `num_ctx * 0.8` and raises `ValueError` with an actionable message if the transcript won't fit. Ollama-only (`local`/`remote` providers) — cloud adapters manage their own larger context.

**2. Post-stream header validation** (`simple_recorder.py`, `process_streaming`)
After streaming completes: checks for a `## Summary` header using the same `startswith` convention as `_parse_streamed_markdown` / `_parse_meeting_markdown`. If missing:
- Writes a debug sidecar `*_summary.invalid.md` (not `*_summary.md` — the renderer never sees it, so no phantom meeting is created)
- Emits `ERROR:` + exits with code 1

This backstops edge cases where the pre-flight estimate misses (unknown model, unusual tokenisation ratio).

The outer `asyncio.run()` call is wrapped so `ValueError` from the pre-flight check surfaces as `ERROR: <msg>` + exit 1, matching the existing failure contract. `SystemExit` is re-raised unchanged so the post-stream guard's own message isn't swallowed.

## Before / after

**Before:** `STREAM_COMPLETE` → hallucinated title → `SAVED:` → UI shows "No summary available"

**After (pre-flight):** `ERROR: Transcript is too long for this model's context window (~15,088 estimated tokens vs a 8,192-token window for llama3.2:3b). Use a model with a larger context window or a shorter recording.`

## What this PR does NOT do

No summary is generated for over-long meetings — that requires map-reduce chunking, tracked in #188. This PR only makes the failure visible and prevents the hallucinated-title / phantom-success state.

## Test plan

- [ ] Short meeting (<50% of model context): happy path unchanged, `SAVED:` emitted as before
- [ ] Long meeting with `llama3.2:3b` (>~26k chars transcript): `ERROR:` emitted before API call, no meeting entry created
- [ ] Manually set `num_ctx` above transcript length: pre-flight passes, post-stream header check acts as backstop if model still returns no `## Summary`
- [ ] Existing `summarize-contract.t2` e2e spec passes (no protocol regression)

---
## Summary by cubic
Show a clear error when transcripts + notes exceed the model’s context window to prevent empty summaries and “phantom” saved meetings. Fixes #245.

- **Bug Fixes**
  - Add pre-flight context check for `local`/`remote` Ollama in `src/summarizer.py` (streaming and non-streaming): estimate `(len(transcript)+len(notes))/4` vs `num_ctx*0.8` and fail with a clear ERROR before the API call if too long.
  - Validate streamed output in `simple_recorder.py`: require a `## Summary` header; if missing, write `*_summary.invalid.md`, print ERROR, and exit 1; wrap `asyncio.run()` to surface failures as a single ERROR line and preserve `SystemExit`.
  - Update `e2e/fixtures/mock-ollama.js` to default to a minimal valid summary with a `## Summary` header so pipeline specs don’t hang under the new guard.

<sup>Written for commit 12f2e0ce7adaf0a4bd9808c2bb829d1124c94779. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,5 @@ ollama.pid
 
 # Local design reference library (cloned, not vendored)
 .reference/
+.worktrees
+worktrees
--- a/e2e/fixtures/mock-ollama.js
+++ b/e2e/fixtures/mock-ollama.js
@@ -11,21 +11,48 @@
 // PR1 org-lock spec doesn't exercise Ollama and so doesn't start it.
 //
 // Contract testing (Phase 4): pass { chatReply } to return a fixed assistant
-// message instead of the default 'ok', and read lastChatPrompt() to assert what
-// prompt the summarizer built (e.g. that it embedded the transcript). Default
-// behaviour is unchanged, so the existing @pipeline specs are unaffected.
+// message instead of the default summary, and read lastChatPrompt() to assert
+// what prompt the summarizer built (e.g. that it embedded the transcript).
+//
+// The default reply is a minimal but *valid* meeting summary markdown — it must
+// contain a `## Summary` header. process_streaming (simple_recorder.py) now
+// rejects a streamed summary with no `## Summary` section (a context-overflow
+// truncation signal) by exiting non-zero without writing the *_summary.md the
+// @pipeline specs poll for, so a bare 'ok' reply would make those specs hang
+// until their file-poll timeout. Specs that assert on the reply text override
+// chatReply explicitly and are unaffected.
 const http = require('http');
 
 const OLLAMA_PORT = 11434;
 
+// Minimal valid summary the @pipeline specs accept: a real `## Summary` header
+// plus the other sections the markdown prompt asks for, so _parse_streamed_markdown
+// yields a non-empty meeting and process_streaming writes the summary file.
+const DEFAULT_CHAT_REPLY = [
+  '## Summary',
+  'A short end-to-end test summary of the pipeline run.',
+  '',
+  '## Key Topics',
+  '### Pipeline',
+  'The synthetic recording exercised the transcribe-and-summarise pipeline.',
+  '',
+  '## Key Points',
+  '- The pipeline ran end to end.',
+  '',
+  '## Action Items',
+  '- None.',
+  '',
+].join('\n');
+
 /**
  * Start the mock Ollama on 11434.
  * @param {object} [opts]
- * @param {string} [opts.chatReply='ok'] assistant content returned by /api/chat.
+ * @param {string} [opts.chatReply] assistant content returned by /api/chat.
+ *   Defaults to a minimal valid summary markdown (with a `## Summary` header).
  * @returns {Promise<{ close: () => Promise<void>, lastChatPrompt: () => string|null, chatCalls: () => number, pullCalls: () => number }>}
  */
 function startMockOllama(opts = {}) {
-  const chatReply = opts.chatReply ?? 'ok';
+  const chatReply = opts.chatReply ?? DEFAULT_CHAT_REPLY;
   let lastChatPrompt = null;
   let chatCalls = 0;
   let pullCalls = 0;
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -861,6 +861,40 @@ def process_streaming(audio_file, name, notes):
 
         print("STREAM_COMPLETE", flush=True)
 
+        # Validate the streamed summary before we treat it as a success. A
+        # context overflow (e.g. a transcript longer than the model's num_ctx)
+        # makes the model emit truncated output with no `## Summary` header.
+        # _parse_meeting_markdown then yields an empty summary and the UI shows
+        # "No summary available" — but the file was saved as a success and a
+        # title was hallucinated from the garbage. Detect the missing header
+        # (same convention _parse_streamed_markdown / _parse_meeting_markdown
+        # use) and fail loudly instead. We still write the raw output to a
+        # debug sidecar (NOT the *_summary.md the renderer reads) so the broken
+        # output can be inspected without the app navigating to a fake meeting.
+        has_summary_header = any(
+            line.strip().startswith('## Summary')
+            for line in streamed_md.split('\n')
+        )
+        if not has_summary_header:
+            audio_path = Path(audio_file)
+            debug_path = recorder.output_dir / f"{audio_path.stem}_summary.invalid.md"
+            try:
+                debug_path.write_text(streamed_md, encoding='utf-8')
+            except Exception as e:
+                logger.warning(f"Failed to write invalid-summary debug file: {e}")
+            # ERROR: + non-zero exit mirrors the existing failure contract the
+            # Electron main relies on (close code != 0 -> processing-complete
+            # success:false). No SAVED: is emitted, so no hallucinated meeting
+            # is navigated to.
+            print(
+                "ERROR: Summary generation failed: the model returned no "
+                "'## Summary' section, which usually means the transcript "
+                "exceeded the model's context window. Try a model with a "
+                "larger context window or a shorter recording.",
+                flush=True,
+            )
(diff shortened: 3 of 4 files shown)


Optic00:
Both findings fixed in 5d09a50:

**P2 #1 (non-streaming path unguarded):** Added the same pre-flight context check to `summarize_transcript()` — same guard, same threshold, same error message.

**P2 #2 (notes excluded from estimate):** Token estimate now uses `(len(transcript) + len(notes or "")) / 4` in both paths.

Optic00:
Windows T2 pipeline failing on `audio-import.t2 - leaves-the-original-on-disk` and `transcription-pipeline.t2 - summary-real-dir-untouched` — both are unrelated to this PR's changes (summarizer.py / simple_recorder.py). #215 passed Windows CI on the same runner today. Retriggering by pushing a no-op commit.

Optic00:
Superseded by #250 (map-reduce summarization for long meetings).

#250's PR description states it *"Replaces the prior #246 context-overflow `ValueError` with a short/long fork"* — instead of erroring on overflow, it chunks + summarizes long meetings and routes failures through `STREAM_ERROR:` with a model recommendation. #252 (report templates) already builds on #250, not on this branch, so this is the path forward. Closing to avoid two divergent PRs touching the same `process_streaming` / summarizer overflow path.

**One backstop worth porting into #250 (not present there):** the post-stream `## Summary` header guard in `simple_recorder.py`. It catches the case where the model *streams non-empty output with no `## Summary` header* (no exception raised) on the short path — currently that would still be saved as a success with a hallucinated title. #250 handles the overflow/empty-reduce cases but not this specific header-less-success case. Recommend folding the guard (write `_summary.invalid.md`, emit `ERROR:` + exit 1) into #250's short path as a small follow-up. The e2e `mock-ollama.js` default-reply change from this branch is also covered by #250's own mock-ollama extension.

PR #247, opened 2026-06-22
fix: rescue live transcript when batch transcription returns empty (#207)

### LINKED ISSUE #207: Empty final transcription discards a working live transcript and is saved as a "No speech detected" meeting
**Summary**

When the post-recording batch transcription returns empty, the app saves a meeting whose body is `No speech detected in audio` — even when (a) the audio demonstrably contains speech and (b) a **complete live transcript was produced and shown during the recording**. The live transcript is held in the main process but never persisted or used as a fallback, so known-good content is silently discarded. The empty batch result itself looks like a rare non-reproducible transient; the data-loss-shaped UX around it is deterministic and fixable.

**What happened (real meeting)**

29-min system-audio meeting, Parakeet, German. Live transcription was enabled and **worked the whole time** — the user watched segments stream in and toggled the live view during the call. After stop: `No speech detected in audio` and an empty transcript file.

Forensics on the saved `.webm`:
- File intact: 29:02, both channels carry real audio (system `RMS max 0.091`, well above the gate).
- Bundled ffmpeg decodes it fully (Opus/WebM → 55 MB PCM).
- The **byte-identical release backend** re-transcribes the same file correctly: 21k chars, `detected_language=de`, properly diarised — in both a dev run and the packaged binary under a stripped PATH.
- Running the live `transcribe-stream` sidecar concurrently with the final `process` pass (two Parakeet/MLX contexts on the GPU at once, harsher than the real stop overlap), across several iterations, **did not reproduce** the empty result.

So the empty batch transcription is an unreproducible transient. The fixable parts:

**Gap 1 — a working live transcript is discarded**

`liveTranscriptState.segments` (`app/main.js:2672`) accumulates the full live transcript and is only ever sent to the renderer; the final `process-streaming` call receives just the audio path. When the batch pass yields empty, that captured transcript is thrown away instead of being used as a fallback.

**Gap 2 — empty-on-audio is saved as success, not flagged**

In `transcribe_diarised` (`src/transcriber.py:1028-1058`), a channel that passes the RMS-energy gate (`*_has_audio == True`) but returns empty text with `transcription_failed` unset falls through silently, contributing no segments → the pipeline writes `No speech detected in audio` and saves a normal meeting. The RMS/empty contradiction is already visible in the logs (`max=0.091` for a channel that transcribed to nothing).

**Expected**

A recording with detected audio energy and/or a captured live transcript should never be silently saved as an empty "No speech detected" meeting. Either fall back to the live transcript, or mark the meeting as a transcription failure (preserve audio, offer retry — the existing `channel_failed` path at `transcriber.py:1063` already does this).

**Proposed fix**

Prefer Gap 1 (recovers the content the user already saw): when the batch transcript is empty but a non-trivial live transcript exists for the session, persist the live transcript as the meeting transcript (flagged as live-sourced). As a backstop, implement Gap 2 so an empty-on-energy channel surfaces as a retryable failure rather than a saved empty meeting. Happy to send a PR with a model-free T2 spec.

Environment: macOS, Steno v0.5.4, Parakeet TDT v3, Apple Silicon (M4, 24 GB).

Optic00:
Fixes #207.

## What happened

When a 29-min meeting with live transcription working correctly
throughout was stopped, the batch transcription returned empty and the
meeting was saved as "No speech detected" — silently discarding the
live transcript the user had seen all along.

Two independent gaps caused this:

**Gap 1:** `liveTranscriptState` accumulated the live segments in Electron
but they were never used when batch came back empty.

**Gap 2:** A channel that passes the RMS energy gate but returns no text
was treated as a silent success instead of a failure.

### Gap 1 — live transcript fallback

- Electron snapshots live segments at recording stop into an OS-temp
  file and passes it to `process-streaming` via `--live-transcript`
- Python uses the live transcript when: batch transcription failed
  (`transcription_failed=True`) OR result exactly matches the silence
  sentinel string
- Saved meeting is marked `is_live_transcript: true` so UI/future code
  know no batch transcript exists; audio is kept for reprocessing
- Snapshot waits on the sidecar's `close` event (not `exit`) to ensure
  the final utterance is flushed before reading; 8 s safety-net (5 s
  watchdog + 3 s grace) prevents hanging on a zombie process
- Quick stop→start safe: stale sidecar stdout handler exits early when
  superseded; kill watchdog and drain promise are per-process-instance,
  not global

### Gap 2 — empty result on energetic channel

- `transcribe_audio()` sets `transcription_empty: True` **before**
  substituting the silence sentinel, so the downstream diarised path
  can distinguish a truly empty result from short speech (previously the
  sentinel made the text truthy and the check was dead code)
- Fails the meeting only when **no channel** produced usable text; one
  channel with real text + one noisy-but-empty channel still succeeds

## Tests

Four new deterministic unit tests (model-free):
- `test_empty_text_on_energetic_channel_fails_meeting` — Gap 2 core
- `test_silent_channel_is_not_a_failure` — regression: silence ≠ failure  
- `test_one_channel_text_one_empty_succeeds` — regression: mixed channels
- `test_live_transcript_marker_round_trips` — `is_live_transcript` in frontmatter
- `test_batch_meeting_has_no_live_marker` — regression: normal batch unaffected
- `test_silence_sentinel_constant_agreement` — sentinel value stays in sync

---
## Summary by cubic
Falls back to the live transcript when batch transcription fails or returns only silence, so users don’t lose the transcript they watched during recording. Also fixes diarisation to only fail meetings when no channel yields real text.

- **Bug Fixes**
  - Electron snapshots live segments at stop and passes them to `process-streaming` via `--live-transcript`.
  - Waits for the sidecar to fully drain on `close` with a per-process promise; 5s watchdog + 3s grace; safe on quick stop→start.
  - Backend uses the live transcript only if batch failed or equals the silence sentinel; rescues even very short live transcripts; marks `is_live_transcript` and keeps audio.
  - Adds `SILENCE_SENTINEL` and a `transcription_empty` flag; fail only when all energetic channels return empty; mixed channels succeed.
  - Writes the canonical `_transcript.txt` via a shared helper with UTF‑8 encoding so files are consistent and multilingual-safe; cleans up the temp snapshot.
  - Adds unit tests for empty-on-energy failure, mixed channels, silence ≠ failure, live marker round-trip, normal batch, and sentinel sync.
  - Adds a model-free e2e test that drives `process-streaming` with a silent stereo WAV + `--live-transcript`; verifies rescue (`is_live_transcript: true`), transcript overwrite (no sentinel), audio retention; extends `make-wav` to generate stereo/silence.

<sup>Written for commit 294ca46eefc656986ad6e124be10fd111116d430. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -2470,7 +2470,22 @@ function spawnLiveTranscribe(sessionName) {
   liveTranscribeSessionName = sessionName;
   liveTranscribeStdoutBuf = '';
 
-  liveTranscribeProcess.stdout.on('data', (data) => {
+  // Per-instance drain promise: resolves when THIS process has fully closed
+  // its stdio (all stdout `data` events processed), not merely exited. We bind
+  // the resolver to `proc` so a quick stop→start can't have the old process's
+  // teardown resolve/null the NEW process's state (#207 review-2, Finding 2).
+  const proc = liveTranscribeProcess;
+  proc._drainResolve = null;
+  proc._drainPromise = new Promise((resolve) => {
+    proc._drainResolve = resolve;
+  });
+
+  proc.stdout.on('data', (data) => {
+    // Drop output from a superseded process. On a quick stop→start the OLD
+    // sidecar may still be draining; without this guard its stdout would write
+    // into the global buffer / liveTranscriptState and contaminate the NEW
+    // recording with stale segments (#207 review, Blocker 1).
+    if (liveTranscribeProcess !== proc) return;
     // Stdout arrives in arbitrary-sized chunks; concatenate then split on
     // newlines so a multi-line frame straddling reads is handled.
     liveTranscribeStdoutBuf += data.toString();
@@ -2482,26 +2497,51 @@ function spawnLiveTranscribe(sessionName) {
     }
   });
 
-  liveTranscribeProcess.stderr.on('data', (data) => {
+  proc.stderr.on('data', (data) => {
     // Python logger output goes to stderr — bubble through debug log
     // without spamming the renderer.
     sendDebugLog(`[live-transcribe] ${data.toString().trim()}`);
   });
 
-  attachProcessingStderr(liveTranscribeProcess, 'live-transcribe');
+  attachProcessingStderr(proc, 'live-transcribe');
 
-  liveTranscribeProcess.on('exit', (code, signal) => {
-    sendDebugLog(`Live transcribe sidecar exited code=${code} signal=${signal}`);
-    liveTranscribeProcess = null;
-    liveTranscribeSessionName = null;
-    liveTranscribeStdoutBuf = '';
-    // Clear the load clock if the sidecar died before LIVE_READY, so a later
-    // path can't log a duration against this stale stamp.
-    parakeetLoadStartedAt = 0;
+  // Wait on `close`, not `exit`: `exit` fires when the child terminates but
+  // does NOT guarantee stdout is fully drained and every `data` event has been
+  // processed — so the FINAL segment may not yet be in liveTranscriptState.
+  // `close` fires only after all stdio streams have closed, which is the
+  // correct barrier for the #207 fallback snapshot (review-2, Finding 1).
+  proc.on('close', (code, signal) => {
+    sendDebugLog(`Live transcribe sidecar closed code=${code} signal=${signal}`);
+    // Only clear globals if THIS process is still the active one. A quick
+    // stop→start may have already installed a fresh sidecar; nulling here
+    // would destroy the new process's state (review-2, Finding 2).
+    if (liveTranscribeProcess === proc) {
+      liveTranscribeProcess = null;
+      liveTranscribeSessionName = null;
+      liveTranscribeStdoutBuf = '';
+      // Clear the load clock if the sidecar died before LIVE_READY, so a later
+      // path can't log a duration against this stale stamp.
+      parakeetLoadStartedAt = 0;
+    }
+    // Unblock this instance's drain waiter (#207, Fix 2) now that all stdout
+    // has been flushed and parsed by the stdout handler above. Resolver is
+    // bound to `proc`, so it only ever resolves for the right process.
+    if (proc._drainResolve) {
+      const resolve = proc._drainResolve;
+      proc._drainResolve = null;
+      resolve();
+    }
   });
 
-  liveTranscribeProcess.on('error', (err) => {
+  proc.on('error', (err) => {
     sendDebugLog(`Live transcribe sidecar error: ${err.message}`);
+    // A spawn-time error means `close` may never fire — unblock the drain
+    // waiter so the fallback snapshot doesn't hang on a process that died.
+    if (proc._drainResolve) {
+      const resolve = proc._drainResolve;
+      proc._drainResolve = null;
+      resolve();
+    }
     if (mainWindow && !mainWindow.isDestroyed()) {
       mainWindow.webContents.send('live-transcript-error', {
         sessionName,
@@ -2516,7 +2556,7 @@ function spawnLiveTranscribe(sessionName) {
   // immediate errors). Without a listener, EPIPE bubbles to uncaught and
   // crashes the main process. Race window: chunks queued in the IPC
   // pipeline land here after the sidecar exited (stop, crash, kill).
-  liveTranscribeProcess.stdin.on('error', (err) => {
+  proc.stdin.on('error', (err) => {
     if (err && err.code === 'EPIPE') {
       // Expected when Python exited mid-write. The exit handler will
       // null the ref so subsequent chunks short-circuit at the guard.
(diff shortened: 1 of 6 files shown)


Optic00:
Both P2s fixed in c74374e:

**P2 #1 (UTF-8 encoding):** `_write_transcript_file` now opens transcript files with `encoding='utf-8'` explicitly — safe for multilingual content on all platforms, including Windows.

**P2 #2 (100-char threshold):** Dropped `_LIVE_TRANSCRIPT_MIN_CHARS` entirely. The condition now checks `live_transcript_text.strip()` — any non-whitespace live content is worth rescuing when batch fails or returns silence. A 3-minute stand-up that produces 60 chars of live transcript shouldn't be lost.

ruzin:
Reviewed as part of a batch review of the open PRs. The logic is solid — the per-process drain/watchdog binding, the `close`-not-`exit` drain barrier, and the stale-stdout guard correctly handle quick stop→start and zombie sidecars, and the empty-vs-silence dead-code fix (setting `transcription_empty` *before* the sentinel substitution) is a real catch.

One blocking item before merge, plus a few non-blocking follow-ups:

**Blocking**
- **Missing model-free T2 e2e spec.** This is a user-facing behavior change (a previously-discarded meeting is now rescued and saved with `is_live_transcript: true`), so per CLAUDE.md's standing rule it needs a model-free T2 spec asserting disk state in the *same* change. The unit tests can't cover the Electron↔Python wiring (`--live-transcript` arg passing, temp-file snapshot/drain ordering against a real stop). `honest-failure.t2` is a good template — suggested assertion: drive a stop where batch returns the silence sentinel, then assert a meeting with `is_live_transcript: true` lands on disk and the temp snapshot is consumed.

**Non-blocking**
- The two `LiveTranscriptFallbackTests` (`tests/test_transcriber_failure.py`) assert a locally-reimplemented `should_fall_back()` / inline overwrite rather than driving the real `process_streaming` predicate — they'd stay green even if production diverges. Worth pointing them at the actual function.
- The temp snapshot in `os.tmpdir()` is only unlinked in `processNextInQueue`'s `finally` — a quit/crash while the job is still queued leaks the file. A startup sweep of `stenoai-live-*.txt` would close it (minor).
- Description says "touches app/main.js and simple_recorder.py only", but it also changes `src/transcriber.py` (the load-bearing part) and the test file — just a description nit.

CI is green; happy to re-review once the T2 spec is in.

Optic00:
Added the model-free T2 spec you asked for: `e2e/specs/live-transcript-fallback.t2.spec.ts` (+ a small `makeWav` stereo/silence option). Pushed in 294ca46.

**What it does (model-free, runs in the fast t2 lane — not `@pipeline`):**
- Feeds a **silent stereo** WAV through the real bundled backend's `process-streaming`. Both channels fail the RMS energy gate, so `transcribe_diarised` returns the silence sentinel **with no model loaded** (`WhisperTranscriber` is lazy — `self.model = None` at construction, the ggml weight only loads inside `transcribe_audio`, which the gate prevents reaching). Summarisation goes to the mock Ollama.
- Drives the rescue via `--live-transcript` and asserts on disk:
  - the note frontmatter carries `is_live_transcript: true`
  - the canonical `_transcript.txt` holds the **live** text, not the sentinel (this is the file the crash path used to leave missing entirely)
  - the **summariser prompt** embedded the live text and not the sentinel (so a regression that still fed the sentinel into the note body can't pass)
  - the audio is **preserved** for a later retry

Verified green locally against a freshly built bundle (`1 passed`, ~0.9 s), and run past an independent review.

**Honest scope caveat.** This pins the half the bug actually lived in — the `--live-transcript` arg contract + the Python rescue decision + the on-disk result. The **Electron snapshot/drain** half (`snapshotLiveTranscriptForFallback` reading `liveTranscriptState`) is *not* model-free testable: only the live ASR sidecar populates `liveTranscriptState`, and there's no model-free seam to seed it through the app. I left it to manual/`@pipeline`; happy to add an `@pipeline` variant that drives a real stop→empty-batch→rescue if you'd prefer that covered too.

On your non-blocking notes: the `LiveTranscriptFallbackTests` reimplementing the predicate — this T2 now exercises the **real** `process_streaming` rescue end to end, which is the integration-level answer to that. The startup sweep for leaked `stenoai-live-*.txt` snapshots I've left as a noted follow-up to keep this PR tight; can fold it in if you want it here. Also fixed the description (it does touch `src/transcriber.py`).

ruzin (approved):

PR #275, opened 2026-07-02
Website redesign: hero demo, Features grid, mobile + perf/SEO pass

WilliamDrewett:
## Description

Redesigns the marketing site's hero and Features section with an animated app-window demo and card previews, then rounds it out with a mobile-responsiveness pass (nav hamburger menu, responsive demo/card scaling), new TrustStrip and CTAFooter sections, an SEO pass (OG/Twitter meta, JSON-LD, sitemap.xml, robots.txt, favicon), and a bundle-size/perf pass (route-level code-splitting, framer-motion's LazyMotion, deferred PostHog load). Also fixes a missing accessible name on the desktop GitHub nav link and removes dead mockup code left over from the Features redesign.

## Type of Change

- [x] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested locally with `npm run dev`
- [x] Tested on macOS

## Additional Notes

Website-only change (`website/`), no Electron/Python app surfaces touched. Verified at 375/390/430/768/1280px viewports with no console errors; self-reviewed the full branch diff (8-angle review) before opening.

---
## Summary by cubic
Redesigned the marketing site with an animated hero demo and a new Features grid, plus a mobile-friendly nav. Faster loads, better SEO, and a more resilient UX via route-level code splitting, `framer-motion` `LazyMotion`, idle‑deferred `posthog-js` and GA4, and an error boundary for lazy sections.

- **New Features**
  - Animated app-window hero demo; redesigned Features grid with motion cards.
  - Mobile: hamburger nav; responsive scaling for the demo/cards; fixed Models section height on small screens.
  - New TrustStrip and CTAFooter; Discord link added to the footer.

- **Performance & SEO**
  - Perf + resilience: route-level code splitting with `React.lazy`/`Suspense` and `LazyMotion`/`m`; error boundary for lazy sections; retry-scroll anchor links; safer hero demo (inert buttons, guarded layout reads); hide mock cards from screen readers.
  - Analytics + SEO: idle-deferred `posthog-js` and GA4 behind PROD gates; added OG/Twitter meta, canonical URLs, JSON-LD, `sitemap.xml`, `robots.txt`, and a new favicon.

<sup>Written for commit 45e3b166900ec9a7720e8a718b2f616f08ca86ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


website/src/components/AppWindowDemo.jsx
```diff
@@ -0,0 +1,722 @@
...
+  useEffect(() => {
+    const ms = SCREEN_MS[screen] ?? 8000;
+    const t = setTimeout(() => setScreenIdx((i) => (i + 1) % SCREENS.length), ms);
+    return () => clearTimeout(t);
+  }, [screen]);
+
+  useEffect(() => {
+    const el = outerRef.current;
+    if (!el) return;
+    const measure = () => setScale(Math.min(1, el.offsetWidth / NATIVE_W));
```
Optic00: This `measure()` has no guard against a transient `offsetWidth === 0` read (e.g. bfcache restore), unlike the sibling `ScaledApp` in `Features.jsx` (`if (!s) return; // transient 0-width measurement ...`) added in this same PR. Since `height: NATIVE_H * scale` derives directly from this, a zero reading collapses the whole hero demo to 0px height. Since this component renders eagerly in the hero (not lazy), it's reachable on first paint — the exact scenario the sibling component already guards against.

```suggestion
    const measure = () => {
      const s = Math.min(1, el.offsetWidth / NATIVE_W);
      if (!s) return;
      setScale(s);
    };
```

website/src/App.jsx
```diff
@@ -1,27 +1,34 @@
...
       <TrustStrip />
-      <HowItWorks />
-      <Features />
-      <Models />
-      <Industries />
-      <FAQ />
-      <CTAFooter />
-      <Footer />
-    </>
+      <Suspense fallback={null}>
```
Optic00: Two related things worth considering for this new `lazy()`/`Suspense` boundary:

1. **Anchor nav breaks during the load window.** `Nav.jsx`'s links are plain `<a href="#how">` etc., and those ids (`#how`, `#features`, `#industries`, `#faq`, `#download`) only exist once each lazy chunk mounts. A click before that chunk resolves does nothing, with no retry/scrollIntoView fallback — all six chunks fetch immediately on mount (since `Suspense` renders as soon as this tree mounts), so the window is real but short on a fast connection; larger on slow ones.
2. **No error boundary anywhere in `website/src`.** If a lazy chunk fails to load (network blip, or a stale deployed `index.html` referencing an evicted chunk hash after a redeploy — a classic SPA deploy race), `React.lazy`'s throw is uncaught and unmounts the whole tree, including `Nav`/`Hero` — a blank page instead of just a missing section. This risk didn't exist before this PR (no lazy imports previously). Worth wrapping this `Suspense` in a small error boundary that falls back to `null`/a reload prompt rather than taking the whole page down.

website/src/sections/Features.jsx
```diff
@@ -33,20 +905,52 @@ export function Features() {
...
+                    turn into a wall of oversized previews. */}
+                <div
+                  style={{
+                    height: "clamp(180px, 45vw, 320px)",
+                    position: "relative",
+                    overflow: "hidden",
+                    background: "var(--surface-raised)",
+                  }}
+                >
+                  {createElement(Image)}
```
Optic00, line 938: The old code marked each feature's icon `aria-hidden="true"` (visible as deleted lines in this diff). None of the new fake-UI mockup components (`ChatWithNotesImage`, `NotesPane`, `NinetyNineLanguagesImage`, etc.) carry `aria-hidden`/`role` anywhere in this file. These aren't purely visual — they render real text nodes (fake chat Q&A, "New chat" button, dates, 30 language names, a fake code snippet) that a screen reader will read as if it were actual card content, interleaved with the real title/body. Worth wrapping each mockup's container `<div>` in `aria-hidden="true"` the way the old icons were.

website/src/components/AppWindowDemo.jsx
```diff
@@ -0,0 +1,722 @@
...
+    measure();
+    const ro = new ResizeObserver(measure);
+    ro.observe(el);
+    return () => ro.disconnect();
+  }, []);
+
+  return (
+    <div aria-hidden="true" className="relative max-w-full" style={{ padding: "2px 2px 32px", textAlign: "left", pointerEvents: "none", userSelect: "none" }}>
```
Optic00: This root wrapper is `aria-hidden="true"` (correctly hiding the decorative demo from AT), but it still contains ~12 native `<button>` elements with no `tabIndex={-1}`. `aria-hidden` on an ancestor doesn't remove descendants from the native tab order, so a keyboard user tabbing through the hero lands on a dozen silent/dead tab stops (buttons with no visible focus semantics, since the `aria-hidden` ancestor suppresses their accessible name for AT). Worth adding `tabIndex={-1}` to each button in here, or one wrapping `inert` if it's supported in your target browsers.

website/index.html
```diff
@@ -3,13 +3,74 @@
...
+        "operatingSystem": "macOS, Windows",
+        "offers": {
+          "@type": "Offer",
+          "price": "0",
+          "priceCurrency": "USD"
+        }
+      }
+    </script>
+    <!-- Google tag (gtag.js) -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-48VX3C2M17"></script>
```
Optic00: This GA4 snippet has no equivalent to `analytics.js`'s `if (!import.meta.env.PROD) return` gate for PostHog, and no idle/lazy deferral — it's static markup, so it loads and fires `gtag('config', ...)` synchronously on every environment including `npm run dev` and PR preview deploys. That both pollutes the real GA property with non-production traffic and undercuts this PR's own "analytics deferred via idle callback" framing (which only actually applies to PostHog). Might be worth loading this the same way PostHog is loaded (idle callback + dynamic injection), or at least gating it on a build-time env check.

website/index.html
```diff
@@ -3,13 +3,74 @@
   <head>
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/stenoai-logo.svg" />
-    <link rel="icon" type="image/png" href="/dragonfly-logo-512.png" />
+    <link rel="icon" type="image/png" href="/favicon-64.png" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta name="description" content="Steno is the AI powered intelligence layer for all your confidential workflows. No cloud, no usage limits and no bots joining your calls." />
-    <title>stenoAI - confidential intelligence layer</title>
+    <meta name="description" content="Free AI meeting notes with full privacy. Steno transcribes and summarizes meetings entirely on your device — no cloud, no bots joining calls, no usage limits." />
+    <title>stenoAI — Free, Private AI Meeting Notes (No Cloud)</title>
```
Optic00, line 9: This new title/meta-description (and the matching OG/Twitter/JSON-LD copy) pivot the positioning to "Free, Private AI Meeting Notes" — but the actual on-page H1/subhead in `Hero.jsx` is unchanged by this PR and still reads "AI for your confidential workflows" / "the AI powered intelligence layer for all your confidential workflows," with no mention of meetings, transcription, or notes anywhere. Visitors arriving from a search result or social share promising "AI Meeting Notes" land on a page that doesn't say that. Might be worth aligning one side to the other (either update the hero copy, or soften the meta copy back to the confidential-workflows framing) so search/social previews match what people actually see.

Optic00 (commented):
Independent review (8-angle diff scan, each finding verified against the actual checked-out code before posting). Nice redesign overall — flagging a few things worth a look before merge, mostly around the new lazy-loading, a couple of real a11y regressions on the new mockup content, and an analytics/SEO consistency gap. The download-tracking race condition I originally found is already fixed by the latest commit (`loadPosthog()` now calls `init()` inside the same promise chain) — no action needed there, just noting it for the record.

website/index.html
WilliamDrewett: I want to do a bit more work on the value propositions / statements next week. It'll require a bit of work defining our ICPs etc so was planning to align everything once that work is done and approved.

ruzin (dismissed):

PR #284, opened 2026-07-03
fix(transcript): true per-channel speaker attribution for live transcript

valentinweyer:
## Problem
Live transcription mixed mic + system audio into a single mono stream
before transcription, then guessed who was speaking after the fact from
RMS energy ratios. This frequently misattributed segments to the wrong
speaker.

## Fix
Mic and system audio now stay on two separate channels all the way to
the transcription sidecar. Each channel runs its own independent Silero
VAD + Parakeet pipeline, so speaker attribution is a structural fact
(which channel the audio actually came from) instead of a guess. Reuses
the batch pipeline's existing Jaccard + RMS bleed-dedup logic to catch
genuine mic echo of system audio, rather than reinventing it.

## Known limitation / future optimization
The two channels' transcription calls run sequentially, not
concurrently — if both people talk at the exact same moment, one
channel's Parakeet call blocks the other channel's VAD/decode until it
returns. Batching both channels into a single `model.generate()` call
was investigated: parakeet-mlx does accept a batched mel input and the
shared encoder pass would genuinely run batched, but the token-decode
loop is implemented as a sequential per-utterance loop internally
regardless, so batching would only amortize the encoder pass — not
worth the added complexity without profiling evidence it's the actual
bottleneck. A real future win would be dispatching each channel's
`transcribe_samples()` call to a small thread pool (MLX/onnxruntime
release the GIL during compute), so simultaneous speech from both
people actually transcribes in parallel instead of one call stalling
the other.

## Test plan
- [x] Verified end-to-end with a real two-person conversation (headphones, no acoustic bleed) — correct [You]/[Others] labels, correct word order for overlapping long/short turns, no cross-speaker duplication
- [x] Verified mic-only recording still works with low latency
- [x] Self-reviewed the full diff (8-angle review); fixed 5 real bugs found: LIVE_READY/VAD-load ordering, same-speaker partial clobbering, non-deterministic RMS tie-break, bleed pairs split across dedup passes, and a docstring inaccuracy
- [ ] CI (T1/T2 e2e, typecheck, lint) — runs automatically once the PR is open

---
## Summary by cubic
Fixes live transcript speaker attribution by keeping mic and system audio on separate channels and labeling segments at the source. Adds end-to-end coverage to ensure correct per-channel tags and stable ordering.

- **Bug Fixes**
  - Send 16 kHz interleaved stereo (mic=L, system=R) to the sidecar; it de‑interleaves and runs two independent Silero VAD + Parakeet pipelines.
  - Reused batch logic to drop true mic echo of system audio: brief hold + Jaccard text match with RMS tie-break favoring mic.
  - `LIVE_SEG` now includes `speaker`; the renderer stops guessing and `liveRmsBuffer.ts` is removed. `ipc.ts` updated accordingly.
  - Live transcript state keeps finals sorted and maintains at most one trailing partial per speaker to avoid clobbering/out‑of‑order inserts.
  - Added a T2 e2e spec that feeds real interleaved-stereo speech to `transcribe-stream` and asserts correct per-channel `speaker`, no cross-channel clobber, and stable ordering; skips cleanly without Parakeet/Metal or macOS TTS tools.

<sup>Written for commit 1c03b723c3a5255eb6f17d3272b0ffdbb4a4425f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Optic00 (changes requested):
Reviewed the diff (including an independent Codex pass) - the dual-channel pipeline change itself looks sound: IPC shape is consistent between main/preload/renderer, the sequential-transcription trade-off is reasonable (avoids MLX/ONNX thread-safety risk from concurrent model calls, and batching wouldn't remove the decoder bottleneck anyway per your own investigation), and no correctness or security issue was found in the visible diff.

One thing to close before merge: this changes the live-transcript speaker-attribution contract but doesn't touch any `e2e/specs/*` file. CLAUDE.md's standing rule is to add/update e2e coverage in the same change for user-facing feature changes, preferring a model-free T2 spec that drives the backend directly over a UI/T1 spec.

`e2e/specs/incremental-write.t2.spec.ts`'s "coverage gap (intentional)" comment only excuses the real-capture half (needs actual `getUserMedia`/`getDisplayMedia`) - it doesn't cover the new per-channel sidecar contract itself, which is exactly the kind of thing `e2e/specs/live-transcript-fallback.t2.spec.ts` already shows a model-free pattern for (drives the Python sidecar directly via its CLI contract, no real capture, no model needed for the specific behavior under test).

Suggested fix: a T2 spec that feeds deterministic interleaved stereo samples into the live sidecar and asserts the `LIVE_SEG` output carries the correct `speaker: "You"` / `speaker: "Others"` per channel with stable ordering - no cross-channel clobber. If that's only achievable by loading the real Parakeet model, tag it `@pipeline` instead of leaving it uncovered; either way it shouldn't ship with zero coverage of the actual bug this PR fixes.

Latency/backpressure under simultaneous speech (two people talking at once) is worth keeping an eye on post-merge, but I wouldn't block on it without profiling evidence - flagging as a possible follow-up, not a blocker.

Optic00 (approved):
Re-review after the test commit (1c03b72) - my earlier blocker is fully addressed. Approving.

The one thing I asked for before merge was e2e coverage of the new per-channel speaker-attribution contract. `live-transcript-speaker.t2.spec.ts` is exactly that: it drives the real `transcribe-stream` sidecar directly (mirrors `live-transcript-fallback.t2.spec.ts`'s direct-CLI pattern), feeds a deterministic non-overlapping interleaved-stereo stream (mic=L / system=R, two distinct pangrams), and asserts the `LIVE_SEG` output carries the right `speaker` per channel with no cross-channel clobber and stable ordering - plus the real-user-data isolation keystone. The `@pipeline` tag is the right call and the skip logic is loud/annotated (backend missing, non-mac / no `say`+`afconvert`, non-parakeet lane, or a model-missing `LIVE_ERROR`), so it never silently green-passes.

Since `@pipeline` specs skip on GitHub CI (no Metal) and only run on local Apple Silicon, I verified it actually passes rather than just trusting it: I ran the fixture through this branch's sidecar directly. Per-channel attribution is correct end to end -

```
LIVE_SEG ... "The quick brown fox jumps over the lazy dog." is_final speaker="You"     (mic/L, start 0.704)
LIVE_SEG ... "Pack my box with ... dozen liquor ..."        is_final speaker="Others"  (system/R, start 5.056)
```

Every assertion in the spec holds against that real output: both speakers present, `youText` matches `/fox/` and not `/dozen|liquor/`, `othersText` matches `/dozen|liquor/` and not `/fox/`, and You precedes Others. The structural per-channel split does what the PR claims - no RMS-ratio guessing.

The production diff itself I'd already reviewed as sound in the first pass (IPC shape consistent, sequential-transcription trade-off reasonable), and the test commit only adds the spec + fixture, so nothing there changed. The simultaneous-speech latency point stays a post-merge follow-up (you've documented it as the thread-pool future win), not a blocker.

LGTM.

PR #286, opened 2026-07-03
fix: navigate to note when clicking "Note ready" notification

WilliamDrewett:
## Summary
- Clicking the "Note ready" desktop notification previously only refocused the Steno window; it now navigates straight to the note that just finished.
- Threads the finished note's `summaryFile` through `showNoteReadyNotification`, and on click, main.js sends a new `navigate-to-meeting` event that the renderer listens for and navigates to `/meetings/<summaryFile>`.
- Hard-failure notifications (nothing was ever written) keep the old focus-only behavior, since there's nothing to open.
- Auto-navigate-while-idle logic is untouched — only the explicit click case changes, since a click is unambiguous user intent, unlike the app auto-navigating someone away from what they're doing.

## Testing note
I was not able to manually verify the click behavior end-to-end. In this sandbox:
- The Python backend isn't built (no `dist/stenoai/stenoai`), so the onboarding wizard can't complete.
- More importantly, macOS itself refuses to deliver notifications from an **unsigned dev-mode Electron build** — confirmed by testing a bare `new Notification().show()` outside the app entirely, which emitted a `failed` event even with notification permissions correctly granted and no Focus/DND active. This is a known macOS/Electron constraint (code signing required for `UNNotification` delivery), not a bug in the app.
- No Developer ID signing identity is available in this workspace's keychain, so I couldn't produce a signed local build to work around it either.

What I did verify:
- `npm run typecheck:renderer` and `npm run lint:renderer` pass with no new issues.
- All 18 hermetic T1 e2e specs pass (no regressions).
- The change mirrors the exact `webContents.send` → `ipcRenderer.on` → `navigate()` pattern already used and tested elsewhere in the app (e.g. `auto-record-requested`, `shortcut-start-recording`).
- Updated `notifications.t2.spec.ts` to pass `summaryFile` through the existing notifications-gate test.

Recommend a real click-through test on a properly signed local build (or after this lands) before/soon after merge.

## Test plan
- [ ] On a signed local build (or after merge), background the app, let a note finish processing elsewhere, click the "Note ready" banner, confirm it opens the correct note.
- [ ] Confirm the hard-failure ("Processing failed") notification still only focuses the window.

---
## Summary by cubic
Clicking the "Note ready" desktop notification now opens the exact note instead of just refocusing the app. This threads the note’s `summaryFile` through to main and emits a `navigate-to-meeting` IPC event that the renderer uses to navigate.

- **Bug Fixes**
  - Added `navigate-to-meeting` IPC and `stenoai.on.navigateToMeeting`; renderer navigates to `/meetings/<summaryFile>` on click.
  - Threaded `summaryFile` through `stenoai.settings.showNoteReadyNotification`; updated IPC docs and e2e to include it.
  - Hard-failure notifications still only focus the window; idle auto-navigate behavior is unchanged.

<sup>Written for commit 48e0c94d15f0c53d38b5eb6035730ba855b39671. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/docs/ipc-contract.md
+++ b/app/docs/ipc-contract.md
@@ -163,6 +163,7 @@ truth for elapsed time and paused state so remounts recover correctly.
 | `summary-complete` | M→R | yes | `stenoai.on.summaryComplete(cb)` |
 | `processing-complete` | M→R | yes | `stenoai.on.processingComplete(cb)` |
 | `processing-progress` | M→R | yes | `stenoai.on.processingProgress(cb)` |
+| `navigate-to-meeting` | M→R | yes | `stenoai.on.navigateToMeeting(cb)` — fired on "Note ready" notification click when a `summaryFile` exists; renderer navigates to that note's detail route |
 
 ```ts
 interface SessionInfo {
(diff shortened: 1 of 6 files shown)


Optic00 (approved):
Reviewed the full diff directly: reuses the existing webContents.send -> ipcRenderer.on -> navigate() pattern already used elsewhere (auto-record-requested etc.), IPC contract docs and the notifications.t2 e2e spec are updated consistently, hard-failure path is untouched. finishedSummaryFile is a pre-existing variable already used for the auto-navigate route, so the new usage is sound. Author was upfront about not being able to test the actual notification click in this sandbox (macOS blocks notification delivery from unsigned dev builds) but did what's testable (typecheck, lint, T1 e2e).

Issue #278, opened 2026-07-03
Architecture debt follow-ups from 2026-07-03 review


Optic00:
Tracking issue for lower-priority findings from a whole-codebase architecture review (Fable-5 Architect pass, commit c094b51). The two highest-leverage items (dead legacy setup IPC channels + the process-streaming/reprocess stdout buffering bug) are being fixed directly rather than tracked here. This issue bundles the rest so they don't get lost, without implying any of them is scheduled.

Each item below deserves its own PR/plan if and when it's picked up - this issue is a backlog, not a commitment.

- [ ] **IPC contract has no conformance check.** `main.js` (156 `ipcMain.handle`/`.on` registrations), `preload.js`, `renderer/src/lib/ipc.ts`, `docs/ipc-contract.md`, and `e2e-mock-ipc.js` are hand-synced across 5 places with nothing catching drift. A `node --test` that cross-checks the channel sets would catch this cheaply.
- [ ] **`config.json` lost-update race.** `Config._save()` (`src/config.py`) does whole-file read-modify-write per CLI process with no cross-process coordination. Two near-simultaneous settings writes can silently revert each other. Fix: either serialize config-mutating spawns through one queue in `main.js` (matches the existing provider-lock pattern), or add file-lock-based merge in `_save()`.
- [ ] **No consistency owner for the per-meeting file set, and `list-meetings` is O(all meetings) per call.** A meeting is 4-5 files (audio, transcript, `.md`, `_reports.json` sidecar, org-backup state) with no cross-file transaction or repair pass, and every `list-meetings` call re-parses every meeting's markdown via a fresh subprocess spawn. Worth a meetings index/manifest before Reference Material Import / Custom Keywords add more sidecars.
- [ ] **Deep-link parsing has zero test coverage at any tier.** `parseShortcutUrl`/`handleShortcutUrl` (`main.js`, ~line 439) is a pure, security-relevant input surface with no unit or e2e coverage. Candidate for the same extraction treatment as `meeting-detect.js`.
- [ ] **Google Calendar OAuth hardcodes a client secret** (`main.js:495`). Not an emergency (Google treats desktop-app secrets as non-confidential), but the Outlook flow already proves a PKCE-no-secret pattern works in this codebase - migrating Google onto it is a real OAuth-flow change needing real-account testing, not a quick fix.
- [ ] **`Settings.tsx` is 2,808 lines**, the renderer's one emerging monolith (next largest file is 1,685 lines). Split by tab before it grows further.
[...]

PR #291, opened 2026-07-04
fix: #278 - close the config.json cross-process lost-update race

Optic00:
### The bug
The app spawns a fresh `simple_recorder.py` subprocess per operation (no daemon). `Config._save()` did whole-file read-modify-write per process with no cross-process coordination, so two near-simultaneous writers each did (load whole config -> mutate one key -> write whole file) and silently reverted each other's unrelated keys. `_atomic_write_json` already prevents torn files, but not the lost UPDATE across the read-modify-write cycle.

main.js is **not** the only writer, so serialising there can't fix it: config.json is also written by load-time migrations in `Config.__init__` (fired from any subprocess, including pure readers like `get-status`), `get_anonymous_id()` telemetry, the seed-org-auto-backup subprocess, and direct terminal CLI use. The fix therefore lives at the Python write site.

### The fix - snapshot-diff merge under a cross-process file lock
- **Snapshot on load:** `__init__` captures `self._snapshot = copy.deepcopy(self._config)` - exactly what was read from disk.
- **`_save()` under a `filelock.FileLock`** (fcntl on POSIX / msvcrt on Windows; auto-released if a holder crashes): re-read config.json fresh as the merge base, overlay only the keys this process changed since load, atomic-write the result inside the lock, then adopt it and refresh the snapshot.
- **Recursive merge** (`_apply_changes`): recurses into dict-valued keys so two processes editing *different* sub-keys of the same nested dict (per-provider `cloud_models`, `template_overrides`) both survive - we assert only the sub-keys we actually touched and keep the concurrent writer's sub-keys straight from disk. Scalars/lists overlay wholesale (last-writer-wins; no clean per-element merge). Nested deletions (the `reset_template` path) propagate.
- **Timeout fallback:** on `filelock.Timeout` it degrades to today's plain unlocked atomic write (a stuck lock must never block saves) and refreshes the snapshot so later saves stay consistent. Non-`Timeout` errors inside the lock keep the existing `-> False` contract.
- `filelock>=3.13` pinned in `requirements.txt` (already transitively present via `huggingface_hub`; no PyInstaller spec change needed - normal import analysis bundles it).

### Design
Chosen over a main.js serialisation queue after an architecture review: only the Python-side lock covers migration/telemetry/seed/terminal-CLI writers. Granularity is per (possibly nested) key; two writes to the *same* leaf remain last-writer-wins, which is a genuine conflict.

### Tests (model-free, `tests/test_config_persistence.py`)
- `test_two_writers_different_keys_both_survive` - the keystone: two `Config` instances load the same baseline, set different top-level keys; both survive on disk (fails on the old whole-file `_save`, passes now).
- `test_concurrent_edits_to_same_nested_dict_both_survive` + `test_nested_deletion_still_propagates` - the recursive-merge cases.
- `test_load_time_migration_and_concurrent_setter_both_survive`, `test_second_save_in_same_process_diffs_correctly`.
- `test_concurrent_processes_each_key_survives` - a real `multiprocessing` (spawn) stress test with a `Barrier` so all N workers load the same baseline then save simultaneously, forcing the race; proven to catch the old behaviour (1/6 keys survive old vs 6/6 fixed). Terminates timed-out children.
- All pre-existing atomic-write / corrupt-file recovery tests still pass unchanged.

### Verification
- `python -m unittest discover tests` -> 361 pass (13 in test_config_persistence). Stress test run 5x, no flake.
- `ruff check` clean.
- Independent Codex review of the recursive merge across 6 interleaving edge cases (nested concurrent add, untouched-dict disk additions, deletion vs concurrent sibling add, type changes, aliasing, missing snapshot key): no defects.

Part of #278.

---
## Summary by cubic
Fixes the cross-process lost-update race in config.json by saving under a file lock and merging only the keys changed by the current process. Addresses #278 and makes writes safe across CLI subprocesses, migrations, and telemetry.

- **Bug Fixes**
  - Save under a cross-process `filelock` with a 10s timeout; on timeout, fall back to an unlocked atomic write to avoid blocking.
  - Re-read disk and snapshot-diff merge only this process’s changes; recursive for dicts so nested keys from concurrent writers survive; scalars/lists remain last-writer-wins; deletions propagate.
  - After saving, adopt the merged result and refresh the snapshot so subsequent saves diff correctly.

- **Dependencies**
  - Add `filelock>=3.13` (already transitive via `huggingface_hub`).

<sup>Written for commit 1272fa88f120d8f5a1565df11c13e586356861c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,6 +29,13 @@ pywhispercpp>=1.2.0
 # TypeError on it, which the streaming path would swallow into an empty summary.
 ollama>=0.5.0
 click>=8.1.0
+# Cross-process file lock guarding config.json's read-modify-write in
+# src/config.py (_save). The app spawns a fresh CLI subprocess per operation,
+# so concurrent writers would otherwise revert each other's unrelated keys
+# (lost update). Already present transitively via huggingface_hub; pinned
+# explicitly since we now import it directly. Pure Python (fcntl / msvcrt),
+# so PyInstaller picks it up via normal import analysis.
+filelock>=3.13
 pydantic>=2.5.0
 python-dateutil>=2.8.0
 openai>=1.0.0
--- a/src/config.py
+++ b/src/config.py
@@ -466,29 +466,59 @@ class Config:
             return None
         return data if isinstance(data, dict) else None
 
+    @classmethod
+    def _apply_changes(
+        cls, base: Dict[str, Any], current: Dict[str, Any], snapshot: Any
+    ) -> Dict[str, Any]:
+        """Overlay onto `base` (fresh from disk) only the changes between
+        `snapshot` (what we loaded) and `current` (our in-memory dict).
+
+        Recurses into dict-valued keys so two processes editing DIFFERENT
+        sub-keys of the SAME nested dict (e.g. per-provider `cloud_models`, or
+        `template_overrides`) don't clobber each other — we assert only the
+        sub-keys THIS process actually touched and keep the concurrent writer's
+        sub-keys straight from disk. Scalars and lists are overlaid wholesale
+        (a list has no clean per-element three-way merge, so last-writer-wins,
+        matching the design's genuine-conflict stance). A key whose type
+        changed between dict and non-dict also falls back to a wholesale
+        overlay. Nested deletions (config's only deletion path, reset_template
+        dropping a `template_overrides` entry) are propagated too.
+
+        Returns a new dict; does not mutate `base`, `current`, or `snapshot`."""
+        result = dict(base)
+        snap = snapshot if isinstance(snapshot, dict) else {}
+        for key, cur_val in current.items():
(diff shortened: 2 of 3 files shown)


ruzin (approved):
Reviewed: cross-process config lost-update fix — filelock + snapshot-diff per-key 3-way merge, well-tested and isolated. LGTM.

Issue #271, opened 2026-07-02
Silence auto-stop doesn't trigger with background system audio (user ended up with a 5h recording)


Optic00:
## Summary

Silence-based auto-stop (`silence_auto_stop_enabled`, default on, default 15 min) did not trigger for a user, resulting in a 5-hour recording after the meeting had ended.

## Report (from Discord, #support)

Sylvain Ramousse reported: "to add on my previous point, I ended up with a 5h recording because I forgot to stop it at the end of the meeting." A second user (maplesyrup) confirmed hitting the same thing. Sylvain confirmed the auto-stop setting was on with the 15-minute default, so this isn't a case of the feature being disabled/misconfigured.

## Likely root cause

`useSystemAudioCapture.ts` (silence detector, ~line 460-480) only resets the silence window when **both** mic RMS and system-audio RMS drop below `SILENCE_RMS_THRESHOLD`:

```ts
if (micRms > SILENCE_RMS_THRESHOLD || sysRms > SILENCE_RMS_THRESHOLD) {
  lastActiveAtMs = Date.now();
  return;
}
```

That's intentional (per the comment above it, mic-only would false-trigger during a call where the user is just listening). But it means any background app producing sound through system-audio loopback - music, a video, a notification chime, another app with an open audio tab - keeps `sysRms` above threshold indefinitely and the timer never accumulates, even though the meeting itself has long ended. This matches ruzin13's own hypothesis in the thread ("Any backgrounds?").

## Repro (best guess, unconfirmed)

1. Enable "Record system audio" + silence auto-stop (any duration).
2. Start a recording during a call.
3. End the call, but leave something producing system audio in the background (e.g. a browser tab, music player, another app).
4. Auto-stop never fires because system-audio RMS stays above `SILENCE_RMS_THRESHOLD`.

## Discord thread

ruzin13 was already triaging live and asked Sylvain to retest with the interval set to 2 minutes to narrow it down - that data point isn't in yet as of this issue being filed. Filing this so the report doesn't get lost, not as a claim the root cause above is confirmed.

Optic00:
Status update: v0.5.8 (just released) ships the diagnostics groundwork for this (#292). The detector's decisions are now visible in **Settings → Developer → Debug console**, prefixed `[auto-stop]`: when it arms (with the enabled/minutes settings and whether a system-audio analyser exists), every guard-state transition (disabled / paused / inactive / re-armed), and the countdown decisions.

[...]

PR #292, opened 2026-07-04
feat: log silence auto-stop decisions to the Developer debug console (#271 diagnostics)

Optic00:
## Make the silence auto-stop detector's decisions visible in the Developer debug console

Groundwork for diagnosing #271 (silence auto-stop never fires; one user got a 5h recording).

### Why
A user shared a repro recording. Running it through the app's exact detector (8-bit `getByteTimeDomainData`, 512-sample window, 1 Hz, threshold 0.01) showed **both channels silent for the full duration** - no system audio at all, and the mic only picking up room noise below threshold (peak 8-bit RMS 0.0083 < 0.01). So the detector *should* have auto-stopped at the configured 2 min, but didn't. The "background noise keeps it awake" theory doesn't hold for that file.

That means the real cause is the detector not firing on genuine silence - and today that is **undiagnosable remotely**: the detector's only logs are `console.error`/`warn` in the renderer, which go to DevTools, not the user-facing debug log. A normal user can't get at them.

### What
Route the silence-auto-stop detector's decisions into the same renderer ring buffer that backs **Settings > Developer > Debug console** (`appendDebugLog`, already copy-able there). It now logs, `[auto-stop]`-prefixed:

- **armed** - once, with `enabled`, `minutes`, and whether a system-audio analyser exists.
- **idle (disabled|paused|inactive) / re-armed** - guard-state transitions (logged on change, not every tick), so the log shows *why* the timer keeps resetting.
- **silence started / activity** - transitions across the RMS threshold, with the `mic`/`sys` RMS values.
- **still silent Ns / limitS** - a 30 s heartbeat while silence accumulates, so you can see the window actually growing (the missing signal in #271).
- **threshold reached / recording stopped / stop failed / stop error** - the fire path and its result.

For a normal 2 min silent stop that is ~5 lines, not 1/s - it logs transitions plus a 30 s heartbeat, so the 1000-line buffer isn't flooded.

### Safety
Observation-only. The stop **decision** is unchanged - the sole logic-touching edit refactors `if (micRms > TH || sysRms > TH)` into `const silent = !(...); if (!silent)`, which is logically identical; thresholds, the `silenceMs < limitMs` window, and the in-flight latch are byte-for-byte the same. `logAutoStop` wraps `appendDebugLog` in try/catch so a misbehaving log subscriber can never throw into the detector tick.

### Verification
- `typecheck:renderer` clean; `lint:renderer` no new warnings; `npm run build` (vite + electron) succeeds.
- Independent Codex review: refactor confirmed logically identical, log volume confirmed transitions+heartbeat (not per-tick); its two low findings (harden the log call, reset log-transition state on the stop-failure path) are folded in.

No e2e spec: asserting these logs needs a real multi-minute silent recording (heavy, model-bearing), out of proportion for an additive logging change.

Once this ships, a #271 reporter can copy the Developer log and we'll see exactly where the detector goes wrong. Part of #271.

---
## Summary by cubic
Log the silence auto-stop detector’s decisions to the Developer debug console so we can diagnose “auto-stop never fires” cases (issue #271). No behavior change; stop logic and thresholds are unchanged.

- **New Features**
  - Send detector transition logs to the renderer debug ring buffer via `appendDebugLog` (`@/lib/debugLogs`), `[auto-stop]`-prefixed.
  - Logs: armed config; idle/paused/inactive and re-armed; silence/activity crossings with RMS; 30s “still silent” heartbeat; threshold reached; stop result/errors.
  - Emits on transitions + 30s heartbeat only, to avoid flooding the 1000-line buffer.
  - Logging is wrapped in try/catch and logging state resets on stop failure to keep output consistent and non-invasive.

<sup>Written for commit f580f8622c2bc85b616eff42e95d7c25fb99d678. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/292?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/hooks/useSystemAudioCapture.ts
+++ b/app/renderer/src/hooks/useSystemAudioCapture.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { ipc } from '@/lib/ipc';
+import { appendDebugLog } from '@/lib/debugLogs';
 import { isMac } from '@/lib/utils';
 import {
   LIVE_RMS_HZ,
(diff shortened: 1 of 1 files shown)


ruzin (approved):
Reviewed: observation-only [auto-stop] debug logging, guarded so it can't affect the detector; single-file, isolated. LGTM.

PR #295, opened 2026-07-04
chore: retire the stale ipc-contract.md doc (code + conformance test are the contract)

Optic00:
## Retire the stale `app/docs/ipc-contract.md` IPC inventory

The 758-line hand-written IPC contract doc is a finished migration artifact and ~24% stale, so it's a liability rather than an asset. Deleting it.

### Why
- **It's a migration-era artifact, not a living contract.** Its structure gives it away: a "Needed" column tracking which channels the *new* React renderer would adopt, a section 13 "channels the new renderer will not expose", and a header reading "Phase 1's preload.js *will* implement exactly what is listed here". That migration is done.
- **Hand-maintenance has empirically failed.** 36 of the 150 registered `ipcMain.handle/on` channels (24%) are absent from the doc (`save-template`, `delete-report`, `set-user-name`, `org-get-policy`, `get-meeting`, ...) despite a bold-face "update both in the same commit" rule and two source files calling it "the source of truth". A doc that claims authority while being 24% wrong is the worst case for traceability - it already misled once in the other direction (it briefly looked nonexistent to a contributor who checked the wrong path).
- **Everything mechanically enforceable is already enforced, better, elsewhere:** channel shapes by `tsc` via the typed bridge (`app/renderer/src/lib/ipc.ts`), the channel whitelist by `app/preload.js`, and channel-set consistency across main/preload/ipc.ts/e2e-mock by the `app/ipc-contract.test.js` conformance check (added in #288, fails on drift). Option "keep + enforce" would only enforce *mention* (a regex that every channel appears in the doc), not *truth*, at a permanent per-channel doc tax.
- **The prose worth keeping already lives elsewhere, more current.** The recording-architecture semantics are in `MECHANICS.md` (the single renderer-driven capture path, the Parakeet-only live sidecar, the AI-provider lock's fire-and-forget reconcile), and the `process-recording` queue/fire-and-forget behaviour is already a detailed comment on its handler. Nothing unique is lost.

### What this does
- `git rm app/docs/ipc-contract.md`.
- Repoints the one remaining reference the enforcement work didn't cover - `app/ipc-sentinels.js` - to the module itself as the source of truth for the sentinel literals (the three-way main/e2e-mock/renderer string sync). The `preload.js` / `ipc.ts` comment repointing to the conformance test lands via #288.

This was decided with an independent architecture review (verdict: delete, not demote - a 758-line "may be stale" doc is the worst of both). Coordinates with #288 (adds the conformance test + repoints preload/ipc.ts) and #294 (which dropped its one-line touch of this doc to avoid a modify/delete conflict).

---
## Summary by cubic
Retired the stale IPC contract doc; the code and conformance test now define the IPC contract. This reduces drift and keeps enforcement in TypeScript, the preload whitelist, and the conformance test.

- **Refactors**
  - Deleted `app/docs/ipc-contract.md` (outdated migration artifact).
  - Updated `app/ipc-sentinels.js` comment to make this module the source of truth for sentinel literals used by main/e2e-mock/renderer.

<sup>Written for commit 7b93a522137eb99d5a9ba029fcbc430726230fff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/295?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ruzin (approved):
Reviewed: retiring the stale 758-line ipc-contract.md; enforcement is now the conformance test (#288) + tsc + preload whitelist. Sound. LGTM.

PR #298, opened 2026-07-04
feat: redesign templates settings UI and add provider support

Vassista:
## Description

This PR redesigns the templates settings UI with a grid layout and improved card visuals, and adds support for Bedrock and custom API providers to the setup flow.

## Type of Change

- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested on macOS

## Additional Notes

Includes styling changes to the UI for better visual aesthetics when picking a template.

---
## Summary by cubic
Redesigned the Templates settings into a responsive card grid with a full-height editor and clearer default/lock states. Expanded Setup to support AWS Bedrock and custom OpenAI‑compatible providers, with safer validation and more reliable saving.

- New Features
  - Templates grid: cards show name, prompt preview, and badges (Default, Locked, Built‑in) with quick actions and a “New Template” card.
  - Default handling: default template is pinned and can be set from the editor or any non‑default card via “Make Default.”
  - Template editor: full-height view with improved header, language selector, markdown hint, clearer errors, and “Save Template.”
  - Setup: provider choices now include “AWS Bedrock (Claude)” and “Custom (OpenAI‑compatible),” with fields for AWS region, optional inference profile, and API base URL; values are saved before testing the API key.

- Bug Fixes
  - Setup: fixed initial user-name seeding; validate AWS region/API URL before enabling Begin; trim values before saving.
  - Templates: show “Uses structured format” when a built‑in has no prompt; removed an unused isLoading; disable “Make Default” while the request is pending.

<sup>Written for commit a2eba91ab072603a88e93f3753e544aa91e2a029. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -129,6 +129,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -498,6 +499,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -521,6 +523,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -787,7 +790,6 @@
       "dev": true,
       "license": "BSD-2-Clause",
       "optional": true,
-      "peer": true,
       "dependencies": {
         "cross-dirname": "^0.1.0",
         "debug": "^4.3.4",
@@ -809,7 +811,6 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
-      "peer": true,
       "dependencies": {
         "graceful-fs": "^4.2.0",
         "jsonfile": "^6.0.1",
@@ -2985,6 +2986,7 @@
       "integrity": "sha512-o4PXJQidqJl82ckFaXUeoAW+XysPLauYI43Abki5hABd853iMhitooc6znOnczgbTYmEP6U6/y1ZyKAIsvMKGg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.10.4",
         "@babel/runtime": "^7.12.5",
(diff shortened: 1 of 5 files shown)


Optic00:
Reviewed the current head (`09cc302`). The grid redesign is a nice improvement, and the "Make Default" button you added on the card footer already covers the concern I had about locked builtins becoming unreachable once the old "Default template" dropdown was removed — good catch there.

A few things left to look at before merge:

1. **`app/renderer/src/routes/Setup.tsx:181-182`** — `cloudReady`/`canBegin` only checks `cloudApiKey.trim().length > 0`. Picking "AWS Bedrock" or "Custom (OpenAI-compatible)" with a key but no region / no API base URL still enables "Begin Setup" (the blank field write is silently skipped at [line 257](https://github.com/ruzin/stenoai/blob/09cc302/app/renderer/src/routes/Setup.tsx#L257)/[260](https://github.com/ruzin/stenoai/blob/09cc302/app/renderer/src/routes/Setup.tsx#L260)), so the user only finds out something's missing from an opaque connection failure at `testCloudApi` instead of an inline validation message.

2. **`app/renderer/src/routes/Settings.tsx:1350`** — the new prompt-preview text falls back to `'No prompt provided.'` whenever `t.prompt` is empty. The locked "Standard" template ships with `prompt: ""` by design (`src/templates.py:31` — it routes through the structured JSON-schema path, not a free-form prompt), so the fully-working default template now reads as broken/unconfigured in the grid.

3. **`app/renderer/src/routes/Setup.tsx:257-260`** — `bedrockProfile` is trimmed before saving (`.trim()`) but `bedrockRegion` and `apiUrl` aren't. A pasted value with trailing whitespace goes straight into the Bedrock runtime hostname, so a stray space in the region turns into a DNS/connection failure instead of a clear validation error.

4. **`package-lock.json`** (new file at repo root) — empty `packages: {}`, and there's no root `package.json` for it to belong to (only `app/package.json`). Looks like `npm install` got run from the repo root by mistake. Combined with the unrelated `"peer": true` churn across ~15 entries in `app/package-lock.json`, might be worth dropping both from this PR — neither is related to the templates/provider work.

5. Minor: the "Make Default" button inside the template editor header (`Settings.tsx` ~1493) doesn't have a `disabled={setDefault.isPending}` guard, unlike the one you just added to the card footer — a fast double-click can fire the mutation twice with no feedback that the first click landed.

Nothing here blocks the core UX, but 1/2/3 are worth a quick fix and 4 is a one-line revert of the stray file.

Optic00 (approved):
All review findings addressed in a2eba91 (validation gap, misleading prompt text, trim consistency, stray lockfile, pending guard). Ran an additional adversarial security pass on the new Bedrock/custom-provider fields — nothing blocking this PR specifically; the one real finding (Bedrock region string isn't validated before being interpolated into the request URL) is a pre-existing gap in `bedrock_converse_url()`, not introduced here, so filed separately as #299 rather than holding this up.

Nice redesign, thanks for the quick turnaround on the fixes.

Issue #279, opened 2026-07-03
Empty summary after reprocess/generate-report on very short (~2min) meeting


Optic00:
Reported by Ben: recorded a ~2 minute meeting on a Mac (v0.5.7 release DMG), transcript exists, but both "Regenerate notes" and "Generate report" end up back at "No summary available for this meeting" - no error banner shown. UI briefly shows "analyzing transcript" then reverts to the empty state.

## Investigation so far

Initially suspected this was the same class of bug fixed on `chore/architecture-review-followup` (dropped/corrupted stdout protocol lines under chunk-boundary timing in the `reprocess-meeting` handler) - but traced through `simple_recorder.py`'s `reprocess()` command (~line 1944) and ruled that out:

- `streamed_chunks` (the list that becomes the persisted `streamed_md`, `simple_recorder.py:2005-2035`) is collected Python-side, independent of what Electron parses from stdout. A dropped/garbled `CHUNK:` line on the Electron side would only corrupt the *live preview*, not the persisted file.
- `STREAM_COMPLETE` is printed only after the file write (`simple_recorder.py:2130`, per the write-before-complete fix for the Windows CRLF race), so completion signal timing isn't the issue either.

**More likely root cause:** for the `.md` format path, `md_lines.append(streamed_md)` (`simple_recorder.py:2099`) writes whatever `streamed_md` is, even if it's an empty string. If `summarize_transcript_streaming()` (`src/summarizer.py:1515`) yields zero chunks for this specific short transcript (e.g. an edge case in prompt construction, or an empty/short LLM response), the note gets saved with a genuinely empty summary, `STREAM_COMPLETE` still fires "successfully", and the UI shows no error - matching the reported symptom exactly (brief "analyzing" flash, no error banner, ends empty).

Not confirmed - no local repro yet, no access to the debug/processing log from the affected session. Duration was ~2 minutes, non-chunked path (`_needs_chunking` should be false for this length, so this is the direct `_create_markdown_prompt` + `_stream_completion` path, not map-reduce).

## Next steps

- Try to reproduce locally with a very short recording + reprocess.
- Check `_stream_completion` / the local Ollama call path for what could cause an empty response on very short transcripts (model warm-up race, empty-response handling, prompt edge case).
- If reproduced: add a guard so an empty/near-empty streamed result is treated as a failure (STREAM_ERROR) rather than a silent "success" with no content - so the existing "Summary generation failed" banner surfaces instead of a silent empty note.

[...]

PR #300, opened 2026-07-04
fix: validate Bedrock region shape to prevent host redirection

### LINKED ISSUE #299: Bedrock region field allows host redirection via user-supplied string (no format validation)
## Problem

`bedrock_converse_url()` (`src/summarizer.py:35-58`) builds the Bedrock REST endpoint by directly interpolating the user-supplied region into the URL authority:

```python
return f"https://bedrock-runtime.{region}.amazonaws.com/model/{encoded}/converse"
```

`Config.set_bedrock_region()` (`src/config.py:995-1004`) accepts any non-empty string by design — the comment there says "Bedrock validates the region on the wire, so a typo surfaces as a clear 404 / DNS error at request time rather than silently here." That reasoning holds for an honest typo, but not for a region value crafted to redirect the request: a value like

```
x@127.0.0.1:8443/
```

turns the URL authority into `127.0.0.1:8443` (via the `user@host` URL syntax) while the request still attaches the real Bedrock bearer credential — sending it to whatever host is embedded in the region string instead of AWS, silently (no 404, no DNS error, just a request to the wrong host).

Found while running an adversarial security pass (Codex) over #298, which added a second UI entry point (the setup wizard) to this already-existing region-config path — the gap predates that PR and isn't introduced by it.

## Proposed fix

Validate the region against the real AWS region-code shape before persisting it (e.g. `^[a-z]{2}(-gov)?-[a-z]+-\d$`) in `set_bedrock_region()`, and/or in the renderer inputs (Setup.tsx, Settings.tsx AiTab) before it's ever sent over IPC. Reject anything that doesn't match rather than passing it straight into the URL builder.

## Scope

Backend (`src/config.py`, `src/summarizer.py`) plus the two renderer entry points that write this value (`app/renderer/src/routes/Setup.tsx`, `app/renderer/src/routes/Settings.tsx`).

Optic00:
## Summary

Fixes #299.

`Config.set_bedrock_region()` accepted any non-empty string, and `bedrock_converse_url()` (`src/summarizer.py`) interpolated it directly into the request URL's host segment. A region value using `user@host` URL syntax (e.g. `x@127.0.0.1:8443/`) would silently redirect the Bedrock request — with the real bearer credential attached — to an unintended host, instead of the expected 404/DNS error a genuine typo would produce.

Found while running an adversarial security pass (Codex) on #298, which added a second UI entry point (the setup wizard) to this pre-existing region-config path — the gap predates that PR.

## Fix

Both the config layer (`set_bedrock_region`) and the actual network-request sink (`bedrock_converse_url`) now reject anything not shaped like a real AWS region code, via a shared `BEDROCK_REGION_RE` in `src/config.py` (imported by `src/summarizer.py`) so the two checks can't drift apart — matching the existing pattern in this file for keeping the URL-encoding rules centralised.

A format check rather than an exhaustive region allowlist, so new AWS regions don't need an app update; a format-valid-but-wrong region still surfaces as a normal 404/DNS error, per the original design intent.

## Test plan

- [x] New tests: `set_bedrock_region` rejects malformed values (`x@127.0.0.1:8443/`, path-traversal-shaped, freeform text) and still accepts real AWS region shapes (`us-east-1`, `eu-west-2`, `us-gov-west-1`, `cn-northwest-1`, `ca-central-1`)
- [x] New tests: `bedrock_converse_url` raises `ValueError` on the same malformed inputs, independent of the config-layer guard
- [x] `python -m unittest discover tests` — 359 tests pass
- [x] `ruff check` — clean

---
## Summary by cubic
Validate Bedrock region format to block host redirection. Both the config setter and URL builder now enforce real AWS region shapes with ASCII digits and full-match, so typos still fail as normal 404/DNS errors.

- **Bug Fixes**
  - Added shared `BEDROCK_REGION_RE` in `src/config.py`; used by config and URL sink.
  - `Config.set_bedrock_region` trims input and rejects empty or malformed regions.
  - `bedrock_converse_url` validates with `fullmatch` and raises `ValueError` on invalid input.
  - Tests cover valid AWS shapes and reject redirect-shaped, freeform, Unicode-digit, and trailing-newline cases.

<sup>Written for commit f3802c81d81e0a22855cb7b5c599ba1869ec8937. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import platform
+import re
 import shutil
 import sys
 import tempfile
@@ -20,6 +21,15 @@ from src.whisper_models import SUPPORTED_WHISPER_MODELS as _WHISPER_REGISTRY
 from src import templates as _templates
 
 logger = logging.getLogger(__name__)
+
+# AWS region shape: <code>[-gov]-<name>-<digit(s)>, e.g. us-east-1, eu-west-2,
+# us-gov-west-1, cn-northwest-1. Centralised here (not just format-checked at
+# one call site) because both this config layer and bedrock_converse_url()
+# (src/summarizer.py, which imports this) must agree — a region string
+# crafted to redirect the request via `user@host` URL syntax (e.g.
+# "x@127.0.0.1:8443/") has to be rejected by both, not just whichever one an
+# attacker didn't think to route around. See issue #299.
+BEDROCK_REGION_RE = re.compile(r"^[a-z]{2}(-gov)?-[a-z]+-\d{1,2}$")
 
 
 def _atomic_write_json(path: Path, payload) -> None:
@@ -993,12 +1003,17 @@ class Config:
         return value.strip()
 
     def set_bedrock_region(self, region: str) -> bool:
-        """Persist the AWS region. Accepts any non-empty string — Bedrock
-        validates the region on the wire, so a typo surfaces as a clear
-        404 / DNS error at request time rather than silently here."""
+        """Persist the AWS region. A legitimate typo (e.g. "us-eest-1")
+        still surfaces as a clear 404 / DNS error at request time rather
+        than silently here — but the value must at least be shaped like a
+        real AWS region code, so a string crafted to redirect the request
+        to a different host (`user@host` URL syntax) can't be saved."""
         cleaned = (region or "").strip()
         if not cleaned:
             logger.error("Bedrock region cannot be empty")
+            return False
+        if not BEDROCK_REGION_RE.match(cleaned):
+            logger.error(f"Rejected malformed Bedrock region: {cleaned!r}")
             return False
         self._config["bedrock_region"] = cleaned
         return self._save()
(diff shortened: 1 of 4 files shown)


Optic00:
Ran a second Codex pass on this before merge (per house rule for security-relevant changes). Two things came back:

1. **Regex too loose for a security boundary** — `match()+"$"` still allows a trailing `\n`, and `\d` is Unicode-aware so it accepted digit lookalikes (e.g. Arabic-Indic `١`). Neither gap re-opens the original host-redirection attack, but both are unwanted slack in a check like this. Fixed in f3802c8: switched to `fullmatch()` with `[0-9]` instead of `\d`.
2. **Pre-existing, separate bug**: Bedrock (and OpenAI/Anthropic) streaming exceptions are swallowed in `_stream_completion()` and silently produce an empty "successful" summary instead of a visible error — same failure class as #279, just for the cloud path. Not introduced by this fix (confirmed: the same broad `except Exception: ... return` already existed and would've swallowed the old DNS-failure case identically), so filed separately as #301 rather than scope-creeping this PR.

Both new commits keep `python -m unittest discover tests` green (362 tests) and `ruff check` clean.

ruzin (approved):

PR #289, opened 2026-07-04
refactor: #278 - split the 2.8k-line Settings.tsx monolith by tab

Optic00:
## #278 item #6 - split the Settings monolith by tab

`app/renderer/src/routes/Settings.tsx` was the renderer's one emerging monolith at **2809 lines** (next largest file is 1685). This splits it by tab so it stops growing unbounded.

### What changed
Purely mechanical, behaviour-preserving refactor - code was **moved**, not rewritten:
- `Settings.tsx`: **2809 -> 150 lines**. Now a thin shell owning only the `TABS` registry, `TabId` type, tab state (incl. `?tab=` deep-link seeding), the `TabButton` chrome, header/scroll layout, and the version footer. Renders one `<XTab />` per tab.
- New `app/renderer/src/routes/settings/` dir, one file per tab: `GeneralTab`, `TranscriptionTab`, `AiTab`, `TemplatesTab`, `OrganisationTab`, `AdvancedTab`, `DeveloperTab`. Single-consumer helpers (e.g. `OAuthPrompt`, `CopyableValue`, provider-config blocks) are co-located with their one caller.
- Three shared modules for the genuinely cross-tab bits: `primitives.tsx` (`SettingRow`, `SectionHeading`, compact-button class consts), `languages.ts` (the whisper/parakeet language option lists), `model-card.tsx` (`ModelCard` + `PullProgressBar` + size/percent formatters).

### Shared state
Neither context nor prop-drilling was needed: every tab already pulls its own data through TanStack Query hooks (`useSettings`/`useAi`/`useModels`/`useOrg`/`useTemplates`), so there is no mutable state flowing from the shell into tabs. Only stateless helpers are shared, imported directly. Adding a `SettingsContext` would have been ceremony with nothing to carry.

### Preserved (verified present in the new files, no renames)
All e2e/T1 anchors: `data-testid="settings-page"`, `role="tablist"` + the tab button label text, `data-settings-tab="..."`, `data-testid="user-name-input"|"theme-select"|"org-sign-in-card"|"org-signed-in-card"`, `data-oauth-prompt`, `aria-label="Auto-back up new notes to org"`, and all Select `role="option"` label text. These back `org-lock.t1/.t2`, `shared-notes-policy.t1`, `chat-model-indicator.t1`, `org-lock-lifecycle.t2`, `parakeet-language-picker.t1`.

### Verification
[...]

PR #311, opened 2026-07-05
fix(settings): four small pre-existing bugs found during the #289 split review

### LINKED ISSUE #304: Settings > Advanced: Reset storage location button doesn't actually clear the custom path
## Problem

The "Reset" button next to the storage location in Settings > Advanced doesn't actually reset anything. It hides itself but leaves the custom path persisted.

## Root cause

`resetFolder` in `Settings.tsx` (currently lines 2703-2707 on `main`; will live in `app/renderer/src/routes/settings/AdvancedTab.tsx` once #289 merges):

```js
const resetFolder = () => {
  if (storage.data?.default_path) {
    setStorage.mutate(storage.data.default_path);
  }
};
```

It mutates the storage path *to the default path itself*, which `set-storage-path` treats as a custom path. That's enough to make the "Reset" button disappear (the UI condition for showing it is "path differs from default"), but the backend still stores it as a user-set custom path rather than clearing back to "no override" (`''`).

## Proposed fix

Mutate with `''` (or whatever `set-storage-path` expects for "use default") instead of `storage.data.default_path`.

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so it's a pre-existing bug, not something introduced by the split.

### LINKED ISSUE #305: Settings > Organisation: Enter-key sign-in bypasses busy/empty-field guard; sign-out doesn't clear stale error
## Problem

Two small inconsistencies in Settings > Organisation sign-in/sign-out:

1. Pressing Enter in the password field can submit sign-in even when required fields are empty or a request is already in flight, bypassing the disabled-state guard on the sign-in button.
2. Signing out doesn't clear a previously-shown error, so a stale error from an earlier failed sign-out can flash when the view switches back to the sign-in card.

## Root cause

In `Settings.tsx` (currently on `main`; will live in `app/renderer/src/routes/settings/OrganisationTab.tsx` once #289 merges):

Enter-key handler (~line 2654), mirrors none of the button's disabled checks:
```js
onKeyDown={(e) => {
  if (e.key === 'Enter') void onSignIn();
}}
```

`onSignOut` (~lines 2501-2507), unlike `onSignIn`/`onGoogleSignIn`, never clears `error` before mutating:
```js
const onSignOut = async () => {
  try {
    await logoutMutation.mutateAsync();
  } catch (e) {
    setError(e instanceof Error ? e.message : String(e));
  }
};
```

## Proposed fix

- Guard the Enter-key handler with the same `busy`/required-field checks used to disable the sign-in button.
- Call `setError(null)` at the start of `onSignOut`, matching `onSignIn`/`onGoogleSignIn`.

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so pre-existing, not introduced by the split.

### LINKED ISSUE #308: Settings > Templates: failed template delete surfaces as unhandled promise rejection
## Problem

Deleting a template in Settings > Templates: if the delete request fails, the failure surfaces as an unhandled promise rejection instead of a clear, recoverable error state in the confirmation dialog.

## Root cause

In `Settings.tsx` (currently on `main`, ~lines 1446-1448; will live in `app/renderer/src/routes/settings/TemplatesTab.tsx` once #289 merges):

```js
onConfirm={async () => {
  if (!deleteTarget) return;
  await del.mutateAsync(deleteTarget.id);
  setDeleteTarget(null);
}}
```

`await`-ing `mutateAsync` without a catch means any rejection bubbles up as an unhandled rejection in the dialog's click handler.

## Proposed fix

Use `del.mutate(deleteTarget.id, { onSuccess: () => setDeleteTarget(null) })` instead of `await mutateAsync`, matching the pending/close behaviour without throwing.

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so pre-existing, not introduced by the split.

### LINKED ISSUE #309: Settings: delete-model icon button missing aria-label
## Problem

The icon-only "delete model" button in Settings (model card) relies on `title` text for its accessible name, which isn't reliably exposed to keyboard/screen-reader users.

## Root cause

`Settings.tsx` (currently on `main`, ~line 639; will live in `app/renderer/src/routes/settings/model-card.tsx` once #289 merges):

```jsx
<button
  type="button"
  onClick={onDeleteModel}
  title="Delete this model to free up disk space"
  className="flex size-[28px] cursor-pointer items-center justify-center rounded-[6px] border-0 bg-transparent"
  style={{ color: 'var(--fg-muted)' }}
>
```

## Proposed fix

Add an explicit `aria-label="Delete this model to free up disk space"` alongside the existing `title` (keep `title` for the hover tooltip).

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so pre-existing, not introduced by the split.

Optic00:
## What & why

While reviewing #289 (the mechanical Settings.tsx tab split), cubic flagged 8 behavioral/a11y issues in the extracted code. All 8 were confirmed present identically in `main`'s current monolithic `Settings.tsx` before the split — pre-existing bugs, not introduced by #289. This PR fixes 4 of the 6 follow-up issues filed for them (the small, independent ones); a sibling PR covers the other 2 (OAuth cancel race + name-seed race in `GeneralTab.tsx`).

Stacked on `chore/278-settings-split` (#289) since the code only exists in that shape there — this will auto-retarget to `main` once #289 merges.

## What's in this PR

- **#304** — Settings > Advanced: "Reset" storage-location button now actually clears the custom path (`mutate('')` instead of `mutate(default_path)`), with a UI-driven e2e test proving it (mutation-tested: fails against the old handler, passes against the fix).
- **#305** — Settings > Organisation: Enter-key sign-in now respects the same busy/required-field guard as the button; sign-out clears a stale prior error, matching sign-in.
- **#308** — Settings > Templates: a failed template delete now surfaces a visible error and keeps the dialog open for retry, instead of an unhandled promise rejection.
- **#309** — Settings: the icon-only delete-model button gets an explicit `aria-label` alongside its `title`.

Reviewed by Codex (`/codex:review` against this branch's base) — no actionable findings.

## Test plan

- [x] `npm run typecheck:renderer` clean
- [x] `npm run lint:renderer` clean (no new warnings)
- [x] `npm run test:unit` — all unit/component tests green, including new tests per fix
- [x] New e2e test for #304 (`settings-roundtrip.t2.spec.ts`) — verified locally, including a mutation test (temporarily reverted the fix, confirmed the test fails; restored, confirmed it passes)

Fixes #304, Fixes #305, Fixes #308, Fixes #309

---
## Summary by cubic
Fixes four pre-existing Settings bugs found during the tab-split review to improve reliability, error handling, and accessibility. Adds focused unit and e2e tests to prevent regressions.

- **Bug Fixes**
  - Advanced: Reset storage location now truly clears the custom path (sends ''), with a real UI e2e test and a unit test.
  - Organisation: Enter-to-sign-in respects disabled/busy guards; sign-out clears stale errors. Tests cover both paths.
  - Templates: Failed delete shows an error and keeps the dialog open for retry (no unhandled rejection), with test coverage.
  - Models: Icon-only delete-model button now has aria-label="Delete model" for screen readers, with an accessibility test.

<sup>Written for commit 4d3592b7653ae7c83fcd45e4756329722e54435e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,9 @@ app/playwright-report/
 .claude/
 .github/skills/
 
+# Local git worktrees (agent workspaces, not project content)
+.worktrees/
+
 # OS
 .DS_Store
 Thumbs.db
--- a/app/renderer/src/routes/settings/AdvancedTab.test.tsx
+++ b/app/renderer/src/routes/settings/AdvancedTab.test.tsx
@@ -1,0 +1,65 @@
+import { describe, test, expect, beforeEach, vi } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { AdvancedTab } from './AdvancedTab';
+
+// Reset button on the storage-location row must clear the custom override by
+// sending '' to the backend (set-storage-path treats '' as "use default"),
+// NOT the default path itself — which the backend records as a fresh custom
+// override, so "Reset" would visually hide itself while leaving the override
+// in place (#304).
+
+const setStoragePath = vi.fn(async () => ({ success: true }));
+
+vi.mock('@/lib/ipc', () => ({
+  ipc: () => ({
+    settings: {
+      getStoragePath: async () => ({
+        success: true,
+        storage_path: '/custom/path',
+        default_path: '/default/path',
+        custom_path: '/custom/path',
+      }),
+      setStoragePath,
+      pickStorageFolder: async () => ({ success: true, folderPath: null }),
+      getTelemetry: async () => ({
+        success: true,
+        telemetry_enabled: false,
+        anonymous_id: null,
+      }),
+      setTelemetry: async () => ({ success: true }),
+    },
+    system: { clearState: async () => ({ success: true }) },
+  }),
+}));
+
+function renderTab() {
+  const qc = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false, gcTime: Infinity },
+      mutations: { retry: false },
+    },
+  });
+  return render(
+    <QueryClientProvider client={qc}>
+      <AdvancedTab />
+    </QueryClientProvider>,
+  );
+}
+
+describe('AdvancedTab storage reset', () => {
+  beforeEach(() => {
+    setStoragePath.mockClear();
+  });
+
+  test('Reset clears the custom override with an empty string, not the default path', async () => {
+    renderTab();
+
+    // Reset only appears once the storage query resolves with a custom path.
+    const reset = await screen.findByRole('button', { name: 'Reset' });
+    fireEvent.click(reset);
+
+    await waitFor(() => expect(setStoragePath).toHaveBeenCalledTimes(1));
+    expect(setStoragePath).toHaveBeenCalledWith('');
+  });
+});
--- a/app/renderer/src/routes/settings/AdvancedTab.tsx
+++ b/app/renderer/src/routes/settings/AdvancedTab.tsx
@@ -74,11 +74,11 @@ export function AdvancedTab() {
     }
   };
 
-  const resetFolder = () => {
-    if (storage.data?.default_path) {
-      setStorage.mutate(storage.data.default_path);
-    }
-  };
+  // Send '' — the backend treats an empty path as "use the default location"
+  // and clears the custom override. Passing the default *path* instead would
+  // be recorded as a fresh custom override, so Reset would hide itself without
+  // actually resetting anything (#304).
+  const resetFolder = () => setStorage.mutate('');
 
   const custom =
     storage.data?.custom_path &&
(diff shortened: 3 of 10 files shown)


ruzin (approved):
Reviewed (rebased onto main post-split): four small settings-tab fixes (reset-storage clears custom path, org sign-in Enter guard + stale-error clear, template-delete failure surfacing, delete-model a11y name) + tests. CI green. LGTM.

PR #312, opened 2026-07-05
fix(settings): OAuth cancel race + name-seed race in GeneralTab (#289 split follow-up)

### LINKED ISSUE #306: Settings > General: OAuth connect flow races (Cancel doesn't cancel handshake; late rejection can overwrite dismissed prompt)
## Problem

Two races in the calendar OAuth connect flow on Settings > General:

1. Clicking "Cancel" on the OAuth prompt dismisses the modal but doesn't cancel the in-flight OAuth handshake — the connection can still complete (or later error) after the user believed they'd backed out.
2. A late connect rejection always writes error state, even if the user already dismissed the prompt or started connecting a different provider in the meantime — a stale/irrelevant error can reopen the modal.

## Root cause

In `Settings.tsx` (currently on `main`; will live in `app/renderer/src/routes/settings/GeneralTab.tsx` once #289 merges):

`startConnect` catch handler (~lines 878-895) unconditionally sets error state on any rejection, including a user-triggered cancel:
```js
const startConnect = async (provider: 'google' | 'outlook') => {
  setOauth({ provider, state: 'pending' });
  try {
    if (provider === 'google') await google.connect.mutateAsync();
    else await outlook.connect.mutateAsync();
  } catch (err) {
    setOauth({ provider, state: 'error', ... });
  }
};
```

`OAuthPrompt`'s `onClose` (~line 984) just clears local dialog state, it never calls a cancel mutation on the provider connect flow:
```js
<OAuthPrompt
  state={oauth}
  onClose={() => setOauth(null)}
  onRetry={() => oauth && void startConnect(oauth.provider)}
/>
```

## Proposed fix

- Have Cancel invoke the provider's `cancel` mutation before clearing dialog state, so the backend handshake actually stops.
- In the catch handler, ignore a `Cancelled` rejection and only write error state if `oauth` is still `{ provider, state: 'pending' }` for that same provider (i.e. nothing else has superseded it).

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so pre-existing, not introduced by the split.

### LINKED ISSUE #307: Settings > General: display-name field seed can clobber early keystrokes on slow load
## Problem

On a slow load of Settings > General, typing into the display-name field right after the page opens can have your keystrokes silently overwritten once the fetched name arrives.

## Root cause

In `Settings.tsx` (currently on `main`, ~lines 832-846; will live in `app/renderer/src/routes/settings/GeneralTab.tsx` once #289 merges), the seed effect only guards against re-seeding after it already ran once — not against the user having already typed something before the fetch resolved:

```js
const [nameDraft, setNameDraft] = React.useState('');
const nameSeededRef = React.useRef(false);
...
if (nameSeededRef.current) return;
if (userName.isPending || userName.isPlaceholderData) return;
if (userName.data !== undefined) {
  setNameDraft(userName.data);
  nameSeededRef.current = true;
}
```

If the user starts typing before `userName.data` resolves, the effect still fires once the query settles and clobbers the draft.

## Proposed fix

Only seed if the draft is still untouched, e.g.:
```js
setNameDraft((current) => (current === '' ? userName.data : current));
```

## Provenance

Found by cubic during review of PR #289 (mechanical Settings.tsx tab split). Confirmed present identically on `main` before that PR, so pre-existing, not introduced by the split.

Optic00:
## What & why

Sibling PR to #311 — same origin: cubic flagged pre-existing bugs during the #289 (Settings.tsx tab split) review, confirmed present identically in `main` before the split. This PR covers the two riskier ones, both in `GeneralTab.tsx`'s calendar OAuth connect flow and name-field seeding, so they got their own PR and extra review passes.

Stacked on `chore/278-settings-split` (#289) — will auto-retarget to `main` once #289 merges.

## What's in this PR

- **#306** — OAuth connect/cancel race:
  - Cancel now actually cancels the in-flight handshake (calls the provider's `cancel` mutation) instead of just closing the dialog — previously the loopback OAuth server kept running and could silently complete the connection after the user backed out.
  - A synchronous ref lock replaces an `isPending`-based in-flight guard that could race on rapid double-clicks (same weakness as the `Home.tsx` reference pattern this was ported from — found by Codex review here and fixed properly).
  - Stale/cancelled rejections (`'Cancelled'` and otherwise) can no longer resurrect a dismissed dialog or overwrite a different/fresher attempt for the same provider — the latter needed an attempt-token check, not just provider/state (also a Codex finding).
- **#307** — the display-name field's seed-from-fetch race:
  - No longer clobbers keystrokes typed before the initial fetch resolves.
  - The dirty guard now releases once the edit is committed (blur), instead of permanently blocking re-sync for the component's whole lifetime — a second Codex finding on the first fix.

Went through 3 rounds of Codex review (`/codex:review` against this branch's base); the last pass reported no actionable findings.

## Test plan

- [x] `npm run typecheck:renderer` clean
- [x] `npm run lint:renderer` clean (no new warnings)
- [x] `npm run test:unit` — 41/41 green, including new deterministic component tests for every race described above (each written test-first, watched fail against the unfixed code)
- [ ] Manual dev-app sanity pass (connect → Cancel → confirm no error reappears) — not run in this environment (needs the PyInstaller-bundled backend); behavior is covered by the component tests instead

Fixes #306, Fixes #307

---
## Summary by cubic
Fixes two races in Settings → General. Canceling calendar OAuth now aborts the backend handshake and prevents stale errors (including same‑provider retries) or double-connects; the display-name field no longer gets overwritten by a late seed and re-syncs after commit. Fixes #306 and #307.

- **Bug Fixes**
  - Calendar OAuth: Cancel calls the provider `cancel` mutation, immediately releases the in-flight lock, ignores "Cancelled" rejections, and uses an attempt token so late errors can’t resurrect or overwrite a newer attempt. A synchronous ref lock prevents double-click double-connects.
  - Display name: Early typing isn’t clobbered by a late seed. The dirty guard clears on blur/Enter so the field re-syncs from `userName.data` after a committed edit.
  - Added deterministic unit tests in `GeneralTab.test.tsx`.

<sup>Written for commit e0aff13a36dcaef5cf53f80da57258e1f75acded. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/routes/settings/GeneralTab.test.tsx
+++ b/app/renderer/src/routes/settings/GeneralTab.test.tsx
@@ -192,6 +192,37 @@ describe('GeneralTab OAuth connect/cancel race (#306)', () => {
     expect(screen.queryByText("Couldn't connect to Google")).toBeNull();
   });
 
+  test("a stale same-provider retry's late rejection does not overwrite the fresh attempt", async () => {
+    const firstAttempt = h.defer<void>();
+    // First Google attempt gets a controllable promise; the retry falls back to
+    // the default never-resolving mock, so it stays pending.
+    h.google.connect.mutateAsync.mockReturnValueOnce(firstAttempt.promise);
+    render(<GeneralTab />);
+
+    await act(async () => {
+      clickConnect('Google');
+    });
+    await act(async () => {
+      clickCancel();
+    });
+    // Immediately retry the SAME provider — a fresh pending attempt.
+    await act(async () => {
+      clickConnect('Google');
+    });
+    expect(screen.getByText('Connecting to Google')).toBeTruthy();
+
+    // The original (abandoned) attempt rejects late with a non-Cancelled error.
+    // provider + state still superficially match the fresh attempt, so only the
+    // attempt token can tell them apart.
+    await act(async () => {
+      firstAttempt.reject(new Error('Timed out — no response from Google.'));
+      await firstAttempt.promise.catch(() => {});
+    });
+
+    expect(screen.getByText('Connecting to Google')).toBeTruthy();
+    expect(screen.queryByText("Couldn't connect to Google")).toBeNull();
+  });
+
   test('two synchronous Connect clicks fire only one connect mutation', async () => {
     render(<GeneralTab />);
     const btn = screen.getByRole('button', { name: 'Google' });
--- a/app/renderer/src/routes/settings/GeneralTab.test.tsx
+++ b/app/renderer/src/routes/settings/GeneralTab.test.tsx
@@ -209,3 +209,23 @@ describe('GeneralTab OAuth connect/cancel race (#306)', () => {
     expect(h.google.connect.mutateAsync).toHaveBeenCalledTimes(1);
   });
 });
+
+describe('GeneralTab name-field seed race (#307)', () => {
+  test('typing before the userName query resolves is not overwritten by the seed', async () => {
+    h.userName = { data: undefined, isPending: true, isPlaceholderData: false };
+    const { rerender } = render(<GeneralTab />);
+
+    const input = screen.getByTestId('user-name-input') as HTMLInputElement;
+    fireEvent.change(input, { target: { value: 'Ruzin' } });
+    expect(input.value).toBe('Ruzin');
+
+    // The canonical value now arrives from disk.
+    h.userName = { data: 'Old Name', isPending: false, isPlaceholderData: false };
+    await act(async () => {
+      rerender(<GeneralTab />);
+    });
+
+    // The user's in-progress edit must survive the late seed.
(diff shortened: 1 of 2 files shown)


ruzin (approved):
Reviewed (rebased onto main post-split): GeneralTab OAuth-cancel race + name-seed race fixes with tests. CI green. LGTM.

PR #313, opened 2026-07-05
feat(recording): transcription pill dock — recording coexists with the app

ruzin:
## Summary

Reworks the live-recording + notes UX around a **coexisting transcription pill**, **stop-is-the-new-pause**, **continue-recording**, and a decoupled **My notes** tab. Merged up to date with `main`.

Spec: `.specs/2026-07-10-transcription-pill-dock.md` (local; zoro-cut, user-approved) + its round-2 addendum.

### Transcription pill dock (coexistence)
- Recording is **status-driven, not route-driven** (`PrimaryDock`): the compact Granola-style pill (wave + elapsed + expand + stop glyph) docks left of the Ask bar; on `chat`/`settings`/`processing` it docks alone; the Ask bar renders visible-but-disabled ("Chat available after recording").
- **New note** navigates to the live-note editor so you can write notes while recording; background triggers (hotkey/tray/auto) coexist silently. Coexist = the pill *follows* you, not that you start nowhere.
- Recording **wins the dock slot** over ProcessingDock (back-to-back notes), so Stop is never unreachable.

### Stop-is-pause + continue-recording
- **No manual pause** anywhere (pill or expanded footer) — Stop ends the segment. Resume appears only when the *system* auto-paused (sleep / meeting-app mic drop).
- **Continue-recording**: a note detail offers a mic that records *into* that note. Backend `process-streaming --append-to` folds the segment's transcript in with a `--- Resumed HH:MM ---` separator, extends duration, marks `notes_stale: true`, and skips summary/title; silent/failed continuations exit non-zero leaving the note untouched. The floating CTA then reads **Regenerate notes**; `reprocess` clears the stale flag.

### My notes tab (decoupled notes layer)
- Studied meetily's separate `meeting_notes` store; kept Steno file-based but treat `## User Notes` as an **independent, always-editable layer**. `MeetingDetail` gains a **Summary / My notes** switcher + autosaving editor (debounced, flush on blur). `update-meeting` upserts the tail `## User Notes` section (summary + transcript stay byte-intact); a light `useUpdateUserNotes` avoids a full list re-scan per keystroke.

### Also
- Silence auto-stop default **15 → 2 min**. Single Generate-notes CTA (in-note duplicate removed). MECHANICS.md refresh; `.specs/` gitignored.
- Merged `origin/main` (#318–#329): unified the `start-recording-ui` signature to `(name, trigger, appendTo)` — main's analytics `trigger` + this branch's `appendTo` coexist.

### Tests
- New `pill-dock.t1` (coexistence, adjacent row, disabled Ask bar, chrome/processing precedence, expand/collapse, stop→processing, auto-pause rescue, Whisper variant, continue affordance, New-note navigation), `my-notes.t1` (tab switch + autosave + reload) and `my-notes.t2` (real `update-meeting` body surgery + real Python parser round-trip), `continue-recording.t2` (append + regenerate + failed-continuation contract). Full T1 (28) green post-merge; typecheck clean.
- **Verified at runtime** against the real app + backend (record → pill next to disabled Ask bar, expand → live panel, stop → processing; My notes autosave persists to disk).

### Deferred (own follow-up)
- **Instant stop / kill the "Analysing transcript" interstitial** — transcript-first save-at-stop so the note shows immediately (Regenerate if a summary existed). To be zoro-spec'd separately.
- Live chat against an in-progress recording.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ko3LY6o46uEpW8PWDM8gYN

---

### Update: instant stop + note-icon cleanup (this push)
- **Instant stop** — stopping a Parakeet recording lands you on the note immediately (no "Analysing transcript" takeover): main writes a placeholder note from the live transcript, the pipeline upgrades transcript + summary in place; `processing` flag cleared on success/failure/startup sweep; My-notes edits during the window are preserved. Auto-summarise streams the summary inline on the note. Whisper/import keep the processing dock. Spec: `.specs/2026-07-12-instant-stop.md`. Tests: `instant-stop.t1` + `instant-stop.t2`.
- **Note-list icons** — replaced the hash-colored monograms (arbitrary chroma that collided with the reserved status hues) with a single neutral note glyph, per DESIGN.md paper+ink.


Optic00:
## Review (independent pass, 8-angle finder + verify)

Nice mechanism overall (the bridge does close the double-fire/dead-`streamCache` bug from the first draft for the two existing triggers). Found a few gaps specific to the new floating trigger, worth fixing before/along with merge:

### 1. Floating button skips the `transcription_failed` guard the other two triggers respect
`MeetingDetail.tsx:531`
```js
const summaryPending = notesNotGenerated && Boolean(meeting.transcript);
```
The header "Regenerate" icon is wrapped `{!info.transcription_failed && (...)}` (line 599, with an explicit comment about why: "reprocess would exit non-zero and strand the UI on a spinner"), and the in-body CTA only renders when `transcriptionFailed` is false (the ternary at line 873 checks it first). `summaryPending` doesn't inherit that exclusion, so if a note ever has `transcription_failed: true`, `notes_generated: false`, and a non-empty `transcript`, the page shows "Transcription failed" and hides its own controls — while the floating button still renders enabled and calls `startReprocess`. Suggest ANDing in `!transcriptionFailed` (or better, deriving `summaryPending` from the same condition the render ternary already uses, so there's one source of truth).

### 2. Floating button's disabled-state lag reopens a narrow double-fire window
The header/in-body buttons read `reprocess.isPending || streamPhase !== 'idle'` directly off local state (disables in the same render as the click). The floating button's `disabled` comes from the bridge, which only updates via a `useEffect` one commit later — a real window where it still looks clickable. Checked `main.js`'s `reprocess-meeting` handler: `activeReprocessJobs` has no same-file dedupe (its own comment says overlapping jobs are expected — for different files; nothing blocks the same file twice). A fast double-click on the floating button specifically could fire two concurrent `stenoai reprocess` processes for the same summary file. Might be worth a client-side guard (e.g. disable synchronously on click via a ref before the effect catches up) or a server-side same-file lock.

### 3. Floating button also disables during unrelated report generation
`ReportSwitch` renders whenever `reportTemplates.length > 0` (`MeetingDetail.tsx:837`), not gated on having a summary. `onGenerateReport` sets the same `streamPhase` that feeds the bridge's `streaming` flag, so on a pending note with a custom template available, generating a report disables the floating "Generate notes" button too, with no visual explanation.

### Smaller stuff
- `reprocessBridgeStore.ts` reimplements the publish-on-mount/clear-on-unmount pattern `askBarContext.tsx`'s `useActiveMeeting` already has — could probably be a field on `AskBarContext` instead of a second store, especially since the roadmap (step 2, the dock rework) will likely want more of these remote-trigger surfaces.
- `TranscriptBar` and `GenerateNotesBar` each get their own `BottomDockSlot` even though they're mutually exclusive on `transcriptOpen` — one shared slot swapping its child (like the primary dock already does for Live/Processing/Ask) would save the duplicate `useSidebarCollapsed`/`useSidebarWidth` subscriptions.
- The T1 spec's "hides for a normal note" case uses `SEED_MEETING`, which has no `notes_generated` field at all (so it takes the "no summary" path, not the "has a real summary" path) — doesn't actually cover a summarized note.

Happy to take a pass at 1-3 if useful.

ruzin:
@Optic00 thanks — three of these were real; addressed in 8a0b187 (the branch has since grown into the full pill-dock rework, see the updated PR description):

**1. `transcription_failed` guard — fixed.** `summaryPending` now mirrors the render ternary: `notesNotGenerated && !transcriptionFailed && Boolean(meeting.transcript)`. The floating trigger can no longer offer reprocess on a note whose own controls are hidden.

**2. Double-fire window — fixed, client-side.** `startReprocess` now has a synchronous re-entrancy guard using `streamCache` (a module-level Map written synchronously in the same call), which can't lag the way the bridge's effect-published `disabled` does. Guarding in `startReprocess` itself covers all three triggers, not just the floating one. Went client-side rather than a main.js same-file lock because main's overlapping-jobs model is deliberate and shared with cross-file concurrency — a renderer guard at the single choke point is smaller blast radius.

**3. Disabled during report generation — accepted behavior, not fixing.** `streamPhase` gates the whole detail's `StreamingView`; a reprocess started mid-report-generation would fight the same streaming surface. Disabling the floating CTA whenever *any* stream occupies the detail is the conservative-correct behavior; the button re-enables when the stream settles. Agreed there's no visual explanation, but adding one for a transient state on a pending-note-with-custom-templates edge didn't seem worth the copy.

**Smaller stuff:**
- *Bridge vs `askBarContext` field* — keeping the store. The dock rework (now in this PR) made the primary dock status-driven and the bridge is consumed outside the AskBar provider's conceptual scope; a zustand store with an ownership-checked `clear()` stays the smaller coupling.
- *Shared 72-band slot* — real but micro (two extra sidebar-width subscriptions); noted as a follow-up rather than folded into an already-large PR.
- *T1 seed fidelity — fixed.* `SEED_MEETING` now carries a non-empty summary, so the "hides for a normal note" case genuinely takes the has-summary path (verified `buildTranscriptBundle`/transcript-export don't read `summary`; full T1 still 22/22).

Optic00 (changes requested):
## Re-review (full pill-dock scope, after 8a0b187)

Round-2 fixes check out: the `transcription_failed` guard now mirrors the render ternary, and I verified the `startReprocess` re-entrancy guard is genuinely synchronous (`streamCache.set` before `reprocess.mutate`), so it covers all triggers. Nice.

New scope, new findings though - the first one explains your red T2 lanes. Independent pass by me plus a GPT-5.5 review as second angle; everything below is hand-verified against the code, and finding 1 is reproduced locally with a debugger-level proof.

### 1. Blocker - T2 red on macOS and Windows: `GenerateNotesBar` compares two different representations of the same path

`auto-summarize.t2` times out waiting for `generate-notes-dock-button`. Reproduced locally; instrumenting the two sides of the bar's gate gives:

```
activeSummaryFile (route):        /var/folders/.../output/notesoff_summary.md
bridge summaryFile (payload):     /private/var/folders/.../output/notesoff_summary.md
```

`GenerateNotesBar` renders only when `summaryFile === activeSummaryFile`. But the two values have different origins:

- `activeSummaryFile` is the **raw route param** (`MeetingDetail` -> `useActiveMeeting(summaryFile)`).
- The bridge's `summaryFile` is `info.summary_file` from the **get-meeting payload**, and the `get-meeting` handler canonicalises via `validateMeetingFilePath()` (realpath) and builds the meeting from `realResolved` - so the payload carries the symlink-resolved path.

Any storage dir behind a symlink breaks the strict equality: every macOS temp dir (`/var` -> `/private/var`, which is why both e2e lanes fail deterministically), Windows 8.3 short paths, and - the product case - any user whose custom storage path involves a symlink. Since d886e9d removed the in-note CTA, an affected user has **no way to generate notes at all**.

Suggested fix at the identity source rather than a compare-time normalisation: have `MeetingDetail` feed the same canonical value to both sides, e.g. `useActiveMeeting(meeting.data?.session_info.summary_file ?? summaryFile, ...)`, so the whole renderer keys on the payload's canonical path. (The old in-note CTA never compared paths, which is why this first shows up now.)

### 2. Processing route loses the real job on back-to-back recording

Start a new recording while on `/meetings/processing`: `useRecording().sessionName` flips to the new recording, `Processing.tsx:39-42` re-binds `activeSession` to it, and the `e.sessionName !== activeSession` filters (`:89-105`) then discard the completion/failure events of the job that is actually summarising. The page stops tracking its own job. Keeping `activeSession` pinned to `currentJob` while `hasRecording` is true would fix it.

### 3. `GenerateNotesBar` can overlap the expanded live transcript panel

The bar's gate only checks the saved-note `transcriptOpen`. Viewing a transcript-only note during a recording and expanding the live transcript leaves the floating CTA rendered on top of the panel (the panel grows well past the 72px band). Gate it on the live-transcript-open state (or on recording) as well.

### Non-blocking

- **Silence auto-stop default 15 -> 2 min** deserves its own visibility rather than riding along in a UI PR: 2 minutes of bilateral silence happens inside real meetings (shared reading, breaks), so this will split some meetings into two notes - and the new default silently applies to every existing user who never persisted the setting. Also worth connecting to #271: with background system audio the detector currently never fires, so the aggressive default doesn't help the motivating "forgotten recording" case much, while raising false-positive risk where detection does work. 5 min might be the safer step.
- **Typed draft is lost on expand**: `PrimaryDock` swaps the whole row for `LiveTranscriptBar`, unmounting `AskBar` - exactly the remount the stable-tree comment guards against. Type a draft, start recording, expand, collapse: draft gone. Keeping the row mounted (visually hidden) under the panel would close it.
- Nit: if the start IPC takes >2s, an interval poll can still clobber the optimistic `hasRecording: true` (`cancelQueries` only cancels in-flight fetches, not future ticks). Only bites on slow starts; fine as a follow-up.

T1 coverage of the new surface is good, and the MECHANICS.md refresh is genuinely useful.

Optic00 (changes requested):
## Re-review (pill dock + instant stop, head 9a6cb21)

Big picture: the pill-dock coexistence model and the instant-stop flow are good product moves, and finding 1 from my last round is properly fixed (the bridge now publishes under the route identity, with a clear comment - nice). But the new My-notes/continue-recording surface has two data-loss bugs that I'd consider blockers, one of which the PR itself accidentally arms. Independent pass by me plus a GPT-5.6 review as second angle; everything below is hand-verified against the code.

### 1. Blocker - `update-meeting` silently truncates any note whose body contains `---`, and this PR guarantees such notes exist

`app/main.js:3116` still parses the file with:

```js
const parts = raw.split('---', 3);
```

Unlike Python's `maxsplit`, the JS `limit` **discards the remainder**. Proof:

```js
> "---\ntitle: x\n---\nSummary\n\n--- Resumed 10:00 ---\n\n## Transcript\n\nhello".split('---', 3)[2]
"\nSummary\n\n"   // transcript gone
```

`body = parts[2]` is then upserted and written back at `main.js:3161-3165`, so everything from the first `---` inside the body to EOF is deleted on disk.

This line is pre-existing on main, but it used to be near-unreachable (rename only, and summaries rarely contain a thematic break). This PR arms it twice:

- `_append_segment_to_note` writes a `--- Resumed HH:MM ---` separator into the Transcript section of **every continued note**.
- The My notes tab autosaves through `update-meeting` (debounced, per typing pause).

So: continue a recording, then type anything into My notes on that note -> the transcript from the separator onward (the whole resumed segment and anything after it) is silently destroyed, and each further autosave keeps rewriting the truncated file.

The fix already exists in this same PR: `clearNoteProcessingFlag` (`main.js:4421-4428`) splits with no limit and rejoins the tail, with a comment explaining exactly this failure mode. `update-meeting` needs the same treatment. A unit test with a `---` in the body would lock it in.

### 2. Blocker - "My-notes edits during the window are preserved" doesn't hold: the pipeline snapshots the notes at start and writes them back minutes later

`process-streaming` reads the placeholder's `## User Notes` **once, at command start** (`simple_recorder.py:868`), then transcribes and summarises (minutes for a real meeting), then writes that stale `notes_text` back in the final save (Step 4, ~line 1170). Any My-notes edit made during the upgrade window - which is precisely the scenario the instant-stop design promises to support - is clobbered by the final rewrite.

`instant-stop.t2` doesn't catch this because it makes the edit before the pipeline starts, i.e. before the snapshot. Suggested fix: move the `_read_existing_user_notes` call to immediately before each of the two writes (Step 4 and the auto-summarize-off early save at ~line 1065) instead of command start. That shrinks the race from minutes to milliseconds; see also finding 5.

### 3. Startup sweep runs before the custom storage path is loaded

`setImmediate(sweepStuckProcessingFlags)` (`main.js:1286`) fires on the next event-loop turn, but `_cachedCustomStoragePath` is only set after an awaited backend spawn (`main.js:1455`). So the sweep always scans the **default** `userData/output`, and a user with a custom storage path keeps permanently-"processing" notes after a crash - the exact "finishing up forever" state the sweep exists to prevent. The import-marker sweep a few lines below even documents that it MUST run after the storage-path load; this one needs the same ordering.

### 4. `--append-to` accepts any existing file under the allowed roots, including `config.json`

`start-recording-ui` validates the append target with `validateSafeFilePath(appendTo, getAllowedBaseDirs())` + `existsSync` (`main.js:4521-4524`). The allowed roots include the whole userData dir and the project root, so a renderer-supplied `config.json` (or any JSON in there) passes, and `_append_segment_to_note`'s non-`.md` branch will happily rewrite it into a "note". Renderer input is untrusted by this codebase's own standard (#329 just hardened meeting paths for the same reason) - this should require a canonical, output-scoped `*_summary.md`/`.json`, i.e. mirror `validateMeetingFilePath` rather than the generic base-dir check.

### 5. Unlocked read-modify-write on the note file (known-limitation at minimum)

`_append_segment_to_note` and the final pipeline rewrite are whole-file read-modify-write cycles that race the `update-meeting` autosave writer, last-writer-wins. Finding 2 is the wide instance; even with it fixed a small window remains. Fine to ship with, but worth a comment in the code so the next writer doesn't widen it again.

### 6. Still open from last round

- **Processing route re-binds to a new recording** (`Processing.tsx:35-40`): start recording B while note A is on `/meetings/processing` and the `e.sessionName !== activeSession` filters discard A's completion/failure events. Exposure shrank (Parakeet now takes the instant-stop path) but Whisper and imports still land here.
- **GenerateNotesBar overlaps the expanded live transcript panel**: the bar gates on the saved note's `transcriptOpen` only, not `liveTranscriptOpenStore` (`GenerateNotesBar.tsx:18-22`), so it floats over the expanded `LiveTranscriptBar`.

### Minor

- Append target deleted between start and spawn: `main.js:4001` silently drops `--append-to` and the continuation becomes a **new** meeting. Non-destructive, but it contradicts the "failed continuation exits non-zero, note untouched" contract - a notification (or the same non-zero path) would be more honest.
- `upsertUserNotesSection` and `_read_existing_user_notes` both match the **first** occurrence of `\n## User Notes`. If a summary or transcript ever contains that literal line, the JS side truncates from there and the Python side reads the wrong tail. Anchoring on the last occurrence would match the "always the last section" invariant.
- Product question, not a bug: silence auto-stop default 15 -> 2 min. Two minutes of bilateral silence is common in real meetings (muted mic while nobody talks, screen-share reading pauses), and we already have auto-stop complaints on file (#271). Is 2 min deliberate? 5 would feel safer as a default.

Positives worth keeping: the atomic placeholder write, `clearNoteProcessingFlag`'s frontmatter surgery (do copy it into finding 1), the head-only frontmatter scan in the sweep, and the four new specs - `instant-stop.t2` just needs the mid-window edit case from finding 2.

Happy to re-review quickly once 1+2 are addressed; 3+4 I'd also want before merge, the rest is judgment call.

Optic00:
Re-checked at head `e053826`. The 15 Jul commits touch the `get-meeting` read path (`parseMeetingMarkdown`) and surface the notes_stale/processing flags, but they don't reach the write paths my round-3 findings live in, so 1-4 are still open at head. Logging it here so it doesn't get lost - no new blockers:

1. **`update-meeting` `---` truncation - still open.** main.js:3130 is still `raw.split('---', 3)` then `body = parts[2]`, written back at 3175-3179. The correct unlimited-split/rejoin already sits next door in `parseMeetingMarkdown` (1902) and `clearNoteProcessingFlag` (4435); `update-meeting` just needs the same. Still armed by `_append_segment_to_note` writing a `--- Resumed HH:MM ---` separator into every continued note (simple_recorder.py:744).
2. **Stale My-notes snapshot - still open.** `process_recording_streaming` reads the notes once at start (simple_recorder.py:867-870); both writes (early save 1065-1070, Step 4 1170-1175) reuse that snapshot. Re-reading immediately before each write is still needed.
3. **Startup sweep before storage-path load - still open.** `setImmediate(sweepStuckProcessingFlags)` (main.js:1286) still runs before `_cachedCustomStoragePath` is set (1451-1459); the import sweep right below it is sequenced correctly (1466-1472).
4. **`--append-to` accepts any allowed-root file - still open.** main.js:4535-4538 is still the generic base-dir + `existsSync` check, so a `config.json` under userData passes into `_append_segment_to_note`'s JSON-rewrite branch.

Happy to re-review the moment the write paths are touched.

ruzin:
@Optic00 all comments addressed, merging

PR #317, opened 2026-07-06
fix: strip reasoning blocks from AI outputs

Vassista:
This PR hardens the output sanitization layer to ensure that reasoning blocks (e.g., `<think>...</think>`) from models never leak into user-facing views or exports. It expands support for additional model tags, preserves original Markdown formatting more cleanly, and applies the sanitizer universally across all UI components and features.

**Changes:**
- **Sanitizer Improvements:** `stripReasoning` now supports `<thinking>`, `<thought>`, and `<reasoning>` tags. It preserves legitimate Markdown indentation and cleanly removes blank-line residue after a block is stripped.
- **Backend Parity:** Updated the Markdown normalizer (`simple_recorder.py`) to handle the expanded set of reasoning tags, keeping frontend and backend parsing logic in sync.
- **Universal Application:** Reasoning blocks are now consistently stripped from all user-facing paths. This resolves previous leaks in custom report rendering, shareable note uploads, clipboard copying, search indexing, and list previews.
- **Testing:** Expanded unit tests to cover the newly supported tags, formatting preservation, blank-line cleanup, and relevant streaming edge cases.

### DIFF
--- a/app/renderer/src/components/home/PreviousRow.tsx
+++ b/app/renderer/src/components/home/PreviousRow.tsx
@@ -2,6 +2,7 @@ import { Folder as FolderIcon, Loader2 } from 'lucide-react';
 import type { Meeting } from '@/lib/ipc';
 import { navigate } from '@/lib/router';
 import { useMeetingsList } from '@/lib/meetingsListContext';
+import { stripReasoning } from '@/lib/markdown';
 
 interface PreviousRowProps {
   meeting: Meeting;
@@ -166,7 +167,7 @@ function formatDuration(seconds?: number): string | undefined {
 }
 
 function previewText(meeting: Meeting): string | undefined {
-  const summary = meeting.summary?.trim();
+  const summary = meeting.summary ? stripReasoning(meeting.summary).trim() : undefined;
   if (summary) return summary;
   const kp = meeting.key_points?.[0];
   if (typeof kp === 'string' && kp.trim()) return kp.trim();
--- a/app/renderer/src/lib/markdown.test.ts
+++ b/app/renderer/src/lib/markdown.test.ts
@@ -3,7 +3,7 @@ import { stripReasoning } from '@/lib/markdown';
 
 /**
  * Unit coverage for stripReasoning — the guard that removes `<think>` /
- * `<thought>` reasoning blocks emitted by some local models before the text is
+ * `<thought>` / `<thinking>` / `<reasoning>` reasoning blocks emitted by some local models before the text is
  * handed to the markdown renderer. The streaming case (an unclosed tag at the
  * end of a chunk) is the load-bearing edge: a half-streamed reasoning block
  * must not flash into the rendered summary.
@@ -16,9 +16,6 @@ describe('stripReasoning', () => {
 
   test('strips an unclosed reasoning block to the end of a streaming chunk', () => {
     const input = 'visible intro\n<think>still thinking and the chunk cut off here';
-    // The closing tag never arrives, so everything from <think> onward is dropped
-    // (only the trailing newline before the tag survives — strip trims leading,
-    // not trailing, whitespace).
     const result = stripReasoning(input);
     expect(result).toBe('visible intro\n');
     expect(result).not.toContain('still thinking');
@@ -32,5 +29,35 @@ describe('stripReasoning', () => {
   test('returns falsy/empty for empty input', () => {
     expect(stripReasoning('')).toBeFalsy();
     expect(stripReasoning('')).toBe('');
+  });
+
+  test('handles multiple blocks and nested/malformed tags', () => {
+    const input = 'A\n<think>1</think>\nB\n<thinking>2</thinking>\nC';
+    expect(stripReasoning(input)).toBe('A\nB\nC');
+  });
+
+  test('handles reasoning tags', () => {
+    const input = '<reasoning>...</reasoning>content';
+    expect(stripReasoning(input)).toBe('content');
+  });
+
+  test('handles nested tags (outer tag wins)', () => {
+    const input = '<think>outer<thinking>inner</thinking></think>content';
+    expect(stripReasoning(input)).toBe('content');
+  });
+
+  test('preserves spaces and formatting outside blocks', () => {
+    const input = '  # Hello \n\n<think>test</think>\nWorld';
+    expect(stripReasoning(input)).toBe('  # Hello \n\nWorld');
+  });
+
+  test('preserves leading formatting at document start when no reasoning block is present', () => {
+    const input = '  <think>test</think>\n  # Hello';
+    expect(stripReasoning(input)).toBe('  # Hello');
+  });
+
+  test('handles uppercase tags', () => {
+    const input = '<THINK>A</think>B';
+    expect(stripReasoning(input)).toBe('B');
   });
 });
--- a/app/renderer/src/lib/markdown.test.ts
+++ b/app/renderer/src/lib/markdown.test.ts
@@ -41,7 +41,7 @@ describe('stripReasoning', () => {
     expect(stripReasoning(input)).toBe('content');
   });
 
-  test('handles nested tags (outer tag wins)', () => {
+  test('handles different nested tags (outer tag wins)', () => {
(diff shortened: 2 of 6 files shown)


Optic00 (commented):
## Review

Good PR - the two closed gaps are real (reports rendered `activeReport.content` raw, and `composeShareBody` uploaded the raw summary), and I checked the new regex against the tricky cases (case-insensitive backreference, the `$(?![\s\S])` end-of-string anchor under `/m`, the unclosed streaming block, backtracking behaviour on large unmatched input) - all sound. Two remaining gaps of the same class the PR is fixing, plus two smaller notes:

### 1. `copyNotes` still copies the raw summary to the clipboard
`MeetingDetail.tsx:483` - `const summary = meeting.summary?.trim()` goes into `navigator.clipboard.writeText` unfiltered. Same pattern this PR fixes in `composeShareBody`, one screen up. With `meeting.summary = "<think>private chain</think>\nPublic summary"`, the rendered note and the share upload are clean, but "Copy notes" puts the reasoning block on the clipboard. One-line fix, ideally in the same PR.

### 2. List preview and search snippets render the raw summary
- `PreviousRow.tsx` (`previewText`, ~line 169) shows `meeting.summary?.trim()` as the row preview on Home.
- `CommandPalette.tsx:184` passes raw `m.summary` into `snippet()` (`noteSearch.ts:33`), which renders at line 206. This one is double: the *search itself* also matches text inside reasoning blocks, so a query hit inside `<thinking>...</thinking>` surfaces reasoning text as the result snippet.

A summary that *starts* with a reasoning block shows that text in the meeting list and palette even though the detail view is now clean. Wrapping both sources in `stripReasoning` fixes display; whether search should also match stripped text only is a judgement call (I'd say yes, same input for index and snippet).

### 3. Backend tag list now lags the renderer
`simple_recorder.py:79` - `_REASONING_TAG_HEADER_PATTERN` normalizes headers after `think|thought` only. The renderer now also knows `thinking|reasoning`, so `<thinking>...</thinking>### Action Items` inline would skip the parser normalization. Not a blocker, but worth syncing the lists (or extracting one source of truth) so they don't drift further.

### 4. Correction on the PR description (doesn't change the value of the PR)
The regex on `main` has no `/m` flag - `/<(think|thought)>[\s\S]*?(?:<\/\1>|$)/gi` - so `$` already matched end-of-string, and it strips closed multi-paragraph blocks and unclosed streaming blocks correctly (I ran both cases against the old code to be sure). What the new version actually improves: the two extra tag names, no more blank-line residue between stripped blocks, and preserved indentation instead of `trimStart()`. Worth adjusting the description so the changelog doesn't record a bug that wasn't there.

### Nits
- The "handles nested tags" test covers `<think>outer<thinking>inner</thinking></think>` (different tags) but not same-tag nesting: `<think>a<think>b</think>c</think>` leaks `c</think>` with old and new regex alike. Pre-existing limitation and probably fine for real model output - just don't let the test name suggest it's covered.
- Dropping `trimStart()` is a deliberate contract change (the tests pin it), looks right to me - flagging it only so it's a conscious decision.

Happy to take 1-3 as a follow-up if you'd rather keep this PR small.

Vassista:
> ## Review
> Good PR - the two closed gaps are real (reports rendered `activeReport.content` raw, and `composeShareBody` uploaded the raw summary), and I checked the new regex against the tricky cases (case-insensitive backreference, the `$(?![\s\S])` end-of-string anchor under `/m`, the unclosed streaming block, backtracking behaviour on large unmatched input) - all sound. Two remaining gaps of the same class the PR is fixing, plus two smaller notes:
> 
> ### 1. `copyNotes` still copies the raw summary to the clipboard
> `MeetingDetail.tsx:483` - `const summary = meeting.summary?.trim()` goes into `navigator.clipboard.writeText` unfiltered. Same pattern this PR fixes in `composeShareBody`, one screen up. With `meeting.summary = "<think>private chain</think>\nPublic summary"`, the rendered note and the share upload are clean, but "Copy notes" puts the reasoning block on the clipboard. One-line fix, ideally in the same PR.
> 
> ### 2. List preview and search snippets render the raw summary
> * `PreviousRow.tsx` (`previewText`, ~line 169) shows `meeting.summary?.trim()` as the row preview on Home.
> * `CommandPalette.tsx:184` passes raw `m.summary` into `snippet()` (`noteSearch.ts:33`), which renders at line 206. This one is double: the _search itself_ also matches text inside reasoning blocks, so a query hit inside `<thinking>...</thinking>` surfaces reasoning text as the result snippet.
> 
> A summary that _starts_ with a reasoning block shows that text in the meeting list and palette even though the detail view is now clean. Wrapping both sources in `stripReasoning` fixes display; whether search should also match stripped text only is a judgement call (I'd say yes, same input for index and snippet).
> 
> ### 3. Backend tag list now lags the renderer
> `simple_recorder.py:79` - `_REASONING_TAG_HEADER_PATTERN` normalizes headers after `think|thought` only. The renderer now also knows `thinking|reasoning`, so `<thinking>...</thinking>### Action Items` inline would skip the parser normalization. Not a blocker, but worth syncing the lists (or extracting one source of truth) so they don't drift further.
> 
> ### 4. Correction on the PR description (doesn't change the value of the PR)
> The regex on `main` has no `/m` flag - `/<(think|thought)>[\s\S]*?(?:<\/\1>|$)/gi` - so `$` already matched end-of-string, and it strips closed multi-paragraph blocks and unclosed streaming blocks correctly (I ran both cases against the old code to be sure). What the new version actually improves: the two extra tag names, no more blank-line residue between stripped blocks, and preserved indentation instead of `trimStart()`. Worth adjusting the description so the changelog doesn't record a bug that wasn't there.
> 
> ### Nits
> * The "handles nested tags" test covers `<think>outer<thinking>inner</thinking></think>` (different tags) but not same-tag nesting: `<think>a<think>b</think>c</think>` leaks `c</think>` with old and new regex alike. Pre-existing limitation and probably fine for real model output - just don't let the test name suggest it's covered.
> * Dropping `trimStart()` is a deliberate contract change (the tests pin it), looks right to me - flagging it only so it's a conscious decision.
> 
> Happy to take 1-3 as a follow-up if you'd rather keep this PR small.

Thanks for the thorough review and for checking the regex edge cases.

I'm working on fixes for the remaining issues you called out:

* `copyNotes` copying the raw summary
* list preview / command palette using the raw summary (including search/snippet behavior)
* syncing the backend reasoning tag list with the renderer

I'll also update the PR description to accurately reflect what the regex change improves, and I'll clarify the nested-tag test name so it doesn't imply same-tag nesting is supported.

Vassista:
* `copyNotes` now strips reasoning blocks before copying.
* Meeting list previews and search indexing/snippets now use the sanitized summary, so reasoning content is no longer surfaced there.
* Synced the backend reasoning tag list with the frontend to keep them aligned.
* I also updated the PR description to accurately reflect the final behavior and renamed the nested-tags test to clarify that it covers different-tag nesting rather than same-tag nesting. 

Let me know if you spot something else.

Optic00:
Re-checked the follow-up commits - all four points and the nits are addressed:

1. `copyNotes` strips the summary before it hits the clipboard.
2. `PreviousRow.previewText` and `noteSearch` (both `searchNotes` and `snippet`) now work on stripped text, so search matching and result snippets agree - the judgement call landed where I'd hoped.
3. Backend `_REASONING_TAG_HEADER_PATTERN` is synced to the four-tag list.
4. Description + test name corrected.

I also re-ran the new regex shape (`(?:^[ \t]*)?...\n?` with the `startsWithReasoning` guard) against the edge cases from the first pass - indentation preservation, blank-line residue, unclosed streaming block, uppercase tags, cross-tag nesting - all behave as the tests pin them. LGTM.

**Coordination note:** #318 (just opened) rewrites the same `copyNotes` block - it extracts the payload builder to `lib/notesCopy.ts` and fixes "Copy notes" to copy the open template report instead of the Standard note. It already applies `stripReasoning` to both copy sources (summary + report content), matching this PR's direction. So whichever merges second has a small conflict in `MeetingDetail.tsx` with a safe resolution: take #318's `copyNotes` (the strip behaviour from this PR is preserved there). Everything else in the two PRs is disjoint.

ruzin (approved):

PR #318, opened 2026-07-06
fix(meetings): Copy notes copies the open template report, not the Standard note

Optic00:
## Problem

With a generated template report open in MeetingDetail, the header "Copy notes" action still copied the **Standard** structured note. The clipboard silently disagreed with what was on screen.

## Fix

- Extract the clipboard builder into `app/renderer/src/lib/notesCopy.ts` (pure, unit-tested): when a report is open it copies title + date/duration + that report's markdown; otherwise the Standard sections exactly as before (byte-identical output for the standard path).
- Disable "Copy notes" while a summary/report stream is on screen (`streamPhase !== 'idle'`, all transient states) — same bug class: the body shows in-flux streamed text while the clipboard would get the old note.

## Coverage

- Vitest unit tests for `buildNotesCopyText` (standard path, report path, empty-section omission, whitespace-only report).
- New T1 e2e spec `copy-notes-report.t1.spec.ts` driving the real pill switch + Copy notes action against a seeded meeting carrying a report. New `STENOAI_E2E_SEED_REPORT` seam in the mock IPC (env-gated so `transcript-export.t1` keeps its report-less meeting), with a stateful `set-active-report` mock so the pill switch survives the post-mutation refetch like the real sidecar.

Full T1 suite 19/19, vitest 78/78, typecheck + lint clean.

## Not changed (by design)

Template reports intentionally live in the `<stem>_reports.json` sidecar, not as standalone `.md` files (`src/report_store.py`) — "view containing folder" not showing a report md is the designed storage model, out of scope here.

---
## Summary by cubic
Fixes “Copy notes” to copy the open template report (not always the Standard note), strips hidden reasoning blocks, and disables the action while streaming to avoid stale copies.

- **Bug Fixes**
  - Copies title + date/duration + the active report’s markdown; otherwise falls back to the Standard structured note.
  - Strips reasoning blocks (e.g., <think>) from both Standard summaries and report content before copying, matching what’s rendered.
  - Disables “Copy notes” when `streamPhase !== 'idle'` to prevent clipboard/display mismatches.

- **Refactors**
  - Extracted builder to `app/renderer/src/lib/notesCopy.ts` with unit tests.
  - Added e2e spec `copy-notes-report.t1.spec.ts`; mock IPC now supports `STENOAI_E2E_SEED_REPORT` and a stateful `set-active-report` (seeded report includes a <think> block to verify stripping).

<sup>Written for commit 37926a71177a04c92f8ee1663afa9fe6906c81bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -49,6 +49,23 @@ const SEED_MEETING = {
   discussion_areas: [],
 };
 
+// A generated template report attached to SEED_MEETING when
+// STENOAI_E2E_SEED_REPORT=1 (copy-notes-report T1). Gated separately so the
+// transcript-export T1 keeps seeing the report-less meeting it asserts on.
+const SEED_REPORT = {
+  id: 'rep_e2e_1',
+  template_id: 'tpl_e2e_status',
+  template_name: 'Status Report',
+  model: 'mock-model',
+  content: '## Status Report\n\n- Pipeline healthy\n- Next: open the reqs',
+  created_at: '2026-06-19T13:00:00Z',
+};
+
+const seededMeeting = () =>
+  process.env.STENOAI_E2E_SEED_REPORT === '1'
+    ? { ...SEED_MEETING, reports: [SEED_REPORT], active_report: null }
+    : SEED_MEETING;
+
 function install({ ipcMain }) {
   // In-memory stand-in for the org session + provider config that the real
   // handlers persist to disk. Mutated by the org-login / org-logout / set-ai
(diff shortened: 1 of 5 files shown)


Optic00:
Heads-up on overlap with #317: that PR's latest revision also touches `copyNotes` (stripping reasoning blocks from the copied summary). This PR already incorporates that behaviour for both copy sources (Standard summary and open report content), so if #317 merges first the `MeetingDetail.tsx` conflict resolves by taking this branch's `copyNotes`; if this merges first, #317's `copyNotes` hunk becomes redundant. Details in https://github.com/ruzin/stenoai/pull/317#issuecomment-4893387104.

ruzin (approved):

ruzin (approved):

PR #320, opened 2026-07-06
Feature/UI enhancement

Vassista:
## Description
This PR replaces the standard macOS system notifications with a highly polished, custom floating window. 

### Key Features
- **Premium Aesthetic:** A clean, flat design that floats seamlessly under the macOS menu bar.
- **Smart "Join & Take Notes" Action:** Clicking the action button now simultaneously opens the meeting link in the browser *and* automatically starts a StenoAI recording session.
- **Focus App:** Clicking anywhere on the notification body dismisses it and instantly focuses the main StenoAI window.

### Screenshots

CCF6BC9C-FF2D-423E-9F9F-189D5272D6A3

---
## Summary by cubic
Replaces macOS system notifications with a compact in‑app toast under the menu bar and adds one‑click “Join & take notes.” The toast mounts as its own React root for faster load and tighter focus/close behavior.

- **New Features**
  - Pre‑meeting alerts show in a small, always‑on‑top window; clicking the body focuses the main app and dismisses it.
  - “Join & take notes” opens the meeting URL in the browser, starts a recording, then closes the toast.
  - New `/notification` route with `NotificationToast` rendered at the root and wired to `show-notification`; added `notification.close()` IPC.

- **Bug Fixes**
  - Fixed window focus IPC, rare stuck toasts, and a race in the notification close handler; auto‑close after 15s with proper cleanup.
  - Resolved React Hook/ref issues across the UI (e.g., `ImportDropZone`, `MeetingsShell`, `QuitDialog`, `Processing`, `Chat`); fixed an infinite render loop in `MeetingDetail` and restored streaming text animation by reverting `firstNewIdx` to a ref.
  - Reverted unused CSP img‑src widening and stabilized Processing/Chat timers to prevent stale bucketing and stalled elapsed time.

<sup>Written for commit e18b736608e539c50f1b7ebfbdba5f028af61c7f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -8298,7 +8298,7 @@ function createNotificationWindow(event) {
     resizable: false,
     skipTaskbar: true,
     focusable: false,
-    hasShadow: false, // The React component renders its own shadow
+    hasShadow: false,
     backgroundColor: '#00000000',
     webPreferences: {
       nodeIntegration: false,
@@ -8313,7 +8313,6 @@ function createNotificationWindow(event) {
 
   notificationWindow.once('ready-to-show', () => {
     notificationWindow.showInactive();
-    // Ensure the window floats across all macOS spaces/desktops and full-screen apps
     notificationWindow.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
     notificationWindow.setAlwaysOnTop(true, 'screen-saver', 1);
 
@@ -8329,6 +8328,13 @@ function createNotificationWindow(event) {
 ipcMain.handle('close-notification-window', () => {
   if (notificationWindow && !notificationWindow.isDestroyed()) {
     notificationWindow.close();
+  }
+});
+
+ipcMain.handle('focus-main-window', () => {
+  if (mainWindow && !mainWindow.isDestroyed()) {
+    if (!mainWindow.isVisible()) mainWindow.show();
+    mainWindow.focus();
   }
 });
 
--- a/app/main.js
+++ b/app/main.js
@@ -8311,10 +8311,22 @@ function createNotificationWindow(event) {
   const rendererDist = path.join(__dirname, 'renderer', 'dist', 'index.html');
   notificationWindow.loadFile(rendererDist, { hash: '/notification' });
 
+  let autoCloseTimer;
   notificationWindow.once('ready-to-show', () => {
     notificationWindow.showInactive();
     notificationWindow.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
     notificationWindow.setAlwaysOnTop(true, 'screen-saver', 1);
+
+    autoCloseTimer = setTimeout(() => {
+      if (notificationWindow && !notificationWindow.isDestroyed()) {
+        notificationWindow.close();
+      }
+    }, 15000);
+
+    notificationWindow.on('closed', () => {
+      if (autoCloseTimer) clearTimeout(autoCloseTimer);
+      notificationWindow = null;
+    });
 
     notificationWindow.webContents.send('show-notification', {
       title: event.title || 'Meeting starting',
@@ -8328,13 +8340,6 @@ function createNotificationWindow(event) {
 ipcMain.handle('close-notification-window', () => {
   if (notificationWindow && !notificationWindow.isDestroyed()) {
     notificationWindow.close();
-  }
-});
-
-ipcMain.handle('focus-main-window', () => {
-  if (mainWindow && !mainWindow.isDestroyed()) {
-    if (!mainWindow.isVisible()) mainWindow.show();
-    mainWindow.focus();
   }
 });
 
--- a/app/main.js
+++ b/app/main.js
@@ -8311,24 +8311,27 @@ function createNotificationWindow(event) {
   const rendererDist = path.join(__dirname, 'renderer', 'dist', 'index.html');
   notificationWindow.loadFile(rendererDist, { hash: '/notification' });
 
+  const win = notificationWindow;
   let autoCloseTimer;
-  notificationWindow.once('ready-to-show', () => {
-    notificationWindow.showInactive();
-    notificationWindow.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
-    notificationWindow.setAlwaysOnTop(true, 'screen-saver', 1);
+  win.once('ready-to-show', () => {
+    win.showInactive();
+    win.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
+    win.setAlwaysOnTop(true, 'screen-saver', 1);
 
     autoCloseTimer = setTimeout(() => {
-      if (notificationWindow && !notificationWindow.isDestroyed()) {
-        notificationWindow.close();
+      if (!win.isDestroyed()) {
+        win.close();
       }
     }, 15000);
 
-    notificationWindow.on('closed', () => {
+    win.on('closed', () => {
       if (autoCloseTimer) clearTimeout(autoCloseTimer);
-      notificationWindow = null;
+      if (notificationWindow === win) {
+        notificationWindow = null;
+      }
     });
 
-    notificationWindow.webContents.send('show-notification', {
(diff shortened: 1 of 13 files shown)


Optic00 (changes requested):
## Review

Nice direction on the notification redesign, but I'd hold off on merging as-is - one structural issue plus a couple of real bugs bundled into the same diff.

1. **The notification window mounts the entire app, including its singleton side-effect hooks** (`app/renderer/src/App.tsx`) - `useRecordingEvents()`, `useRecordingProcessingEffects()`, `useSystemAudioCapture()`, `useCalendarEvents()`/`useCalendarAuthBus()` all run unconditionally, *before* the `route === '/notification'` check further down. Every time `createNotificationWindow()` (main.js) loads `index.html#/notification`, a second renderer boots the full app tree - including `useSystemAudioCapture`, which spins up its own `MediaRecorder`/`AudioContext` graph whenever recording status is `'recording'`. Pre-meeting notifications fire precisely when the user might already be mid-call in the main window, so this risks a second, independent audio-capture pipeline racing the real one, on top of doubled calendar polling and processing-complete cache writes for the ~15s the toast is alive. Fix: check the route before mounting those hooks, or better, give the notification window its own minimal HTML entry that doesn't pull in the full router/hook tree.

2. **Race in the window-close bookkeeping** (`main.js`, `createNotificationWindow`) - the `'closed'` handler closes over the module-level `notificationWindow` and unconditionally sets it to `null`. If a second pre-meeting event arrives while window A is still alive, the guard at the top calls `A.close()` and immediately reassigns `notificationWindow = B`. `close()` is async, so A's `'closed'` event can fire *after* B is already current - nulling `notificationWindow` out from under B. B is then orphaned: `close-notification-window` and the next dedupe-close silently no-op, and B just sits until its own 15s timer. Capture the instance instead (`const win = notificationWindow; ...; win.on('closed', () => { if (notificationWindow === win) notificationWindow = null; })`).

3. **`Chat.tsx`** - the recents-bucketing `now` moved from `Date.now()` recomputed on every memo pass to `useState(() => Date.now())`, frozen for the component's whole mounted lifetime. If Chat stays mounted across a day boundary, "Today"/"Yesterday" grouping goes stale. Looks like a lint-driven fix (exhaustive-deps) that changed behavior - worth reverting.

4. Smaller, not a blocker: CSP now allows `img-src` from `google.com`/`gstatic.com`/`duckduckgo.com`, but nothing in this diff renders a remote image from those hosts (`attendees` is passed through but never displayed; the camera icon is inline SVG). If it's leftover from a favicon feature that got cut, I'd drop the CSP widening until something actually needs it.

5. Product question, not code: dropping the native `Notification` path means the toast is `alwaysOnTop` + `visibleOnFullScreen: true` unconditionally - no more respecting OS Do Not Disturb, and it'll pop up over a fullscreen screen-share too. Might be exactly what's wanted, but worth an explicit call from @ruzin/@Vassista since it's a real UX behavior change, not just visual polish.

Everything else (the ImportDropZone/MeetingsShell ref-in-effect fixes, the MeetingDetail/FolderPicker/Processing "adjust state during render" rewrites, QuitDialog reorder, StreamingView revert) looks like solid, correct cleanup.

Vassista:
@Optic00  Hi, its an incredibly thorough review. I really appreciate the sharp eyes here. You caught some incredibly subtle edge cases that would have definitely caused problems down the line.  


* **`firstNewIdx` Animation Bug:** Reverted the derived state back to the `useRef` pattern with an ESLint bypass comment to safely maintain the stale block count without immediately forcing a derived-state re-render that strips the animation classes.
* **Notification App Tree Mount:** Moved the notification routing check out of `App.tsx` and directly into `main.tsx`. The notification toast now mounts natively at the React root, entirely sidestepping the heavy app-level hooks and duplicate audio/calendar pipelines.
* **Window Close Race Condition:** Captured the `notificationWindow` instance into a local block-scoped variable (`win`) during creation. The `'closed'` handler now strictly checks `if (notificationWindow === win)` before nullifying the module-level reference, preventing it from orphaning newer pre-meeting windows.
* **`Date.now()` Freezing in Chat:** Moved the `Date.now()` call directly inside the `useMemo` callback instead of freezing it in `useState`. This satisfies the dependency array without freezing the component's concept of "now" across day boundaries.
* **CSP Widening:** Reverted the unused `img-src` widening in `index.html` back to its original secure posture.

and yup, ( btw its @Vassista not arlo over here in GitHub haha )

Optic00 (approved):
Re-reviewed the latest push - all four code points from my earlier review are addressed cleanly:

1. **App-tree mount** - `main.tsx` now branches on `#/notification` at the React root and renders only `<NotificationToast />`, so the toast window no longer boots `App` or any of the singleton side-effect hooks (`useSystemAudioCapture`, `useRecordingEvents`, `useCalendarEvents`, ...). `NotificationToast` itself only pulls in `ipc` + `useTheme`. Exactly the structural fix I was hoping for.

2. **Window-close race** - `const win = notificationWindow` is captured and the `'closed'` handler guards with `if (notificationWindow === win)` before nulling, so a stale window can no longer orphan a newer one.

3. **`Chat.tsx` `now`** - back to `Date.now()` recomputed inside the `useMemo`, so the Today/Yesterday bucketing stays fresh across a day boundary.

4. **CSP** - the `img-src` widening is gone (`index.html` is untouched now).

That's all the blocking items resolved, so I'm approving. Point 5 (dropping the native `Notification` path means the toast ignores OS Do Not Disturb and shows over fullscreen) is still a product call for you and @ruzin rather than a code issue - fine to ship if that's the intended behavior.

One housekeeping note on my end: my original review pinged `@ruzin13/@ArlO` in point 5 - both the wrong GitHub handles. I've edited that comment to `@ruzin/@Vassista`. Thanks for the friendly catch, and sorry for the mixup! 🙂

Vassista:
> Re-reviewed the latest push - all four code points from my earlier review are addressed cleanly:
> 
> 1. **App-tree mount** - `main.tsx` now branches on `#/notification` at the React root and renders only `<NotificationToast />`, so the toast window no longer boots `App` or any of the singleton side-effect hooks (`useSystemAudioCapture`, `useRecordingEvents`, `useCalendarEvents`, ...). `NotificationToast` itself only pulls in `ipc` + `useTheme`. Exactly the structural fix I was hoping for.
> 2. **Window-close race** - `const win = notificationWindow` is captured and the `'closed'` handler guards with `if (notificationWindow === win)` before nulling, so a stale window can no longer orphan a newer one.
> 3. **`Chat.tsx` `now`** - back to `Date.now()` recomputed inside the `useMemo`, so the Today/Yesterday bucketing stays fresh across a day boundary.
> 4. **CSP** - the `img-src` widening is gone (`index.html` is untouched now).
> 
> That's all the blocking items resolved, so I'm approving. Point 5 (dropping the native `Notification` path means the toast ignores OS Do Not Disturb and shows over fullscreen) is still a product call for you and @ruzin rather than a code issue - fine to ship if that's the intended behavior.
> 
> One housekeeping note on my end: my original review pinged `@ruzin13/@ArlO` in point 5 - both the wrong GitHub handles. I've edited that comment to `@ruzin/@Vassista`. Thanks for the friendly catch, and sorry for the mixup! 🙂

totally agree on this, we can use something like `macos-notification-state` module & its Windows equivalent `windows-quiet-hours` for detecting DND status where we can halt the notifications. The one trade-off we'll need to keep in mind is that introducing native C++ dependencies will add some complexity to our build pipeline (e.g., node-gyp compilations and cross-architecture prebuilds), but imho it's worth it to ensure we properly respect user focus modes.

PR #321, opened 2026-07-06
Unify Steno icon, naming, and typeface across app/website/docs

WilliamDrewett:
## Why

Icon, product name, and heading font had drifted out of sync across surfaces. Specifically:

- The app's tray menu and several error dialogs still said **"StenoAI"** while the Dock icon, website, and docs already said **"Steno"** — a prior rebrand (#227) never finished rolling out.
- At least **four separate hand-copied SVG versions** of the dragonfly mark existed across the app, website, and docs, and had drifted out of sync with each other — including an unrelated, off-brand AI-generated glowing dragonfly image sitting on the docs favicon and wordmark, nothing like the actual paper+ink brand system.
- The display serif was **Charter** in most places, not the intended brand typeface.

This PR makes all of it consistent in one pass.

## What changed

- **Icon**: one canonical mark (`brand/mark.svg`) rolled out everywhere — macOS `.icns`, Windows `.ico`, tray icons (idle + recording states), favicons, and the website/docs badges. Background settled on **ink-900 (`#1B1B19`)** to match the existing brand color token, replacing an inconsistent `#3D3D39` that was used in a couple of places.
- **Monochrome tray icon**: the macOS/Windows menu-bar icon is a "template image" — a black shape on a transparent background. The OS automatically re-tints this single asset correctly for both light and dark menu bars, so one asset covers both themes; no separate light/dark variants needed there.
- **Naming**: `"StenoAI"`/`"stenoAI"` → `"Steno"` in all user-facing copy — tray menu, error dialogs, OAuth callback pages, site metadata, legal pages, contributor docs. Every technical identifier is untouched: the `stenoai://` deep-link scheme, bundle ID, data paths, CLI binary name, and release artifact filenames.
- **Typeface**: **Ovo** (self-hosted `woff2`, single 400 weight) replaces Charter as the display/heading serif in the app, website, and docs. Along the way this caught and fixed a real bug on the docs site: Mintlify hard-codes a `font-bold` class on the page title, and since Ovo has no real bold weight, that was forcing the browser to synthesize a fake-bold — headings looked inconsistent as a result.
- **`.gitignore` fix**: a bare `docs/` pattern (meant for an unrelated scratch-notes folder) was silently swallowing new files added to the *real* docs site. Anchored it to the actual folder it was meant for.

image

Screenshot 2026-07-06 at 9 33 34 pm

Screenshot 2026-07-06 at 10 31 47 pm

## Needs a follow-up look

- `website/public/og-image.png` and `readme.png` are **screenshots of the running app**, not logo files — they still show the old UI and need a fresh capture once this merges, not a static asset swap.
- `Sidebar.tsx` and `AppWindowDemo.jsx` each inline their own copy of the mark instead of importing the `AppIcon`/`StenoMark` components that already exist in the same build. Real but minor duplication — left alone to keep this PR's blast radius contained rather than bundling a component refactor in with the rebrand.
- DMG/exe artifact filenames (`stenoAI-macos-arm64.dmg`, etc.) still say `stenoAI` — left deliberately, since renaming touches the release workflow and download links; a decision for a dedicated release-mechanics change, not a copy fix.
- The new mark has sharper stroke caps/joins than the old one — verified this matches the approved source art exactly, so it's not a bug, but flagging since it's a visible style difference from the old logo.

## Test plan

- [x] Self-reviewed the full diff (8-angle pass: correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, CLAUDE.md conventions) — no critical/medium issues; one trivial redundant `!important` fixed.
- [x] `tsc --noEmit` on the renderer passes clean.
- [x] Visually verified live: app sidebar/Dock icon/tray icon (idle + recording, light + dark), website nav/footer/legal pages, docs nav (light + dark).
- [ ] Confirm a full signed DMG build still resolves the icon correctly (verified via an unsigned local `electron-builder --dir` pack in dev).

---
## Summary by cubic
Unifies the Steno brand across app, website, and docs: one canonical dragonfly mark, “Steno” in UI copy, and Ovo for display headings. Also updates the homepage H1.

- **Refactors**
  - Adopt a single `brand/mark.svg` everywhere (.icns/.ico, tray idle/recording template icons, favicons, website/docs badges); unify background to ink-900 `#1B1B19`.
  - Replace “StenoAI” with “Steno” in user-facing copy (tray/menu, dialogs, site metadata, legal, contributor docs, skills). Technical IDs stay the same: `stenoai://` scheme, bundle ID, paths, CLI, artifact names.
  - Switch display serif to Ovo (self-hosted `woff2`) across app/website/docs; fixes Mintlify title fake-bold; adjusts docs nav logo size for legibility.
  - Update homepage hero headline to “AI for your private conversations.”
  - Anchor `.gitignore` rule to `app/renderer/docs/` so repo-root `docs/` isn’t excluded; update tests/copy to match naming.

- **Migration**
  - No breaking changes; links and deep-links are unchanged. After merge, refresh `website/public/og-image.png` and `readme.png` screenshots.

<sup>Written for commit 5c43b7acfd22739ecfb7925bbf3f5719e0df5e71. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/321?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Optic00 (commented):
Reviewed the full diff (plus an independent read to double-check nothing behavioural slipped in). Net: a clean, well-scoped rebrand. The technical identifiers really are untouched - I checked that `stenoai://`, the bundle id, data paths, the CLI binary name, and the DMG/exe artifact filenames don't appear in the diff; 28 `StenoAI` strings removed, 0 re-added. (Two observable strings *do* change but aren't identifiers: the update-check `User-Agent: StenoAI-Updater -> Steno-Updater`, and the CLI banner `Steno Setup Check` - both harmless.) The notes below are one required rebase plus a couple of minor calls, nothing blocking on the code itself.

## Overlap with already-merged work (coordination heads-up)

- **#315 (merged) - direct `.gitignore` conflict, and this is what's making the PR `DIRTY`.** #315 already deleted the bare `docs/` ignore line, because `docs/` is a fully tracked Mintlify site now and the blanket ignore was silently swallowing new pages. This PR rewrites that *same* line to `app/renderer/docs/` - but that's not a safe anchor: **`app/renderer/docs/` already exists on `main` and holds a tracked file, `app/renderer/docs/component-audit.md` (added in #95).** So the anchored ignore doesn't protect a scratch path - it would silently ignore *new* files added under an already-tracked subtree, i.e. re-introduce the exact class of bug #315 fixed. The stale `docs/` rule and that tracked file both came from #95, so the rule was overbroad, not a currently-valid scratch pattern. On rebase, **drop the `.gitignore` hunk entirely** and take `main`'s version.
- **#316 (merged) - same rebrand, disjoint files.** #316 did the `StenoAI -> Steno` copy in the release workflow (`build-release.yml`); this PR does the app/website/docs copy. No file overlap - they're the two halves of the same #227 rollout, so together they finish it. Flagging only so the naming change reads as coordinated rather than duplicated.
- **#320 (merged) - `main.js`.** Your `main.js` edits are all display-string naming; #320 added the notification-window code elsewhere in the same file. No semantic overlap, should rebase cleanly, but the file shifted ~60 lines so it's worth re-verifying the naming hunks land correctly after the rebase.

## Findings

1. **[needs action] Rebase required** - `mergeStateStatus: DIRTY` against current `main`. The only real conflict is the `.gitignore` line above; resolution is to drop your hunk (do *not* keep the `app/renderer/docs/` anchor - see above).
2. **[minor - scope call] Naming completeness.** Two surfaces still say `StenoAI` and aren't in this diff: `MECHANICS.md` (title + one line) and `skills/granola-to-steno/{README,SKILL}.md` - **33 occurrences total** across those three files. All are dev/contributor docs rather than app UI, so leaving them may be a deliberate scope boundary - just want it to be a conscious call rather than a miss, given the title says "across app/website/docs".
3. **[minor] The canonical mark is now inlined in two places** (`app-icon.tsx` *and* `Sidebar.tsx`, plus `AppWindowDemo.jsx` on the site) instead of all importing `AppIcon`. You already flag this under "follow-up look" - no objection to deferring, just noting the copies are byte-identical *today*, and de-duping the mark is the exact thing this PR is about, so it's worth not losing the follow-up.
4. **[minor] The Ovo font is inlined as a ~40 KB base64 data-URI in `globals.css`** (parsed on every renderer load, not separately cacheable), while the website ships `ovo-regular.woff2` as a real file. The data-URI is correct given the `font-src 'self' data:` CSP, but a real `.woff2` + `font-src 'self'` would be lighter and consistent with the site. Non-blocking.

Everything else reads as intentional and correct: `font-weight: 500 -> 400` for the serif headings (Ovo has no bold, so this avoids fake-bold synthesis), the template-image tray icon (one asset, OS re-tints for light/dark), dropping the Charter fallback chain, and dropping `strokeLinecap`/`strokeLinejoin` (the "sharper caps" you already called out).

Nice cleanup overall. Rebase + the `.gitignore` drop and it's good from my side; item 2 is the only open question.

<sub>_(Edited to tighten the `.gitignore` point - `app/renderer/docs/` already holds a tracked file - and correct the doc-occurrence count.)_</sub>

ruzin (approved):

PR #328, opened 2026-07-09
feat: launch Steno on login by default (hidden), with opt-out toggle

WilliamDrewett:
## Description

Steno never started on login before — users had to relaunch it after every reboot. This adds a `launch_on_login` setting (default ON for new **and** existing users, opt-out in Settings) that registers an OS login item and starts Steno hidden in the tray/menu bar. It's wired through config, CLI, IPC, preload, and a General-tab toggle; on a login launch the first window show is suppressed (macOS `wasOpenedAtLogin`, Windows `--hidden`) while keeping `windowReadyToShow` so tray/Dock activation still reveals the window. To keep analytics honest, `app_opened` is now tagged with `launch_source`/`hidden` so background auto-launches don't inflate DAU/funnels (the five affected PostHog insights were updated to filter them out), and the real `launch_on_login` value is reported in telemetry.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested on macOS

Python config tests (default-on, legacy fallback, round-trip), GeneralTab unit mocks, and a `launchOnLogin` case in the `settings-roundtrip` T2 e2e spec all pass; the CLI round-trip was verified against the bundled binary. Login-item registration is gated off under E2E/dev so tests never touch the developer's real login items.

## Additional Notes

The OS login-item registration and hidden-launch behaviour still need a manual check on an installed signed build (install DMG → confirm it appears in macOS Login Items, toggle off removes it, reboot starts Steno hidden) — that can't be exercised in dev/CI.

---
## Summary by cubic
Enable Steno to auto-launch on login by default, starting hidden in the tray/menu bar, with an opt-out toggle in Settings. Also tags launches so background starts don’t skew metrics.

- New Features
  - Added `launch_on_login` setting (default ON for new and existing users) that registers/removes the OS login item and starts hidden on login.
  - Settings > General toggle wired through config, CLI (`get-launch-on-login`/`set-launch-on-login`), IPC, preload, and hooks.
  - Applies the login item on startup and on toggle; no-op in E2E/dev; errors are non-fatal.
  - Platform behavior: macOS detects `wasOpenedAtLogin`; Windows uses `--hidden`. Window show is suppressed only on login launches.
  - Tests cover default-on and round-trip in config and e2e.

- Telemetry
  - `app_opened` now includes `launch_source` and `hidden`, and identity reports the real `launch_on_login` value.

<sup>Written for commit f2922bffe9e651fd93b21ae5ee46adb5843ed929. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Optic00 (changes requested):
The implementation is solid - the wiring through config/CLI/IPC/preload/UI is complete, the stale-login-item case is handled, the E2E/dev gating keeps developer machines clean, and the test coverage (Python default/legacy/round-trip plus the `settings-roundtrip.t2` case that deliberately flips the default) is exactly what we want here. The `launch_source`/`hidden` tagging on `app_opened` is also the right call so background launches don't inflate DAU.

My main concern is the default, and it's the reason for the "request changes":

**1. Autostart should be opt-in, not default ON (my clear opinion, final call is @ruzin's)**

As written, every *existing* install silently registers itself as a login item after the next update - `cfg.launch_on_login !== false` plus the re-apply on every startup is built exactly for that. macOS at least shows the "added as login item" notification, Windows shows nothing at all. I think an app quietly adding itself to autostart is the kind of behaviour users hold against you, especially for a tool whose pitch is being respectful and local-first. It's also the same discussion we had around auto-update, where we deliberately went notification-only instead of silent.

Suggestion: keep all the plumbing as is, but default `launch_on_login` to `false` (new and legacy configs). If we want adoption, a one-time hint in onboarding or a settings callout is the honest way to get it.

**2. Merge precondition: the manual signed-build test you already mention**

`getLoginItemSettings().wasOpenedAtLogin` on macOS 13+ goes through SMAppService and its reliability in Electron has been shaky. The failure mode is benign (window shows at login instead of starting hidden), but since dev/CI can't exercise any of this, the DMG test (install, confirm Login Items entry, toggle off removes it, reboot starts hidden) should happen before merge, not after. Happy to help test on macOS.

**3. Re-apply on every startup can fight an OS-level opt-out**

If a user disables the login item in macOS System Settings or Windows Task Manager instead of the in-app toggle, the startup re-apply re-registers it on every launch. Recent macOS/Windows versions usually preserve the user's disable state, but we haven't verified that. With an opt-in default this mostly goes away (the re-apply then only ever acts on an explicit user choice), which is one more argument for point 1. Alternatively, only call `applyLoginItemSetting` on toggle change instead of every startup.

Small nit, non-blocking: the `whenReady` block now reads and parses `config.json` twice (login-launch detection and the re-apply). Could be a single read.

ruzin (approved):

PR #329, opened 2026-07-09
fix(security): enforce canonical path containment on renderer-supplied meeting paths

Optic00:
## Summary

Closes several path-traversal issues where the untrusted Electron renderer (XSS / compromised renderer / DevTools) could drive the backend or the main process into reading, writing, or deleting arbitrary local files. Enforcement is at the Electron trust boundary; the Python CLI stays intentionally permissive so the documented `query transcript.txt` debug path keeps working.

## Issues fixed

- **Arbitrary file read (+ potential exfiltration)** — `query-transcript` / `query-transcript-stream` passed the renderer's `summaryFile` straight to the backend `query` command, which reads any UTF-8 file (no extension gate on the plain-text path). With a cloud/org provider the contents are also sent to the external AI service.
- **Arbitrary file write** — `add-/remove-meeting-to-folder` passed `summaryFile` straight to the folders backend, which rewrites the target file.
- **Symlink escape** — `update-meeting` used a lexical-only path check, then read/wrote the original (symlink-followed) path across the whole project/storage root.
- **Symlink-followed delete** — `delete-meeting` / `reveal` / system-audio shared the same lexical check.

## Fix

- Route the four meeting-file handlers through `validateMeetingFilePath()` (output/-only, realpath-canonicalized) and pass the canonical `realPath` downstream — never the original renderer string.
- Harden `validateSafeFilePath()` with `realpathSync` canonicalization of both target and base dirs, covering delete/reveal/system-audio with no call-site change.
- Additional hardening on the stream handler this change makes async: guard a `query-cancel` or sender-destroy landing during pre-spawn validation (cancel-race / orphaned backend), type-guard non-string input at the `validateMeetingFilePath` source (avoids an unhandled rejection that could crash main), and make `query-done` delivery exception-safe. The `chat_message_sent` analytics from #323 is preserved.

## Tests

New model-free T2 e2e spec `path-validation-security.t2.spec.ts`: asserts rejection of out-of-`output/` paths, non-string input, and symlink escapes on both validator seams, and that the out-of-bounds file is never read / modified / deleted. `meetings-crud` / `folders-crud` / `chat-sessions` T2 regressions green.

---
## Summary by cubic
Enforces symlink-safe, canonical path containment for renderer-supplied meeting files to block arbitrary local file read/write/delete from the Electron renderer. Validation now happens at the Electron trust boundary; the Python CLI remains permissive for documented debug flows.

- **Bug Fixes**
  - Guard `query-transcript`, `query-transcript-stream`, `update-meeting`, and `add/remove-meeting-to-folder` with `validateMeetingFilePath` (output/ only) and pass the canonical `realPath` downstream.
  - Canonicalize both target and base dirs in `validateSafeFilePath` using `fs.realpathSync`, securing delete/reveal/system-audio without call-site changes.
  - Pre-validate in `query-transcript-stream`, add pending-cancel handling to avoid orphaned backends, and make `query-done` delivery exception-safe.
  - Reject non-string paths early to prevent unhandled rejections and main-process crashes.
  - Add a T2 e2e spec covering out-of-`output/` paths, non-string input, and symlink escapes; asserts no unauthorized disk reads/writes/deletes.

<sup>Written for commit 541654fa0d290d85c90bafe4009577e31a754a8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -866,10 +866,39 @@ function validateSafeFilePath(filepath, allowedBaseDirs) {
     // Resolve to absolute path and normalize
     const resolvedPath = path.resolve(filepath);
 
+    // The renderer is untrusted, so a lexical prefix check alone is
+    // symlink-vulnerable: a symlink placed inside an allowed dir but pointing
+    // outside it would slip through. Canonicalize BOTH the target and each base
+    // dir with realpath before the containment check, so an escaping symlink
+    // resolves to its real (out-of-bounds) target and is rejected. realpath also
+    // normalizes platform quirks that must match on both sides — on macOS
+    // /tmp -> /private/tmp, which the e2e temp data dir relies on — which is
+    // exactly why we canonicalize both sides.
+    //
+    // realpath throws if the path doesn't exist yet (e.g. a file being created),
+    // so fall back to canonicalizing the parent directory and re-appending the
+    // basename; if even the parent can't be resolved, use the lexical path so we
+    // don't break create-new-file flows.
+    let canonicalPath;
+    try {
+      canonicalPath = fs.realpathSync(resolvedPath);
+    } catch (_) {
+      try {
+        canonicalPath = path.join(fs.realpathSync(path.dirname(resolvedPath)), path.basename(resolvedPath));
+      } catch (_) {
+        canonicalPath = resolvedPath;
+      }
+    }
+
     // Ensure it's within one of the allowed base directories
     for (const baseDir of allowedBaseDirs) {
-      const resolvedBase = path.resolve(baseDir);
-      if (resolvedPath.startsWith(resolvedBase + path.sep) || resolvedPath === resolvedBase) {
+      let resolvedBase;
+      try {
+        resolvedBase = fs.realpathSync(path.resolve(baseDir));
+      } catch (_) {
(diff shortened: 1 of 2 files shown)


ruzin (approved):
Reviewed the full diff as a security change (path-traversal / symlink-escape focus). The fix is sound and defense-in-depth: validateMeetingFilePath requires existence and substitutes the canonical realPath into the actual read/write, and validateSafeFilePath canonicalizes both target and base with realpath so escaping symlinks resolve out-of-bounds and fail containment. Checked the hypothesized dangling-symlink write bypass — not reachable: reveal is a harmless Finder call, system-audio requires existsSync, and delete uses unlink (removes the link, not the target). e2e symlink-rejection spec covers the seams. All required checks green.

PR #332, opened 2026-07-11
fix(main): sweep orphaned stenoai-live-*.txt snapshots at startup (#259)

### LINKED ISSUE #259: Live-transcript snapshot temp files (stenoai-live-*.txt) leak on quit/crash while job is queued
### Background

Follow-up from #247 (live-transcript rescue, merged 2026-06-28). Noted as a non-blocker in the review thread.

The live-transcript fallback writes a snapshot of the in-progress transcript to a temp file in `os.tmpdir()` (`stenoai-live-*.txt`) so the Python rescue path can pick it up via `--live-transcript`. That temp file is only unlinked in `processNextInQueue`'s `finally` block.

### Problem

If the app quits or crashes **while the job is still queued** (before `processNextInQueue` reaches its `finally`), the snapshot is never removed. Over time these orphaned `stenoai-live-*.txt` files accumulate in the temp dir.

Impact is minor — temp-dir clutter, no data loss or incorrect behaviour — which is why it didn't block the #247 merge.

### Proposed fix

Add a startup sweep that deletes leftover `stenoai-live-*.txt` snapshots from `os.tmpdir()` on app launch (after any in-flight rescue from the previous session has had its chance to run, or scoped so it only removes snapshots not tied to a queued job). Keeps the normal `finally`-based cleanup as the primary path; the sweep just closes the crash/quit window.

### Notes
- Cross-platform: the sweep must use `os.tmpdir()` (honoring `APPDATA`/temp on Windows), not a hardcoded `/tmp`.
- Low priority / good first issue.

Optic00:
Closes #259.

## Problem

`snapshotLiveTranscriptForFallback` writes `stenoai-live-<ts>-<rand>.txt` into `os.tmpdir()` and the file is only unlinked in `processNextInQueue`'s finally block. If the app quits or crashes while the job is still queued, that finally never runs and the snapshot is orphaned forever, accumulating in the temp dir.

## Fix

New `app/live-snapshot-sweep.js` with `sweepOrphanedLiveSnapshots({ tmpDir, keepPaths, minAgeMs, now, fs })`, called once at startup right after `checkForOrphanedRecording()`. The existing finally-unlink stays the primary cleanup; the sweep only closes the quit/crash window.

Lifecycle finding that drives the keep-set (documented in code): the processing queue (`processingQueue` / `currentProcessingJob`) is pure in-memory state and the `.recording-active` marker stores only a timestamp, so after a restart no job can reference a snapshot. The keep-set is still built defensively from live queue state so the wiring stays correct if requeue-on-restart is ever added. The 60s age guard is the real safety net against racing a snapshot that a concurrent startup path just created.

Safety properties:

- Filename match is pinned to the writer's exact shape (`/^stenoai-live-\d{10,17}-[a-z0-9]{1,8}\.txt$/`), so it cannot delete files this app did not create.
- `lstatSync` + `isFile()`: never follows a symlink out of tmpDir, never touches directories, never recurses.
- Every fs call is wrapped; a single ENOENT/EPERM never throws out of the sweep, and the whole call is additionally wrapped so it can never block launch.
- Keep-set paths are compared case-insensitively on Windows (case-insensitive filesystem).
- Cross-platform: `os.tmpdir()` only, no `/tmp` hardcoding, no shelling out.

## Tests

14 new `node:test` cases in `app/live-snapshot-sweep.test.js` (wired into `test:unit`): deletes old orphans, keeps keep-set entries, age guard via injected clock, pattern negatives (single-digit timestamp, overlong suffix, non-matching names, directories), vanished-file race via injected fs, case-sensitivity both modes, accurate deleted/kept reporting.

`npm run test:unit`: node --test 123/123, vitest 84/84.

No e2e spec: the change has no UI surface; the sweep is a pure module covered by unit tests and the wiring is a guarded startup call.

---
## Summary by cubic
Fixes #259 by adding a startup sweep that removes orphaned live-transcript snapshots (`stenoai-live-*.txt`) in `os.tmpdir()` after crashes. Keeps active or new files and works safely across platforms.

- **Bug Fixes**
  - Added `sweepOrphanedLiveSnapshots` and call it at startup after `checkForOrphanedRecording()`.
  - Strict filename match (`^stenoai-live-\d{10,17}-[a-z0-9]{0,8}\.txt$`), `lstatSync` to avoid symlinks/dirs, best-effort deletes.
  - Keeps files from the current queue and anything younger than 60s; keep-set compare is case-insensitive on Windows.
  - Tests: new `app/live-snapshot-sweep.test.js` (injected clock, empty-suffix edge, tmp cleanup); updated `test:unit` script.

<sup>Written for commit 68db0c09263d008205c088d0fc7612488dcbbb52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ruzin (dismissed):
Reviewed the diff. Exemplary defensive cleanup: regex pinned to the writer's exact filename shape (incl. the degenerate empty-suffix case), lstat+isFile so it never follows a symlink or touches a directory, 60s age guard against racing a fresh snapshot, injectable clock/fs for deterministic tests, and the whole call wrapped so it can never block launch. Even a wrong keep-set field would fail benign (age guard). 14 unit tests, all checks green.

ruzin (approved):
Re-approving after resolving the merge conflict against main (trivial additive: kept both notes-file.test.js + live-snapshot-sweep.test.js in the test:unit script, and both requires + both startup blocks in main.js). Ran the node --test suite locally: 131/131 pass. Original review stands — clean, well-tested defensive sweep.

ruzin (approved):

ruzin (approved):

PR #333, opened 2026-07-11
fix(language): detect transcript language so auto output-language stops defaulting to English (#283)

### LINKED ISSUE #283: Parakeet never detects meeting language, so auto output-language falls back to English
Split out from #281. The default-language fix (#282) sets \`DEFAULT_CONFIG["language"]\` to \`"auto"\`, which works for Whisper (whisper.cpp sets a real \`detected_language\`) - but not for Parakeet, the default transcription engine (\`parakeet-mlx\` on macOS Apple Silicon, \`onnx-asr\` on Windows/Linux).

Traced in both backends:

\`\`\`python
# src/_parakeet_mlx.py:235 and src/_parakeet_onnx.py:284
detected_language = language if (language and language != "auto") else None
\`\`\`

Parakeet TDT v3 is multilingual and language-agnostic at inference time (it just decodes whatever's spoken), but this codebase never asks it - or anything else - what language it actually decoded. \`detected_language\` is purely an echo of the caller-provided language; when \`language="auto"\` (→ passed as \`None\` per \`transcriber.py:682\`), \`detected_language\` comes back \`None\` too. \`_resolve_output_language("auto", None)\` (\`simple_recorder.py\`) then falls through to \`"en"\`.

Net effect: for anyone on the default engine who hasn't manually pinned a language in Settings, meeting summaries are generated in English regardless of the actual spoken language - e.g. a German meeting gets an English summary.

## Possible fix directions (not scoped/decided yet)

- Run lightweight text-based language detection on the transcript *after* transcription completes (engine-independent - works the same for Parakeet and Whisper), then feed that into \`_resolve_output_language\` as the \`detected_language\` hint. Needs a language-ID library/approach that's cheap enough to run on every meeting (a small transcript sample should suffice) and accurate for the languages this app's Settings dropdown already supports (\`app/renderer/src/lib/transcription-languages.ts\` - the curated European set).
- Alternative: have the summarization LLM itself detect/declare the language as part of its existing prompt (it already receives the full transcript), avoiding a new dependency - but this couples output-language detection to the summarization step rather than making it available earlier (e.g. for the title-generation step too).

Needs a design decision before implementation - bigger scope than the #282 default-value fix.

Optic00:
Closes #283.

## Problem

Parakeet's backends never report a detected language, they only echo an explicit pin. With the default `language="auto"`, `detected_language` is `None` and `_resolve_output_language("auto", None)` falls through to `"en"`, so anyone on the default engine who hasn't pinned a language gets English summaries, titles, reports and chat answers for a German/French/... meeting. Whisper is unaffected (whisper.cpp reports a real detected language).

## Fix

**New `src/language_detect.py`**: a dependency-free stopword/function-word classifier over exactly the Parakeet-supported languages (`en, fr, de, es, nl, pt`, the `PARAKEET_LANGUAGES` set minus `auto`). It strips diarisation markers and timestamps, samples the first 8000 chars, and only commits to a language when it clears a hit threshold (15) and leads the runner-up by 1.3x; otherwise it returns `None` and the existing `"en"` fallback applies. No new pip dependency (PyInstaller bundle size and licensing).

**New priority tier** in the output-language resolution, now module-level and unit-testable (`resolve_output_language`, the `MeetingPipeline` method delegates to it):

1. explicit user pin
2. engine-detected (Whisper, unchanged)
3. NEW: text detection over the transcript
4. `"en"` fallback

Wired at every call site that has transcript text: initial pipelines (streaming + non-streaming, incl. auto-summarize-off), reprocess, `generate_report`, `regen-title`, transcript write, and the `query`/`query_streaming` chat paths. Intentionally unchanged: `_handle_transcription_failure` (no transcript to detect from) and `chat_global_streaming` (multi-meeting corpus, single-language detection would be unreliable).

**Provenance rule for persisted values** (`resolve_persisted_output_language`): notes processed before this fix persisted `output_language: "en"` from the buggy fallback, which would otherwise act as a pin and keep the recovery paths English forever. A persisted `output_language` is now honoured only when it is pin-backed (`configured_language != "auto"`) or engine-backed (`detected_language` present); otherwise reprocess/report/regen-title re-detect from the transcript. A real user pin always wins, Whisper detections are respected, and re-detecting a previously text-detected note is idempotent. `report_store.read_meeting` additionally surfaces `configured_language`/`detected_language` to carry that provenance.

Privacy: the one new log line carries only the detected language code, never transcript content.

## Tests

20 unit tests in `tests/test_language_detect.py`: per-language meeting-style samples (with diarisation markers/timestamps), inconclusive/short/empty/symbol inputs return `None`, mixed-language dominance, full priority-order matrix, and the persisted-provenance matrix (stale fallback re-detects; pins and engine detections are never second-guessed).

`ruff check` clean on all touched files (`simple_recorder.py` keeps its 16 pre-existing findings, one old one removed). Full-suite `unittest discover` compared against an origin/main baseline in the same environment: identical pre-existing failures, delta is exactly the 20 new passing tests, zero regression.

No e2e spec in this PR: asserting the output language end-to-end requires a real summarization model, which belongs to the `@pipeline` lane; the resolution logic itself is deliberately module-level and fully unit-covered.

---
## Summary by cubic
Detects meeting language from transcript and uses it in auto mode, so Parakeet recordings no longer default to English. Adds provenance-aware recovery and chat so old notes with a stale "en" recover to the correct language.

- **New Features**
  - Added `src/language_detect.py`: a dependency‑free detector for `en, fr, de, es, nl, pt` that scans early transcript text and ignores diarisation/timestamps.
  - Introduced module‑level `resolve_output_language`: priority is user pin > engine (Whisper) > transcript > "en"; all call sites now pass transcript text where available.
  - Persisted language provenance: markdown front matter now writes `configured_language`/`detected_language` on create and reprocess; `report_store.read_meeting` returns them (coercing `null`/`""` to `None`); `skills/granola-to-steno` sets `configured_language` and `detected_language: null` on import.

- **Bug Fixes**
  - Fixes #283: Parakeet auto meetings no longer fall back to English; recovery paths (reprocess, report, title) and chat resolve via `resolve_persisted_output_language`, trusting a persisted language only when pin‑ or engine‑backed.
  - Query and streaming query detect language over the raw transcript (not combined summary/context) to avoid legacy English summaries flipping detection.

<sup>Written for commit 7ab280782f63d021e2d86becdb141ecee958405e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/simple_recorder.py
+++ b/simple_recorder.py
@@ -516,6 +516,8 @@ Summary output language: {config.get_language_name(output_language)}
             'date': processed_at,
             'duration_seconds': int(duration_seconds) if duration_seconds else None,
             'language': transcript_data.get("configured_language"),
+            'configured_language': transcript_data.get("configured_language"),
+            'detected_language': transcript_data.get("detected_language"),
             'is_diarised': False,
             'transcription_failed': True,
             'reprocessable': True,
@@ -619,6 +621,8 @@ Summary output language: {config.get_language_name(output_language)}
                 'date': processed_at,
                 'duration_seconds': int(duration_seconds) if duration_seconds else None,
                 'language': output_language,
+                'configured_language': gate_config.get_language(),
+                'detected_language': transcript_data.get('detected_language'),
                 'is_diarised': transcript_data.get('is_diarised', False),
                 'notes_generated': False,
             }
@@ -709,6 +713,8 @@ Summary output language: {config.get_language_name(output_language)}
             'date': processed_at,
             'duration_seconds': int(duration_seconds) if duration_seconds else None,
             'language': output_language,
+            'configured_language': configured_language,
+            'detected_language': transcript_data.get('detected_language'),
             'is_diarised': transcript_data.get('is_diarised', False),
         }
         md_lines = _render_frontmatter(md_meta)
@@ -975,6 +981,8 @@ def process_streaming(audio_file, name, notes, live_transcript):
                 'date': processed_at,
                 'duration_seconds': int(duration_seconds) if duration_seconds else None,
                 'language': output_language,
+                'configured_language': gate_config.get_language(),
+                'detected_language': transcript_data.get('detected_language'),
                 'is_diarised': transcript_data.get('is_diarised', False),
                 'notes_generated': False,
             }
@@ -1079,6 +1087,8 @@ def process_streaming(audio_file, name, notes, live_transcript):
             'date': processed_at,
             'duration_seconds': int(duration_seconds) if duration_seconds else None,
             'language': output_language,
+            'configured_language': configured_language,
+            'detected_language': transcript_data.get('detected_language'),
             'is_diarised': transcript_data.get('is_diarised', False),
         }
         # Mark live-sourced meetings (#207) so the UI and future code know this
@@ -2308,6 +2318,13 @@ def _parse_meeting_markdown(md_path):
         'duration_seconds': meta.get('duration_seconds'),
         'summary_file': str(md_path),
         'output_language': meta.get('language'),
+        # Provenance of the output language, so recovery paths (reprocess /
+        # generate-report / regen-title / chat) can tell a real user pin or
+        # Whisper engine detection from a bare Parakeet fallback (#283). Old .md
+        # files predating these keys read as None -> treated as no provenance
+        # (re-detect), preserving prior behaviour.
+        'configured_language': meta.get('configured_language'),
+        'detected_language': meta.get('detected_language'),
     }
     # A meeting whose transcription crashed carries these markers so the UI
     # can render an honest failure state (and a future retry) rather than a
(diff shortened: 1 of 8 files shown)


ruzin (approved):
Reviewed the full diff. Well-reasoned and low-risk by construction: detect_transcript_language only overrides the old 'en' fallback when confident (>=15 stopword hits AND a 1.3x lead over the runner-up), else returns None and prior behavior applies — so worst case is the status quo, never a regression. The provenance model (resolve_persisted_output_language trusts a persisted language only when pin- or engine-backed, else re-detects; stored provenance authenticates, not the caller's current pin) is subtle but correct, and the report_store null-coercion prevents a serialized 'null' from masquerading as engine-backed. 20 unit tests cover the full priority matrix, stale-en recovery, diarisation/timestamp stripping, and the raw-transcript query seam. All checks green. No e2e is justified (output-language assertion needs the @pipeline model lane; the resolution logic is module-level unit-covered).

PR #335, opened 2026-07-12
fix: make launch on login opt-in

Optic00:
## Summary

Follow-up to #328: launch on login becomes strict opt-in. The feature and its implementation stay as they are; only the consent model changes. Steno registers itself as an OS login item only after the user explicitly enables the setting.

- New installs default to `launch_on_login: false` (`src/config.py`).
- Legacy configs without the key resolve to off instead of on, in the Python getter and in both direct `config.json` reads in `main.js` (hidden-launch detection and startup reconciliation), which now require an explicit `launch_on_login === true`.
- Telemetry identity and the Settings toggle fallback report/render off unless explicitly enabled.
- An explicit persisted `true` keeps working exactly as before: the login item is registered and survives restarts.

## Rationale

#328 enabled the feature for everyone, including existing installations, which would silently register Steno as a login item without prior consent. That sits badly with Steno's local-first, user-respecting posture. Since #328 has not shipped in a published release yet, the default can still be corrected cleanly: no migration, no cleanup marker, nobody to un-register.

## Testing

- `python -m pytest tests/test_config.py -q`: 54 passed + 2 subtests, including the rewritten `ConfigLaunchOnLoginTests` (new config off, legacy config without key off, explicit true/false round-trip).
- `cd app && npx vitest run renderer/src/routes/settings/GeneralTab.test.tsx`: 8 passed.
- Focused T2 (`settings setters persist each value to the right config.json key`) passed locally against a freshly built backend bundle. The E2E case now writes `true` (the opposite of the new default) so a no-op setter cannot pass.
- Renderer typecheck clean.

## Out of scope / follow-ups

- No onboarding prompt; a future onboarding step could offer launch on login as an explicit choice.
- Startup reconciliation is intentionally unchanged, including its error handling: if `config.json` exists but cannot be parsed, the reconciliation is skipped and a previously registered login item stays registered (pre-existing behavior from #328). Auto-unregistering on a read error could just as well drop a genuine opt-in, so this is left for a separate decision if it matters in practice.
- OS-level disabling via macOS System Settings or Windows Task Manager is not handled more precisely here; that needs platform-specific testing first.

---
## Summary by cubic
Make Launch on Login opt-in. New and legacy installs default to off; Steno only registers as an OS login item after the user explicitly enables it. Explicit true persists and keeps the login item.

- **Bug Fixes**
  - Default `launch_on_login` to false in config; missing key/file resolves to off.
  - Electron main requires `launch_on_login === true` for telemetry, hidden-launch detection, and startup reconciliation; otherwise remains off.
  - Settings toggle fallback renders off; unit tests and E2E round-trip updated to reflect opt-in behavior.

<sup>Written for commit ed0e89a2e218dc2c6ace946f2793f63b5caec440. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -518,9 +518,8 @@ function applyLoginItemSetting(enabled) {
  * Initialize PostHog telemetry by reading config from Python backend.
  */
 // One config.json read (no extra subprocess) + a token-file existence check
-// for the identify() super-properties below. `launch_on_login` defaults ON
-// (feature ships enabled-for-everyone; users opt out in Settings), so a legacy
-// config missing the key reports true.
+// for the identify() super-properties below. `launch_on_login` is opt-in, so a
+// legacy config missing the key reports false.
 function loadIdentitySuperProperties() {
   let cfg = {};
   try {
@@ -535,7 +534,7 @@ function loadIdentitySuperProperties() {
     ai_provider: cfg.ai_provider || 'local',
     notifications_enabled: cfg.notifications_enabled !== false,
     calendar_connected: calendarConnected,
-    launch_on_login: cfg.launch_on_login !== false,
+    launch_on_login: cfg.launch_on_login === true,
   };
 }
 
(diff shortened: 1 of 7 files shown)


WilliamDrewett:
One thing worth noting: the design doc assumed #328 hadn't shipped in a release yet, which was true at the time you wrote it, but v0.6.0 and v0.6.1 have since gone out and both include the default-on behaviour. 

Given that, and where we are as a product right now - need higher engagement and retention - we've decided to keep auto-launch on by default rather than move it behind an opt-in (discussed with @ruzin). 

The main reasoning: it's what most people would expect from a meeting-recording app (you'd want it running so it doesn't miss something), and it's not really hidden either way - it shows up as a normal login item in System Settings / Task Manager, so someone who doesn't want it can see and remove it there or in our own Settings toggle, which still takes effect immediately.

To be clear, I don't want to dismiss the consent concerns - it's totally valid, we're just weighing it against where the product needs to be right now and landing differently. 

Going to close this one out for now.

PR #336, opened 2026-07-12
fix: make telemetry strict opt-in

Optic00:
## Summary

Telemetry becomes strict opt-in. The PostHog integration, event payloads, and sanitization layer are unchanged; only the consent model changes.

- New installs default to `telemetry_enabled: false`.
- Strict consent semantics: only JSON boolean `true` enables telemetry (`is True` in Python); malformed persisted values can never turn it on.
- One-time consent reset for the installed base: v0.5.8 shipped default-on, and the config layer auto-persists the full default dict on first write, so existing installs carry `telemetry_enabled: true` that was never an affirmative choice. A marker-guarded migration (`telemetry_opt_in_migrated`) resets those configs to `false` once — written through a locked compare-and-set so a concurrent affirmative opt-in is never clobbered, with the snapshot kept in sync on every persist outcome so a later ordinary save cannot bypass that protection. The reset also writes an explicit `false` when the key was merely missing, so a rollback to v0.5.8 (missing key = on) stays off.
- The onboarding and Settings toggles are unchanged and remain the only paths that persist `true`, so post-migration consent is affirmative by definition.
- `disableGeoip: true` is now explicit on both PostHog constructors (already the SDK default; drift protection).

## Rationale

Anonymous, content-free analytics are legitimately useful, but default-on conflicts with Steno's local-first, privacy-respecting posture — the same reasoning as the launch-on-login opt-in correction (#335). Because default-on already shipped in v0.5.8, a default flip alone would not deliver opt-in for existing users; the one-time reset is the price of the previous `true` never having been affirmative.

Known consequence: analytics volume will collapse after this ships, and users who genuinely opted in on v0.5.8 are reset once (locally indistinguishable) and must re-enable. Worth marking the release boundary in the PostHog dashboards so pre/post metrics are not compared as if collection were stable.

## Testing

- `python -m pytest tests/test_config.py -q` — 63 passed + 8 subtests, including the new `ConfigTelemetryOptInTests` (fresh install off, legacy true reset on disk, missing key persisted false, malformed values never enable, marker prevents re-migration, round trip, corrupt config untouched, CAS race adoption, failed-persist no-clobber regression).
- `cd app && npm run typecheck:renderer && npm run lint:renderer` — clean.
- Focused T2 settings round-trip against a freshly built backend bundle — passed; the telemetry case now writes `true` (the opposite of the new default) so a no-op setter cannot pass.

## Release checklist (blocking)

- [ ] Website/docs copy must ship with the release that contains this change: `website/public/privacy.html` ("enabled by default"), `docs/privacy/how-on-device-works.mdx` ("no analytics"), and FAQ claims need release-coordinated corrections in a separate PR — the website deploys on merge while the latest published release still defaults on, so bundling the copy here would create a disclosure/version skew.

## Follow-ups

- A future onboarding step could present the telemetry choice more prominently; intentionally out of scope here.
- Pre-existing, unrelated residual: `_save()`'s lock-timeout fallback does an unlocked wholesale write for every config key by design ("a stuck lock must never block saves"); under two consecutive lock timeouts in one process it could overwrite concurrent writes. Not changed here.

---
## Summary by cubic
Make telemetry strict opt-in. New installs default to off, existing configs are reset once to off, telemetry only runs when `telemetry_enabled` is exactly JSON true, the Setup toggle defaults to off, and the PostHog client now sets `disableGeoip: true`.

- **Migration**
  - Resets `telemetry_enabled` to `false` once for configs missing `telemetry_opt_in_migrated`, and persists the marker; also writes `false` when the key was missing (rollback-safe).
  - Uses a locked compare-and-set; if a user opts in during migration, the on-disk opt-in is kept. Keeps the in-memory snapshot in sync even if the persist fails so later saves can’t clobber an affirmative opt-in.
  - No action needed; users who want telemetry must re-enable it in Settings or during Setup.

<sup>Written for commit b7185dfe8eb74b2f19d10867942ff05c69036b89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -580,7 +580,10 @@ async function initTelemetry() {
     anonymousId = config.anonymous_id;
 
     if (telemetryEnabled) {
-      posthogClient = new PostHog(POSTHOG_API_KEY, { host: POSTHOG_HOST });
+      // disableGeoip is already the posthog-node default; explicit here as
+      // drift protection so an SDK upgrade can't silently re-enable IP
+      // geolocation for opted-in users.
+      posthogClient = new PostHog(POSTHOG_API_KEY, { host: POSTHOG_HOST, disableGeoip: true });
       // Identify user for DAU tracking
       posthogClient.identify({
         distinctId: anonymousId,
@@ -6398,7 +6401,7 @@ ipcMain.handle('set-telemetry', async (event, enabled, source) => {
     if (enabled && !posthogClient) {
       // Bring the client up and identify FIRST so both the super-properties
       // and this event land fresh, rather than waiting for next launch.
-      posthogClient = new PostHog(POSTHOG_API_KEY, { host: POSTHOG_HOST });
+      posthogClient = new PostHog(POSTHOG_API_KEY, { host: POSTHOG_HOST, disableGeoip: true });
       telemetryEnabled = true;
       refreshIdentitySuperProperties();
       trackEvent('telemetry_toggled', { enabled: true, source: toggleSource });
--- a/app/renderer/src/routes/Setup.tsx
+++ b/app/renderer/src/routes/Setup.tsx
@@ -130,7 +130,7 @@ export function Setup() {
   // immediately via the same backend used by the Settings page.
   const telemetry = useTelemetrySetting();
   const setTelemetry = useSetTelemetry();
-  const telemetryEnabled = telemetry.data?.telemetry_enabled ?? true;
+  const telemetryEnabled = telemetry.data?.telemetry_enabled ?? false;
 
   // First name powers the in-app greeting ("Hi <name>, ask anything"). Stored
   // locally only — never sent anywhere. Persisted on blur to avoid a write
--- a/docs/superpowers/plans/2026-07-12-telemetry-opt-in.md
+++ b/docs/superpowers/plans/2026-07-12-telemetry-opt-in.md
@@ -1,0 +1,481 @@
+# Telemetry Opt-In Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Telemetry runs only for users who affirmatively enabled it — strict opt-in with a one-time consent reset for the installed base.
+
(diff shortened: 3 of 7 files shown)


ruzin:
@Optic00 these need to be toggles in settings as discussed + onboarding toggle (if you have time)

WilliamDrewett:
Thanks for the thorough work here - the migration design and the consent-marker approach are solid, and the wholesale-default-merge finding is a gap worth tracking regardless of what we decide below.

I discussed this with @ruzin and we're not going to merge this, and I want to explain why as a product decision rather than a comment on the implementation.

Right now we have low engagement and retention, and very limited telemetry to explain why. We're flying blind on basic questions - which features get used, where people drop off, whether they come back - and we can't fix what we can't see. At this stage, defaulting telemetry off would cut off the data we need most to make the product better.

A few things make default-on acceptable to us here:

- Everything captured is anonymous (a random UUID, no account or device identity tied to it) and metadata-only, never meeting content, transcripts, or file paths.
- The onboarding toggle stays visible; nothing about this is hidden from the user.
- Opt-out remains one click away in Settings at any time, and takes effect immediately.

So the position for now is: keep telemetry opt-out (default on), keep the toggle honest and visible, and keep opt-out fully respected. We'll revisit this if our posture needs to change - e.g. if an enterprise customer's security review requires opt-in by default, or if we hit a scale where the growth case for default-on no longer applies.

The _merge_for_save() gap you found is separate from this default-on/off call and still worth a narrow follow-up - happy to open a small issue for just that piece if useful.

Optic00:
Superseded by #352. After discussing the telemetry/launch-on-login privacy question with @ruzin, we're keeping both **enabled by default** (to preserve failure signal) but making them transparent instead of silent: onboarding discloses both for new users, and existing installs get a one-time soft consent notice on update. That's the opposite direction from this PR's strict opt-in (default off), so closing this in favour of #352 rather than reworking it. The one salvageable piece here (explicit `disableGeoip: true` on the PostHog constructors) is carried over in #352.

Issue #262, opened 2026-06-29
Auto-detect: false positive "Meeting detected" notifications from non-meeting apps


WilliamDrewett:
## Problem

The \"Meeting detected\" notification fires for apps that aren't meetings. Confirmed examples:

- **Conductor** (a native Mac app) triggering the notification when it plays a completion ping sound
- Any **browser mic access** (dictation, Loom, voice search, any WebRTC site) triggering it because all browser bundle IDs are in the allowlist

## Root cause

Two distinct holes in `handleMicEvent` (`app/main.js`):

### 1. Device-level events with no `app_id` pass unconditionally

```js
function isMeetingApp(evt) {
  if (!evt.app_id) return true; // ← notifies for ANY device-level audio event
  return MEETING_APP_ALLOWLIST.some((re) => re.test(evt.app_id));
}
```

When a Mac app plays a notification sound, macOS can briefly activate the microphone for acoustic echo cancellation (AEC). The mic-monitor picks this up as a device-level event with no `app_id`, which passes `isMeetingApp` unconditionally. This is what caused the Conductor ping to fire the notification.

Issue #116 proposes gating auto-detect to macOS 14+, which removes the *intended* use of this fallback (macOS 12/13 device-level events). But the false positive still exists on macOS 14/15 — the AEC/system-audio trigger is unrelated to the OS version.

### 2. Browsers are too broad

The allowlist includes all browser bundle IDs (`com.google.Chrome`, `com.apple.WebKit`, `org.mozilla.*`, etc.) because web meetings route through them. But the notification fires regardless of whether a calendar event exists — the calendar lookup is only used for the notification body text, not as a gate. Any browser microphone use triggers it.

## Proposed fix

Apply a two-tier approach in `handleMicEvent`:

**Tier 1 — native meeting apps (always notify):**
Zoom, Teams, Webex, FaceTime, Discord, GoToMeeting, BlueJeans, Pop. These have narrow, specific bundle IDs. A false positive is genuinely unlikely — if Zoom is capturing your mic, you're almost certainly in a meeting.

**Tier 2 — browsers + device-level events (require calendar corroboration):**
Only show the notification if `getCalendarEventForNow()` returns a match. If there's no calendar event, stay silent.

```
| Scenario                          | app_id               | Calendar? | Notify? |
|-----------------------------------|----------------------|-----------|---------|
| Zoom call (in calendar)           | us.zoom.xos          | Yes       | Yes     |
[...]

PR #345, opened 2026-07-17
fix(auto-detect): attribute mic capture from browser helper processes

evgeny-golubev:
## Summary
- Meeting auto-detect never fired for Arc (and other cases where the mic is opened by a browser helper process): `mic-monitor` resolved `app_id` only via `NSRunningApplication`, which cannot resolve helper/XPC processes, so it emitted `{event, pid, ts}` with no `app_id` — and `isMeetingApp` correctly discards app_id-less events on macOS 14+ (the #262 artifact guard).
- `mic_monitor.swift` now resolves the bundle id in three steps: `NSRunningApplication` → the CoreAudio process object's `kAudioProcessPropertyBundleID` (same macOS 14+ API family already used for the pid) → walking the exe path up to the containing `.app` bundle. Start-time info is cached so `stop` events stay attributed after the process exits.
- Arc's helpers identify as `company.thebrowser.browser.helper` (lowercase `browser`), which the case-sensitive allowlist entry `/^company\.thebrowser\.Browser/` rejected — widened to `/^company\.thebrowser\./i`.

## Test plan
- [x] `node --test meeting-detect.test.js` — 10/10 pass, including a new regression test for browser helper bundle ids (Arc helper lowercase id, Arc main app, Chrome helper)
- [x] `npm run test:unit` — full suite green (74 vitest + node tests)
- [x] `scripts/build-mic-monitor.sh` — compiles clean (arm64)
- [x] Live smoke test on macOS 27: rebuilt helper attributes Zoom (`us.zoom.xos`) as before, and a bare non-bundle process (ffmpeg capturing the mic) now emits `app_name` via the exe-path fallback instead of a fully anonymous event

---
## Summary by cubic
Fixes meeting auto-detect on macOS 14+ by attributing mic capture to the correct app even when a browser helper/XPC process holds the mic. Arc and other Chromium browsers now trigger detection reliably.

- **Bug Fixes**
  - Resolve bundle id via NSRunningApplication → CoreAudio `kAudioProcessPropertyBundleID` → bundle from executable path; cache start-time info so stop events stay attributed.
  - Make Arc allowlist case-insensitive to match helpers: `/^company\.thebrowser\./i`.
  - Add regression tests for browser helper bundle ids (Arc helper/main, Chrome helper); suite green.

<sup>Written for commit 4b760837b5fd51c05b74e03ee121639f9284546c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/meeting-detect.js
+++ b/app/meeting-detect.js
@@ -22,7 +22,7 @@ const MEETING_APP_ALLOWLIST = [
   /^com\.google\.Chrome/,              // Chrome (+ helpers)
   /^org\.chromium\./,                  // Chromium
   /^com\.microsoft\.edgemac/,          // Edge
-  /^company\.thebrowser\.Browser/,     // Arc
+  /^company\.thebrowser\./i,           // Arc (app: …thebrowser.Browser, helpers: …thebrowser.browser.helper)
   /^com\.brave\.Browser/,              // Brave
   /^org\.mozilla\./,                   // Firefox
 ];
--- a/app/meeting-detect.test.js
+++ b/app/meeting-detect.test.js
@@ -8,6 +8,15 @@ test('isMeetingApp accepts a known meeting app bundle id', () => {
   assert.strictEqual(isMeetingApp({ app_id: 'com.microsoft.teams2' }), true);
 });
 
+test('isMeetingApp accepts browser helper-process bundle ids', () => {
+  // Chromium browsers capture the mic in a helper process, whose bundle id
+  // may differ in case from the main app (Arc: company.thebrowser.Browser
+  // vs company.thebrowser.browser.helper).
+  assert.strictEqual(isMeetingApp({ app_id: 'company.thebrowser.browser.helper' }), true);
+  assert.strictEqual(isMeetingApp({ app_id: 'company.thebrowser.Browser' }), true);
+  assert.strictEqual(isMeetingApp({ app_id: 'com.google.Chrome.helper' }), true);
+});
+
 test('isMeetingApp rejects a non-allowlisted app bundle id', () => {
   assert.strictEqual(isMeetingApp({ app_id: 'com.example.Conductor' }), false);
 });
--- a/mic-monitor/mic_monitor.swift
+++ b/mic-monitor/mic_monitor.swift
@@ -74,7 +74,7 @@ func deviceIsRunningSomewhere(_ deviceID: AudioDeviceID) -> Bool {
 }
 
 @available(macOS 14.0, *)
-func pidsCapturingInput() -> Set<pid_t> {
+func processObjectsCapturingInput() -> [pid_t: AudioObjectID] {
     var listAddr = AudioObjectPropertyAddress(
         mSelector: kAudioHardwarePropertyProcessObjectList,
         mScope: kAudioObjectPropertyScopeGlobal,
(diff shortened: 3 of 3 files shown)


Optic00 (commented):
## Verdict

Comment - substantively this is a correct, tightly-scoped fix for a real detection gap and I'd happily merge it, but there's one easy regex tightening worth doing first (Important 2) and a note that the green checks don't actually cover the Swift change (Important 1). No blockers.

## Summary

On macOS 14+, Chromium browsers (Arc especially) capture the mic in a helper/XPC process that `NSRunningApplication` can't resolve, so `mic_monitor` emitted an `app_id`-less event that `isMeetingApp` correctly discards - and the meeting never got detected. This PR resolves the bundle id in three steps (`NSRunningApplication` -> CoreAudio `kAudioProcessPropertyBundleID` -> bundle id from the exe path), caches the start-time info so `stop` events stay attributed after the process is gone, and widens the Arc allowlist entry to case-insensitive so it also matches the lowercase helper id. A node regression test covers the allowlist change.

The 3-step resolution reuses the same macOS 14+ audio-process API family already used for the pid, and the start-time cache is a clean answer to "the process object is gone by the time the mic is released." Nicely done.

## Findings

### Important

**1. The green CI does not compile the Swift change.** `mic-monitor/mic_monitor.swift` is built only by `build-release.yml` (via `scripts/build-mic-monitor.sh`); the per-PR e2e checks don't, and no compiled binary is committed. So the whole Swift resolution path - the bulk of this PR - is covered only by the local arm64 compile + live smoke you documented, not by the green matrix. That's fine as long as a maintainer recompiles before merge; I just want it on the record that "all checks green" doesn't include this file.

**2. `app/meeting-detect.js:25`** - `/^company\.thebrowser\./i` now matches the *entire* `company.thebrowser.*` namespace, not just the browser. The intent is only to catch Arc's main app (`company.thebrowser.Browser`) and its helper (`company.thebrowser.browser.helper`, differing only in case), so keep the `browser` component while fixing the case: `/^company\.thebrowser\.browser(?:$|\.)/i`. Same behaviour for Arc, but a future non-browser tool in that namespace using the mic won't be mistaken for a meeting.

### Question

**3. Even tightened, this lands Arc on the unconditional-browser path.** Resolving the helper bundle id means Arc mic capture now fires "Meeting detected" on *any* mic use (dictation, voice search), like Chrome/Edge/Brave already do. That's parity, not a new regression, but it makes the still-open hole 2 in #262 (calendar-gating browser events) a bit more pressing. Fine to land as-is and handle #262 separately; flagging since the two touch the same path.

### Nit

**4. `mic-monitor/mic_monitor.swift` (~line 189, the pid-diff loop)** - transitions are tracked by pid only. If a capturing process exits and its pid is reused by another capturing process inside a single 1s poll, no stop/start is emitted and `infoCache` replays the previous app's attribution onto the new one. Extremely unlikely (needs pid reuse within 1s *and* the reused pid to also be capturing), and the cache is the right design since the process object is gone by stop time - but tracking the `AudioObjectID` (or a start identity) alongside the pid would close it.

**5. `mic-monitor/mic_monitor.swift` `bundleIDFromPath`** - it relies on the *first* `.app/` in the exe path resolving to the outermost bundle, which is exactly what you want for a helper nested inside `Foo.app/.../Bar Helper.app/...` (you get `Foo`'s id). Correct, but subtle enough to deserve a one-line comment so it isn't "simplified" later into last-match.

## Test plan verification

`node --test` covers the allowlist regex change (Arc helper lowercase id, Arc main, Chrome helper). The Swift resolution path has no automated coverage, which is inherent - it needs real CoreAudio + a mic-capturing helper on macOS 14+ - so the documented live smoke is the right ceiling here.

ruzin:
@Optic00 @evgeny-golubev so do these have more changes or?

Optic00:
Nothing blocking from my side, @ruzin - I'd merge it. The one open item from my review is optional: tightening the Arc entry to keep the `browser` component (`/^company\.thebrowser\.browser(?:$|\.)/i`) rather than matching the whole `company.thebrowser.*` namespace. It's a one-line, strictly-safer change, so either @evgeny-golubev grabs it now or it's a trivial follow-up - not worth holding the merge for.

One thing worth doing before you merge, though: run `scripts/build-mic-monitor.sh` once. The Swift helper is only compiled by the release workflow, not the per-PR checks, so "all green" here doesn't actually cover `mic_monitor.swift`. @evgeny-golubev documented a clean arm64 compile + live smoke, so this is just a sanity check that it still builds on your side.

ruzin:
@Optic00 ok going with your lead here, on these, if you're happy, feel free to approve and merge.

ruzin (approved):

PR #348, opened 2026-07-18
ci(release): auto-announce releases to Discord

ruzin:
## Summary

Automates the Discord release announcement you currently post by hand. On every tagged release, the `release` job now posts an embed to a Discord channel webhook with the one-line summary, up to 4 headline features, and the contributor thank-you — all parsed from the annotated-tag release notes.

## How it works
- New `scripts/discord-release-announce.mjs` parses the tag notes (summary line, `- **Feature** — …` bullets, and the `### Thanks to our contributors` paragraph) and POSTs a Discord webhook embed.
- New **Announce release on Discord** step in `.github/workflows/build-release.yml`'s `release` job, right after the GitHub Release is created.

## Why a step, not a separate `on: release: published` workflow
The release is created with `GITHUB_TOKEN`, and GitHub **does not fire downstream workflow triggers for `GITHUB_TOKEN`-authored events** — a standalone announce workflow would never run. Putting the step in the same job sidesteps that.

## Safety
- `continue-on-error: true` on the step **and** a no-webhook no-op in the script → a Discord problem can **never** fail the actual DMG release.
- **Inert until configured**: no `DISCORD_WEBHOOK_URL` secret ⇒ it logs "skipping" and exits 0 (forks stay quiet).

## One-time setup (required for it to actually post)
1. In Discord: **announcements channel → Edit Channel → Integrations → Webhooks → New Webhook**, copy the URL.
2. Add it as a repo secret named **`DISCORD_WEBHOOK_URL`** (`gh secret set DISCORD_WEBHOOK_URL`, or Settings → Secrets and variables → Actions).

## Preview (dry-run against the real v0.6.0 notes)
`DISCORD_DRY_RUN=1 node scripts/discord-release-announce.mjs` renders:
> 📢 **Steno v0.6.0 is out**
> Recording that gets out of your way — a docked transcription pill, instant stop, continue-recording, and a My notes tab…
> **What's new** • Transcription pill dock • Instant stop • Continue recording • My notes tab
> **This release includes work from @Optic00, @WilliamDrewett, @Vassista, @luchfilip, and @valentinweyer. 🙏**
> [Download & full release notes →]

## Notes / follow-ups
- Contributor `@handles` are **GitHub** handles, so on Discord they render as plain text (not real pings) — same as the manual post. Real Discord pings would need a GitHub→Discord user map + bot (happy to add if you want).
- This posts into the channel; auto-**crossposting** an Announcement (News) channel to following servers needs a bot token + the crosspost API — can add as a follow-up if you use a News channel.

Documented in CLAUDE.md release process (step 9).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Automatically announces each tagged release to Discord via a webhook. Sends a friendly @here ping with a branded embed: summary, up to 4 headline features, and the contributor thank‑you — replacing the manual post.

- **New Features**
  - Added `scripts/discord-release-announce.mjs` and an “Announce release on Discord” step in `.github/workflows/build-release.yml`; posts “🚀 Hey @here — Steno vX is out!” in message content (with `allowed_mentions`) and an embed with the summary, features, thanks, and a full notes link. Lives in the `release` job to avoid `GITHUB_TOKEN` trigger limits.
  - Safe by default: `continue-on-error`, no‑op if the webhook is missing, and network/webhook failures are caught and logged (exit 0); use `DISCORD_DRY_RUN=1` to print the payload for local testing.

- **Migration**
  - Create a Discord channel webhook and copy its URL.
  - Add it as the `DISCORD_WEBHOOK_URL` repo secret.

<sup>Written for commit 22caf12d76a7221f1112fe9cb3068e70ac4c41d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/348?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -273,3 +273,25 @@ jobs:
           release/*.yaml
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    # Announce the release in Discord. Lives here (not a separate
+    # `on: release: published` workflow) because the release above is created
+    # with GITHUB_TOKEN, and GitHub does NOT fire downstream workflow triggers
+    # for GITHUB_TOKEN-authored events — so a standalone announce workflow would
+    # never run. continue-on-error + the script's no-webhook no-op guarantee a
+    # Discord problem can never fail the actual release. No-ops until the
+    # DISCORD_WEBHOOK_URL repo secret is set (so forks stay quiet).
+    - name: Setup Node.js (Discord announce)
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+
+    - name: Announce release on Discord
+      if: startsWith(github.ref, 'refs/tags/v')
+      continue-on-error: true
+      env:
+        DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
+        VERSION: ${{ github.event.inputs.version || steps.package_version.outputs.version }}
+        RELEASE_URL: https://github.com/${{ github.repository }}/releases/tag/v${{ github.event.inputs.version || steps.package_version.outputs.version }}
+        RELEASE_NOTES: ${{ steps.tag_body.outputs.text }}
+      run: node scripts/discord-release-announce.mjs
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -220,7 +220,8 @@ Releases are automated via `.github/workflows/build-release.yml`. Never create r
    - Builds signed + notarized DMGs for both arm64 and x64
    - Creates a GitHub Release with the tag message as the body
    - Uploads both DMGs as release assets
-10. Do NOT build DMGs locally for releases, do NOT use `gh release create` manually.
+   - Posts a release announcement to Discord (summary + a few headline features + the contributor thank-you), via the `Announce release on Discord` step (`scripts/discord-release-announce.mjs`). This is automatic and requires no manual Discord post. It no-ops unless the `DISCORD_WEBHOOK_URL` repo secret is set (an incoming webhook for the announcements channel), and is `continue-on-error` so a Discord hiccup never fails the release.
+10. Do NOT build DMGs locally for releases, do NOT use `gh release create` manually. Do NOT post the Discord release announcement by hand — the workflow does it.
 
 ## Session Logging
 When the user says "log session" or similar (e.g., "update session log", "document this session"):
--- a/scripts/discord-release-announce.mjs
+++ b/scripts/discord-release-announce.mjs
@@ -102,11 +102,21 @@ if (process.env.DISCORD_DRY_RUN === '1') {
   process.exit(0);
 }
 
-const res = await fetch(webhook, {
-  method: 'POST',
-  headers: { 'content-type': 'application/json' },
-  body: JSON.stringify(payload),
-});
+let res;
+try {
+  res = await fetch(webhook, {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: JSON.stringify(payload),
+  });
+} catch (err) {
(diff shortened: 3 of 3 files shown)


Optic00 (dismissed):
## Verdict

Approve. Clean, low-risk release automation: it can't fail the release (`continue-on-error` + a no-webhook no-op), leaks no secret, and the workflow wiring is correct. The points below are polish, and the one I'd actually action is the fetch guard.

## Summary

Automates the by-hand Discord release post: a new `release`-job step runs `scripts/discord-release-announce.mjs`, which parses the annotated-tag notes (summary line, `- **Feature** — …` bullets, the contributor thank-you) and POSTs a Discord webhook embed. It no-ops without the `DISCORD_WEBHOOK_URL` secret (forks stay quiet) and is `continue-on-error`, so a Discord problem can never break a release. CLAUDE.md's release checklist is updated to match.

The "why a step, not a separate `on: release: published` workflow" call is right and worth the comment - a `GITHUB_TOKEN`-authored release doesn't fire downstream workflow triggers, so a standalone announce workflow would silently never run. The double no-op safety, the `DISCORD_DRY_RUN` mode, and keeping the webhook in an env var (never logged) are all good.

## Findings

### Important

**1. `scripts/discord-release-announce.mjs:105`** - the script's header promises it "NEVER fails the release... logs and exits 0," but that only holds for HTTP error *responses* (`!res.ok`). A transport-level `fetch` failure (DNS, timeout, connection refused) rejects the top-level `await` and exits 1. `continue-on-error` keeps that from failing the release, so the impact is only a red step in the Actions UI - but it breaks the script's own stated contract. Wrap the `fetch`/`res` block in try/catch and exit 0 on throw.

### Nit

**2. `scripts/discord-release-announce.mjs:74,84`** - `contributors` is `rest.join(' ')` with no truncation, and feature *names* aren't bounded either, so the assembled `description` isn't actually guaranteed "well under" Discord's 4096-char embed cap (line 81's comment). A long thank-you paragraph would get the embed rejected with 400 and the announcement dropped (gracefully - `!res.ok` exits 0 - but silently). Truncating `contributors` and capping the final description would make the comment true.

**3. `scripts/discord-release-announce.mjs:39-43`** - the summary is "first line that isn't a heading/bullet/quote," scanned over the whole document. Our notes lead with the one-line summary so this is fine in practice, but if a release ever opened with a section instead, the first prose it hits would be the contributor paragraph - which then becomes both the summary and the contributor block. A cheap guard: only take the summary from before the first `###`.

**4. No unit test for the parser.** The notes-parsing (summary/features/contributors) is exactly the kind of pure logic this repo covers with a sibling `*.test.js` (`meeting-detect.test.js`, `analytics-helpers`, etc.), and `DISCORD_DRY_RUN=1` already makes it testable without a webhook. A small `discord-release-announce.test.js` feeding one sample notes string would lock the format contract and catch the edge cases above.

### Question

**5.** Feature extraction keys on the exact `- **Name** — desc` bullet style, but the release checklist (CLAUDE.md step 5) only says "headline features grouped under `### Section` headers," not that they must be bold-name bullets. Notes written as prose under a section would yield zero features (it degrades gracefully to summary + download link). Worth either pinning the bullet format in the checklist or accepting the degradation as intended.

### Praise

The GITHUB_TOKEN / downstream-trigger reasoning is the kind of thing that's easy to get wrong and costs a silent no-op; documenting it inline is exactly right. Belt-and-suspenders failure handling throughout, and the dry-run mode makes it easy to sanity-check notes formatting before a real tag.

ruzin:
@Optic00 all comments addressed, thanks for your review

PR #368, opened 2026-07-22
fix(processing): break the forever-spinner when a stop produced no job (#343)

### LINKED ISSUE #343: Processing screen spins forever when a stop produces no queued job
## Summary

Stopping a recording navigates to the processing screen unconditionally, but a job is only queued if the capture actually produced a file. When those two disagree, the processing screen spins forever: no `processingComplete` event ever arrives, and there is no watchdog. The user's only way out is the "Home" button.

This is the residual of #212 after #223/#225 fixed the permission-denial case it was filed for. Splitting it out because the cause and the title no longer match: this is not permission-specific, and it does not need a reporter round-trip. Analysis background: https://github.com/ruzin/stenoai/issues/212#issuecomment-5001673100

## Mechanism

- `stopRecording()` navigates to `/meetings/processing` on every stop, before any backend round-trip (`app/renderer/src/hooks/useRecording.ts:182-192`).
- A job is only queued via renderer `stopCapture()` → `recorder.onstop` → `process-system-audio-recording` → `addToProcessingQueue` (`app/main.js:7361`), guarded on a closable file at `useSystemAudioCapture.ts:660-681`. `stop-recording-ui` queues nothing (`app/main.js:4624-4653`).
- `Processing.tsx` leaves the screen only on `processingComplete` or a manual "Home" click. No watchdog.

## Two ways in

1. **Stop clicked while `getUserMedia` is still pending.** The Stop button is live before capture is, which is realistic on first run while the macOS permission dialog is up. `recorderRef.current` is still null, so `stopCapture()` early-returns with no blob.
2. **`processSystemAudio` returns `{success: false}` without throwing**, and the caller treats it as a completed handoff. Unrelated to permissions, same dead end.

Neither loses data (nothing was captured), so this is a confusing dead end rather than a trap. Low severity, but it is a silent failure with no user-facing explanation.

## Proposed fix

A watchdog on `Processing.tsx`, keyed on **queue state**, not elapsed time: if `!isProcessing && queueSize === 0 && !hasRecording` holds across several consecutive polls, show the existing error panel with a "nothing was recorded" message. `retryAudioFile` stays null so "Try again" is correctly disabled.

Keying it on "no events for N seconds" would be wrong: long transcriptions have legitimately quiet stretches and would false-trip.

Also worth unwrapping the swallowed `{success: false}` from `processSystemAudio` so close/enqueue failures surface instead of only being logged.

Not proposed: a permission preflight gate (redundant after #223/#225, and a macOS-only branch for no extra coverage), or removing the optimistic navigation (costs the instant-feedback UX for a narrow case).

## Testing note

An **immediately rejecting** `getUserMedia` does not reproduce path 1: that fires the #223 error path and lands cleanly in idle. Reproducing it needs a **deferred promise**, with Stop clicked while the permission is still pending.

`recording-lifecycle.t2` drives the raw IPC bridge and therefore bypasses the React navigation that carries this bug, so it is a template rather than the place to add coverage. A new spec needs to drive the real UI.

Optic00:
Fixes #343.

## The bug
Stopping a recording navigates to the processing screen unconditionally, but a job is only queued if the capture actually produced a file. When a stop produces **no** job (Stop hit while `getUserMedia` is still pending → `stopCapture` early-returns with no blob; or `processSystemAudio` resolves `{success:false}`), no `processing-complete` ever fires and the screen spins forever. The only escape was the Home button. No data is lost (nothing was captured) — it's a confusing dead end with no explanation.

## The fix — a queue-state watchdog (renderer-only)
`Processing.tsx` now watches for a genuinely idle-and-empty queue rather than a wall clock (long transcriptions have quiet stretches a time-based timeout would false-trip). When the queue stays `status==='idle' && queueSize===0` with no processing activity for a sustained window, it swaps the spinner for a calm **"Nothing to process"** panel (distinct from the crash copy; "Try again" stays disabled since there's no backing job).

The threshold is the subtle part. The normal stop→enqueue handoff is **not instant**: after stop, main clears the recording state but `process-system-audio-recording` waits up to **~8s** for the live-transcript sidecar to drain before `addToProcessingQueue` runs. During that window a queue poll legitimately returns idle+empty for a recording that really has speech. So the watchdog:
- requires **8 consecutive idle ticks (~12s)** to clear that bounded ~8s drain with margin (a genuine dead-end stays idle+empty indefinitely, so it still trips, just a few seconds later);
- only counts once the queue has **successfully loaded real data** (absent query data defaults to idle/empty — a slow first IPC must not look like a dead-end);
- **does not count while the window is hidden** (the poll cadence drops to 10s and the cache goes stale);
- permanently disarms on any real activity (a streamed chunk, progress, or a stage move past `transcribing`), and ignores `summaryFile`-scoped events so a concurrent reprocess of another meeting can't disarm it.

`useRecording` now exposes `queueSize` and `isQueueSuccess`; the header/dock "Processing" chips hide once the watchdog concludes nothing was captured.

## Tests
New `e2e/specs/processing-watchdog.t1.spec.ts` (T1, renderer + mock IPC) drives the real state **sequence** via a scripted `get-queue-status` (a small `STENOAI_E2E_QUEUE_STATE_PATH` seam in `e2e-mock-ipc.js`):
- idle+empty that persists past the handoff → trips → "Nothing to process", "Try again" disabled, chip gone;
- **late-enqueue regression guard:** optimistic processing → idle+empty handoff gap → late enqueue → must NOT trip (verified fail-before with the old ~4.5s threshold, pass-after with ~12s);
- an actively-processing job never trips.

## Review
Independent cross-family review (Codex) caught a critical false-positive in the first draft — the original ~4.5s threshold fired during the legitimate ~8s handoff and would have declared a real recording empty. That, plus a concurrent-reprocess disarm gap and a test that didn't reproduce the real handoff sequence, are all fixed here.

## Verification
- `typecheck:renderer`, `lint:renderer` (0 errors in changed files), `build:renderer` — clean.
- T1 watchdog spec — 3 passed. `instant-stop.t1` + `pill-dock.t1` regression — 9 passed.
- Renderer-only; no `app/main.js` change.

---
## Summary by cubic
Fixes the infinite spinner on the Processing screen when a stop produces no job by adding a queue-state watchdog and a clear “Nothing to process” state. Also recovers cleanly when a new recording starts without a remount, even if it reuses the same name. Fixes #343.

- **Bug Fixes**
  - Added a queue-state watchdog in `Processing.tsx` that shows “Nothing to process” after 8 consecutive idle+empty ticks (~12s) once the queue has loaded; it pauses while the window is hidden, disarms on real activity, and ignores `summaryFile`-scoped events.
  - Recover across new sessions without a remount, including same-name sessions: reset per-session UI and re-arm the watchdog via a name-independent generation; hide the header chip and `ProcessingDock` once nothing was captured.
  - Exposed `queueSize` and `isQueueSuccess` from `useRecording`; disabled “Try again” in the “Nothing to process” state with calmer copy; hid the header “Processing” chip in error.
  - Introduced an e2e seam via `STENOAI_E2E_QUEUE_STATE_PATH`; added T1 specs for: watchdog trip, late-enqueue handoff (no false trip), active processing (never trips), recovery after a new session, and recovery when the new recording reuses the same name.

<sup>Written for commit 0489b64fe77f7008bb85ad53bff69a9a762ea94c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -229,20 +229,50 @@ function install({ ipcMain }) {
       }
       return { success: true };
     },
-    'get-queue-status': async () => ({
-      success: true,
-      isProcessing: rec.processing,
-      queueSize: 0,
-      currentJob: rec.processing ? rec.sessionName : null,
-      currentReprocesses: [],
-      hasRecording: rec.active,
-      isPaused: rec.paused,
-      elapsedSeconds: rec.active
-        ? Math.floor(((rec.paused ? rec.pausedAt : Date.now()) - rec.startedAt) / 1000)
-        : 0,
-      sessionName: rec.active || rec.processing ? rec.sessionName : null,
-      recordingSummaryFile: rec.active ? rec.appendTo : null,
-    }),
+    // Scripted-sequence seam for the processing-watchdog T1: when
+    // STENOAI_E2E_QUEUE_STATE_PATH points at a JSON file, each poll returns that
+    // file's contents (merged over the idle defaults) so a spec can drive the
+    // real stop→enqueue state SEQUENCE over time (optimistic processing → a
+    // legitimate idle+empty handoff gap → late enqueue) by rewriting the file
+    // between polls. Falls through to the rec-based machine if the file is
+    // absent/unreadable (e.g. not yet written on the first poll).
+    'get-queue-status': async () => {
+      const statePath = process.env.STENOAI_E2E_QUEUE_STATE_PATH;
+      if (statePath) {
+        try {
+          const override = JSON.parse(fs.readFileSync(statePath, 'utf-8'));
+          return {
+            success: true,
+            isProcessing: false,
+            queueSize: 0,
+            currentJob: null,
+            currentReprocesses: [],
+            hasRecording: false,
+            isPaused: false,
+            elapsedSeconds: 0,
+            sessionName: null,
+            recordingSummaryFile: null,
+            ...override,
+          };
+        } catch {
+          /* file not ready yet — fall through to the rec-based default below */
+        }
+      }
+      return {
+        success: true,
+        isProcessing: rec.processing,
+        queueSize: 0,
+        currentJob: rec.processing ? rec.sessionName : null,
+        currentReprocesses: [],
+        hasRecording: rec.active,
+        isPaused: rec.paused,
+        elapsedSeconds: rec.active
+          ? Math.floor(((rec.paused ? rec.pausedAt : Date.now()) - rec.startedAt) / 1000)
+          : 0,
+        sessionName: rec.active || rec.processing ? rec.sessionName : null,
+        recordingSummaryFile: rec.active ? rec.appendTo : null,
+      };
+    },
 
     // Live transcript backfill. Real main.js populates liveTranscriptState from
     // the ASR sidecar (a model) — unreachable in T1 — so we seed it here. With
--- a/app/renderer/src/hooks/useRecording.ts
+++ b/app/renderer/src/hooks/useRecording.ts
@@ -270,6 +270,12 @@ export function useRecording() {
      *  Set rather than array so consumers can do O(1) membership checks
      *  inside the meetings list map. */
     reprocessingSummaryFiles: reprocessingSummaryFiles,
+    /** True once the queue poll has resolved successfully at least once. The
+     *  processing-screen watchdog must not count "idle+empty" ticks before real
+     *  data has arrived — absent query data defaults to status:'idle' /
+     *  queueSize:0, so a slow first IPC would otherwise look like a no-job
+     *  dead-end (issue #343). */
+    isQueueSuccess: queue.isSuccess,
     startRecording,
     stopRecording,
     pauseRecording,
--- a/app/renderer/src/hooks/useRecording.ts
+++ b/app/renderer/src/hooks/useRecording.ts
@@ -246,6 +246,12 @@ export function useRecording() {
   return {
     status,
     elapsed: queue.data?.elapsedSeconds ?? 0,
+    /** Number of jobs waiting in the processing queue. Combined with `status`
+     *  it tells the processing screen whether a job actually exists — a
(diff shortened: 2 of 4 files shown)


Optic00:
Thanks @cubic-dev-ai - both are valid. Handling:

**P1 (stuck on "Nothing to process" when a new recording starts on the still-mounted route) - fixed in c18dd43.** The route swaps `activeSession` in place across back-to-back recordings without a remount, but the watchdog error state was only reset on mount. It now resets per-session: the UI state during render (React's "adjust state when a value changes" pattern, so no flash of the prior panel), and the watchdog refs (`sawActivityRef` especially) + the interval re-arm in an effect keyed on `activeSession`. Covered by a new T1 case ("a new session after 'Nothing to process' recovers without a remount", verified fail-before/pass-after). The threshold, `isQueueSuccess` gate and hidden-window guard still apply, so recovery can't reintroduce a false trip.

**P2 (a no-job recording can still spin when another recording/reprocess is concurrently active) - real, but a known limitation of this renderer heuristic, not a regression, so I'm tracking it as a follow-up rather than in this PR.** The watchdog reasons about the *global* queue status, so while another job is active it never observes idle+empty for this screen. Note this is the same behaviour as before this PR (that concurrent case spun before too) - this PR strictly improves the common single-recording case the issue reports. A correct fix needs a *session-bound* signal: have the stop handoff surface whether a job was actually enqueued for **this** session (related to unwrapping the swallowed `processSystemAudio` `{success:false}`), which is a capture-hook/main change the issue deliberately scoped out. I'll open a follow-up for that.

PR #375, opened 2026-07-22
fix(update-toast): expand clickable area on Restart button

WilliamDrewett:
## Description

The update-ready toast's Restart button only responded to clicks on its bottom half. `UpdateToast` is `position: fixed` at the top-right, overlapping the strip `MainToolbar` marks `-webkit-app-region: drag` for window dragging. Electron computes that drag region from CSS, independent of visual stacking/z-index, so without an explicit `no-drag` opt-out, macOS's window-drag hit test intercepted clicks on the upper portion of the pill — only the sliver poking below the toolbar's drag strip was clickable.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Verified via a standalone browser repro of the toast markup, confirming the fix aligns the clickable hit box with the visible pill
- [x] Renderer typecheck passes (`npm run typecheck:renderer`)
- [ ] Not yet verified in the packaged Electron app (drag-region behavior is Electron-specific and wasn't re-tested there)

## Additional Notes

Same class of bug already called out in `Sidebar.tsx`/`MainToolbar.tsx` for the sidebar toggle button — any `position: fixed` element overlapping a draggable strip needs an explicit `no-drag` opt-out.

---
## Summary by cubic
Fixes the update toast so the Restart and Dismiss buttons are fully clickable. The toast container now sets `WebkitAppRegion: 'no-drag'` to prevent the top toolbar’s drag region from intercepting clicks.

<sup>Written for commit 2116fd6eb1106ed9036315be1172e3b758e0c8a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/375?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/components/UpdateToast.tsx
+++ b/app/renderer/src/components/UpdateToast.tsx
@@ -34,7 +34,13 @@ export function UpdateToast() {
         border: '1px solid var(--border-subtle)',
         boxShadow: 'var(--shadow-md)',
         fontFamily: 'var(--font-sans)',
-      }}
+        // Overlaps MainToolbar's draggable top strip. Without an explicit
+        // no-drag here, macOS's window-drag hit test (computed from CSS
+        // regions, not paint z-order) swallows clicks on the upper portion
+        // of the buttons below — only the sliver poking below the strip
+        // was clickable.
+        WebkitAppRegion: 'no-drag',
+      } as React.CSSProperties}
       role="status"
       aria-live="polite"
     >


Optic00 (approved):
## Verdict

Approve. A minimal, correct fix that makes the update toast's Restart/Dismiss buttons fully clickable, and it follows the exact `no-drag` pattern already established in the codebase.

## Summary

The update-ready toast is `position: fixed` in the top-right and overlaps the draggable top strip that `MainToolbar` marks `-webkit-app-region: drag`. Electron computes drag regions from CSS independently of paint z-order, so without an explicit opt-out macOS's window-drag hit test swallowed clicks on the upper portion of the buttons - only the sliver below the strip was clickable. The PR sets `WebkitAppRegion: 'no-drag'` on the toast container, carving the whole pill out of the drag region.

## Findings

### Praise

- Correct root cause and the right fix: it matches the existing precedent exactly (`MainToolbar.tsx:72`, `Sidebar.tsx:564`, and the `.sb-top > * { -webkit-app-region: no-drag }` rule in `globals.css:339`), so it's consistent with how the rest of the app carves interactive controls out of the drag strip. The inline comment explaining the CSS-region-vs-z-order hit test is genuinely helpful.
- No cross-platform risk: `-webkit-app-region` is a no-op where no drag region is defined, so setting it unconditionally is safe on the Windows title-bar path, consistent with the unconditional usage in `MainToolbar`/`Sidebar`. The `as React.CSSProperties` cast is the same one used elsewhere for this non-standard property.
- Honest testing note: the author flags it wasn't re-verified in the packaged Electron app, which is exactly the caveat that matters here (the drag-region hit test is Electron-native and can't be asserted from the renderer-only/T2 harness).

### Question

- Making the whole toast `no-drag` means the top-right pill area no longer drags the window while the toast is visible. That's the correct trade-off for a small transient toast, so no change requested - just confirming it's intentional.

PR #384, opened 2026-07-22
fix(renderer): nav hover contrast, dock alignment, and tooltips

WilliamDrewett:
## Description

Fixes the sidebar/settings-nav hover effect, which was invisible in light mode because it used the same color as the sidebar's own background. While in there, also flattened the active-row style to match hover, fixed a few other missing/misleading hover states, centered the recording pill with the chat composer, swapped the New note icon to a plus, and added tooltips to a number of previously title-only icon buttons across the sidebar, toolbar, transcript panel, and recording pill.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tested on macOS

## Additional Notes

Verified each fix against the running dev app (mocked IPC) with scripted hover/click checks; `typecheck:renderer` and `lint:renderer` both pass.

---
## Summary by cubic
Fixes invisible nav hovers in light mode and improves clarity with better contrast, anchored dock alignment, clearer icons, and contextual tooltips. Adds hover feedback where it was missing and removes misleading hovers.

- **Bug Fixes**
  - Sidebar/settings row hover is now visible in light mode using `--surface-active`; active row matches hover.
  - Anchored the recording pill and transcript toggle to the composer row with small mb offsets, keeping them centered even when the AskBar expands.
  - Added hover feedback to Stop buttons and back links; removed the header-wide hover on the transcript panel.

- **New Features**
  - Added tooltips to key controls: Help, Settings, folder create, sidebar toggle, recording options, copy transcript, language picker, resume/hide/stop.
  - Replaced the “New note” pen with a plus icon; added a language icon to the transcript language selector.

<sup>Written for commit f739908dc4d1a2387d81a8f6a1c662f3f7a5fdc6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/384?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/components/AskBar.tsx
+++ b/app/renderer/src/components/AskBar.tsx
@@ -167,7 +167,9 @@ export function TranscriptToggle() {
       aria-label={transcriptOpen ? 'Hide transcript' : 'Show transcript'}
       aria-pressed={transcriptOpen}
       title="Transcript"
-      className="pointer-events-auto inline-flex h-11 shrink-0 cursor-pointer items-center justify-center gap-0.5 rounded-full border-0 px-3 transition-colors"
+      // mb-[3px] optically centers the 44px toggle against the 50px composer
+      // row it sits beside (PrimaryDock aligns the row items-end; see there).
+      className="pointer-events-auto mb-[3px] inline-flex h-11 shrink-0 cursor-pointer items-center justify-center gap-0.5 rounded-full border-0 px-3 transition-colors"
       style={{
         background: transcriptOpen ? 'var(--surface-active)' : 'var(--surface-raised)',
         border: '1px solid var(--border-subtle)',
--- a/app/renderer/src/components/LiveDock.tsx
+++ b/app/renderer/src/components/LiveDock.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { ChevronUp, Play, Square } from 'lucide-react';
 import { AudioWave } from '@/components/AudioWave';
+import { Tooltip, TooltipTrigger, TooltipContent } from '@/components/ui/tooltip';
 import { useRecording } from '@/hooks/useRecording';
 import { useLiveTranscript } from '@/hooks/useLiveTranscript';
 import { useLiveTranscriptOpen } from '@/hooks/liveTranscriptOpenStore';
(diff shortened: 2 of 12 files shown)


Optic00 (changes requested):
## Verdict

Request changes - one small but visible layout regression in the expanded-chat state; everything else checked out (both themes, a11y, drag-region) and is ready to go.

## Summary

This PR fixes an invisible light-mode nav hover: sidebar rows and footer controls used `--surface-hover`, which resolves to `--paper-1` - the same value as the sidebar's own `--surface-sunken` background - so the hover was a no-op in light mode. It switches those to `--surface-active`, flattens the active row to match, removes a fake clickable affordance on the transcript header, re-centers the recording pill against the composer, swaps the "New note" pen for a plus, and converts `title=` on icon-only controls to Radix tooltips. Renderer-only. I verified the token analysis in the new `globals.css` comment against the definitions - it is exactly right, dark mode still differentiates, and the split (sunken-sidebar elements -> `--surface-active`, page/raised elements -> `--surface-hover`) is coherent.

## Findings

### Important

- `app/renderer/src/components/PrimaryDock.tsx:57` - `items-end` -> `items-center` centers the left control against the whole AskBar flex column, not just the composer. The chat panel renders in-flow inside that column (`maxHeight: 360`), so with an expanded chat session (and likewise the suggestion-chips row) the transcript toggle drifts toward the vertical middle of a ~400px stack instead of staying beside the composer. Suggestion: keep `items-end` and optically center the left slot against the composer row only, e.g. `self-end` plus a small `mb-*` offset of half the height difference - that fixes the 1-line case you were targeting without regressing the expanded one.
- `app/renderer/src/components/Sidebar.tsx:413` - the "New folder" button is `opacity-0` until the group is hovered, so a keyboard user tabbing to it lands on a fully transparent control with a transparent focus outline. Pre-existing (this PR only wraps it in a tooltip, which does surface on focus - partial mitigation), but since you're touching this exact control, `focus-visible:opacity-100` plus a visible focus ring would close the gap cheaply.

### Nit

- `app/renderer/src/routes/OrgShared.tsx:174`, `Processing.tsx:176`, `Recording.tsx:58` - the `hover:text-[color:var(--fg-1)]` half of these back-link hovers has no effect: the inline `style={{ color: 'var(--fg-2)' }}` wins over any class. Pre-existing, and your new background hover works, but since this PR is about exactly this kind of feedback, moving the color into a `text-[color:var(--fg-2)]` class would make the text brighten too.
- `app/renderer/src/components/Sidebar.tsx:352` / `:436` - the drag-over highlights still use `--surface-hover`, i.e. the same "invisible on the sunken sidebar in light mode" trap this PR fixes for hover. `:436` is saved by its `ring-1`, but `:352` ("All meetings") has no ring, so its drop highlight is invisible in light mode. Same one-token fix.
- `app/renderer/src/components/LiveTranscriptBar.tsx:137` vs `:150` - Copy now uses a Radix tooltip while its sibling Minimize (and the footer Resume/Stop) still rely on native `title`. Accessible names are fine either way; just inconsistent tooltip behavior within one header.
- `app/renderer/src/components/Sidebar.tsx` - the Help button's tooltip says "Documentation" while its `aria-label` stays "Help", so sighted and screen-reader users get different names for the same control; pick one.

### Question

- `app/renderer/src/globals.css:355` - the active row is now flat `--surface-active`, identical to hover, so at rest it's distinguished from an inactive row only by `font-weight: 500` and the `--fg-1` icon color. Intentional flatten per your description - just flagging that the weight/icon color now carry the whole "active" signal.

### Praise

- The `globals.css` comments explaining *why* (`--surface-active` vs `--surface-hover` on the sunken sidebar, the non-clickable `mv-transcript-head`) document the exact trap for the next person.
- Every removed `title=` kept its `aria-label` (and `aria-pressed` where present), so accessibility survives the tooltip migration intact, with no double-tooltips - and the de-hovered `mv-transcript-head` is correctly non-clickable in both usages.
- The fixed-position sidebar toggle keeps `-webkit-app-region: no-drag` after tooltip-wrapping, so no drag-region hit-test regression.

CI is fully green including both T2 lanes; as a visual-only change this needs no new e2e spec.

Optic00:
Pushed the dock fix from the review directly (f739908) so this isn't blocked on a round-trip - back to items-end with small optical offsets against the composer row (mb-1 for the pill, conditional so the standalone pill stays put; mb-[3px] for the toggle). Hope that's okay! The remaining nits are yours if you want them - none of them block, and I'll flip my review once CI is green on this commit.

Optic00 (approved):
Approving - the dock-alignment regression I flagged is resolved by f739908 (pushed here): back to `items-end` with small optical offsets against the composer row (conditional for the standalone pill), so the toggle/pill stay beside the composer whether or not a chat is expanded. CI is fully green on this commit including both T2 lanes.

The remaining nits from the review (inert `hover:text` on the back-links, the drag-over highlights on `--surface-hover`, the New-folder keyboard-focus gap, the Copy/Minimize tooltip inconsistency) are all pre-existing or cosmetic and don't block - happy to leave them to a follow-up. Thanks for the fix, the hover contrast and tooltips are a real improvement.

Optic00:
@ruzin this one's ready but needs your hand for the merge. I pushed the dock-alignment fix from my review directly onto the branch (f739908) and approved, but since I'm now the last pusher my own code-owner approval doesn't satisfy the rule - it needs an independent approve from you. CI is fully green (both T2 lanes) and Codex reviewed the fix. Mind giving it a look and merging? The remaining review nits are all pre-existing/cosmetic and don't block.

ruzin:
Maintainer review — thanks @WilliamDrewett. A couple of the cubic findings look substantive and worth addressing before merge:
- The **hover-contrast** change appears **ineffective** where an adjacent inline `color` declaration overrides `hover:text-[color:var(--fg-1)]` (OrgShared / Processing / the Home control) — so the intended contrast doesn't actually apply. Moving the color to the hover rule (or dropping the inline override) would fix it.
- **P3**: the new tooltip on the Stop button is attached to a *disabled* button when `stopped`, so it can't open on hover/focus — wrap the trigger or gate it.
- **P2**: expanding an existing chat session moves the transcript toggle alongside the chat history (layout shift) — confirm that's intended.
No rush — not release-critical; happy to re-review once updated.

PR #398, opened 2026-07-23
feat: make the global record shortcut toggleable + documented (#396)

### LINKED ISSUE #396: Cmd+shift+R Record shortcut is undocumented and unconfigurable
CMD+SHIFT+R is a pretty common shortcut and it appears that steno hijacks it and doesn't document that it does so.

At the least, it should be able to be disabled via config.

Optic00:
Closes #396.

The app registered a **global (system-wide)** hotkey `CommandOrControl+Shift+R` to toggle recording — unconditionally at startup, hardcoded, with no way to disable it. `Cmd+Shift+R` is the browser hard-reload shortcut, so the global hijack broke that system-wide. It was also undocumented (only an in-app hint) and unconfigurable, as the reporter (JackCA) noted.

## What this does
Adds a **Settings toggle** ("Global record shortcut", General tab) to enable/disable the global hotkey, mirroring the existing `notifications_enabled` plumbing end to end:

- **Config:** `record_hotkey_enabled` (default ON) in `src/config.py` + `get-record-hotkey` / `set-record-hotkey` CLI commands.
- **main.js:** startup registration is gated on a **direct `config.json` read** (`recordHotkeyEnabledFromDisk`, no Python subprocess — so it works under E2E). A new **idempotent** `applyRecordHotkey()` helper (guards on `isRegistered()`) is shared by startup and the `set-record-hotkey` IPC handler, which **live-applies** the change without a relaunch (`unregister(ACCEL)` only — never `unregisterAll()`). The getter returns `{ enabled, registered }`.
- **UX:** Settings shows a "couldn't register — another app may be using it" note when `registered === false`. All `⌘⇧R` copy in `hero.ts` **and** the Home empty-state hint is gated on the setting, so a disabled shortcut never leaves dead instructions on screen.
- **Rebind escape hatch (documented, not built):** power users can already bind any key to the `stenoai://record/start` / `record/stop` deep links (macOS Shortcuts / AutoHotkey). Noted in the README so full in-app rebinding isn't needed here.
- **Docs:** README Features bullet.

## Tests
- New model-free **T2 spec** (`record-hotkey.t2.spec.ts`) asserts the **actual** `globalShortcut.isRegistered()` across pre-seeded-off → launch, toggle-on (registered + persisted), toggle-off (unregistered), using a `STENOAI_E2E_RECORD_ACCEL` low-collision accelerator override so it can't flake on an occupied real accelerator.
- Config accessor unit tests; hero-copy disabled-branch unit test; T1 mock stubs.
- Verified locally: ruff clean (no new violations), full unittest, tsc, vitest 32/32, the T2 spec against a freshly-built backend bundle, CLI round-trips.

Independent cross-family Codex review (FIX-FIRST) applied: gate the Home empty-state hint too, make registration idempotent (no spurious "failed" on repeat-enable), and match the Python getter's malformed-value semantics to Electron's `!== false`.

## One decision for you, @ruzin
I kept the **default ON** for back-compat (the hero UX is built around the shortcut, and the harm is bounded — the key is only held while Steno runs, and it's now disableable + documented, exactly what the reporter asked for). **Flipping the default to OFF for new installs is your call** — the honest counter-argument is that `Cmd+Shift+R` is browser hard-reload and hijacking a common shortcut by default is arguably user-hostile, though the audience skews technical. Happy to flip the default if you'd prefer.

---
## Summary by cubic
Adds a Settings toggle to enable or disable the global record shortcut (Cmd/Ctrl+Shift+R). Registration is gated at startup, applied idempotently, and the UI/docs reflect the setting and when another app owns the shortcut.

- **New Features**
  - Settings → General: “Global record shortcut” toggle (default ON).
  - Config/CLI: `record_hotkey_enabled` plus `get-record-hotkey` / `set-record-hotkey` in `simple_recorder.py`.
  - Electron: reads `config.json` on launch; registers before `createWindow()`; idempotent `applyRecordHotkey()`; IPC `get-record-hotkey`/`set-record-hotkey` live-apply and return `{ enabled, registered }`; uses `CommandOrControl+Shift+R` (overridable via `STENOAI_E2E_RECORD_ACCEL` in tests).
  - UI: Hero and Home “Quick start” hints hide the shortcut when disabled or when `registered === false`; Settings shows “Couldn't register” if another app owns it.
  - Tests: T2 e2e asserts `globalShortcut.isRegistered()` with a low-collision accelerator; unit tests added; mock IPC provides `get-record-hotkey`/`set-record-hotkey`.
  - Docs: README documents the shortcut toggle and macOS-only deep-link rebinding.

<sup>Written for commit c735ce2df790d0562fafb8754b677255520b176f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ruzin:
@Optic00 maybe lets have a changeable shortcuts thing like other apps

Optic00:
Agreed on the end state - a proper shortcut recorder like other apps is the right long-term UX. I'd still land this PR as the minimal slice first: it fixes the concrete complaint in #396 (the system-wide Cmd+Shift+R hijack) with a small diff, and everything in it (the config plumbing, the register/unregister helper, the `registered` state surface) is exactly the foundation a rebind UI sits on - a follow-up adds a `record_hotkey_accel` key on top rather than replacing anything here.

Given the size of the current PR queue I'd keep this one small and track the shortcut recorder as a follow-up - opened #417 for it. Also pushed a fix-up addressing cubic's findings (registration now happens before window creation, hero copy also gates on the actual registration state, README rebind note scoped to macOS).

PR #399, opened 2026-07-23
feat(settings): context-aware Cmd+K settings search (from #349)

ruzin:
## Summary

Adds **context-aware `Cmd+K`** to Settings: while the Settings page is open, the global command palette switches to **"Search settings…"** mode — it filters an index of the app's settings (theme, microphone, AI provider, export, …) and jumps to the tab each one lives on, instead of searching notes.

This is the settings-search piece of @Vassista's **#349**, extracted and **adapted to the current nav-rail Settings design** so it does **not** change how Settings looks.

## What changed

- **`CommandPalette.tsx`** — when the route is `/settings`, the palette searches a `SETTINGS_INDEX` and navigates to `/settings?tab=<id>`. Placeholder/aria/empty-state all switch to settings mode; note search is unchanged elsewhere. The index is **retargeted to the current tabs** (transcription settings live on the AI tab now; added an About entry) — the original PR's index used the pre-v0.6.2 tab layout.
- **`Settings.tsx`** — the visible tab now re-derives from the `?tab=` param **after mount** (previously mount-only), so selecting a result switches tabs even when Settings is already open. Purely behavioral — no layout/visual change.

## Why not just merge #349

#349 is 17 commits stale and its centrepiece is a Settings-UI refactor that predates (and would undo) the v0.6.2 nav-rail redesign. This PR takes only the search, rebuilt on the current design. Cloud transcription, notification, and other pieces of #349 are being extracted as separate PRs.

## Tests

- New T1 `settings-cmdk-search.t1.spec.ts`: settings-mode search + filtering, and the route-reactive tab switch (jump from the AI tab to a General-tab setting). Existing notes-search palette specs still pass.
- `typecheck:renderer` green.

Credit: @Vassista (#349).

---
## Summary by cubic
Adds context-aware Cmd+K to Settings: while on `/settings`, the palette switches to “Search settings…” and jumps to the right tab. Note search stays the same elsewhere.

- **New Features**
  - Cmd+K detects `/settings`, filters a settings index, and opens `/settings?tab=…` for the selected item; placeholder/aria/empty-state switch to settings mode. Note search is unchanged on other routes.
  - Settings now reacts to `?tab=` changes after mount so selections switch tabs even when Settings is already open; adds a T1 spec covering settings-mode search, query trimming, and tab switching.
  - Polish from review: trim queries, guard `aria-activedescendant` when the list shrinks, match real labels (e.g., “Post meeting notifications”), and expand the index (Scheduled meetings, Storage location, dock icon); reset the Templates editor flag when leaving that tab.

<sup>Written for commit dd2faaf5906ddcdd616cf3e9fca580ed3f2f113c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/components/CommandPalette.tsx
+++ b/app/renderer/src/components/CommandPalette.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Search } from 'lucide-react';
 import { useMeetings, LIVE_SUMMARY_PREFIX } from '@/hooks/useMeetings';
 import { searchNotes, snippet } from '@/lib/noteSearch';
-import { navigate } from '@/lib/router';
+import { navigate, useRoute } from '@/lib/router';
 import type { Meeting } from '@/lib/ipc';
 
 interface PaletteContextValue {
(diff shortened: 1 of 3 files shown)


ruzin:
Addressed the cubic review in dd2faaf:

- **P3 (conf 10) — "Post meeting notifications" not findable:** index title had a hyphen the real label doesn't; matched it exactly. ✅
- **P3 (conf 10) — whitespace query returns nothing:** now trim before lowercasing/filtering. ✅
- **P3 — aria-activedescendant can reference a nonexistent option:** restored a guard so it's only set when an option exists at `selected`. ✅
- **P2 — undiscoverable settings:** added **Storage location** and **Scheduled meetings**, and surfaced the **menu-bar/tray** setting in the dock entry's subtitle. ✅
- **P2 — stale `templateEditorOpen`:** reset it whenever the active tab isn't Templates (covers both the nav rail and the ⌘K search path). ✅

**Not changing — P2 (Settings.tsx: navigating to `/settings` or an invalid tab leaves the old tab visible):** this isn't a regression. Before this PR the tab was consumed on mount only, so *no* post-mount hash change switched tabs at all; the new effect strictly improves that by switching on any valid `?tab=`. The settings search always navigates to a valid `?tab=`, so the flagged path isn't reachable through the feature. Deliberately preserving the current tab on a param-less `/settings` avoids clobbering the tab on unrelated internal navigations that land on `/settings` without a param. Extended the T1 spec to cover the trim + real-label cases.

Optic00:
Same timing story as #401: review finalised as it merged. One interaction bug worth a follow-up landed in #405 - after a manual tab switch, a repeated ⌘K settings search can silently no-op because tab state and URL diverge. Repro + fix directions in the issue.

PR #401, opened 2026-07-23
feat(transcription): OpenAI-compatible cloud ASR engine (from #349)

ruzin:
## Summary

Adds an **OpenAI-compatible cloud transcription engine**: alongside Parakeet and Whisper, you can point Steno at any OpenAI-compatible `/audio/transcriptions` endpoint (API URL + model + key). Extracted from @Vassista's #349, adapted to the current design, **with the API-key storage reworked to match the app's safeStorage convention**.

## Security fix vs. the original PR
#349 wrote the ASR API key **plaintext into `config.json`**. This PR does not. The key is handled exactly like the cloud-summariser key:
- Encrypted at rest via `safeStorage` in a `.openai-asr-api-key` file under the user-data dir (`saveOpenAiAsrKey`/`loadOpenAiAsrKey`/`hasOpenAiAsrKey`, with legacy-file migration).
- Decrypted only in the main process and injected into the **transcription subprocess** env as `STENOAI_OAI_API_KEY` (new `getTranscriptionEnv()`, applied at both the batch and live-sidecar spawn sites).
- Python reads it env-only (`get_openai_asr_api_key()` → `os.environ`); there is **no** plaintext setter and no CLI path that takes the key as an argument. `config.json` only ever holds the non-secret URL + model.
- The UI shows the key only as set/not-set (`api_key_set`, overridden by `hasOpenAiAsrKey()`); the value is never returned to the renderer. Clearing removes the file.

## Other hardening (ported)
- `_run_openai_asr` blocks cross-origin redirects (`NoRedirectHandler`) so the Bearer token can't leak to a redirected host; `verbose_json`→`text` fallback.
- Diagnostics redaction for the config channel.

## UI
Re-homed onto the **current** AiTab engine picker (the PR's `TranscriptionTab.tsx` no longer exists) — a "Cloud API" option behind a privacy confirm dialog, plus an API URL / model / key sub-panel using the current `SettingRow` primitives. No settings-layout change.

## Tests
- New `e2e/specs/cloud-asr-config.t2.spec.ts`: sets/gets the non-secret config (persists to `config.json`) and asserts the key flips `api_key_set` true **without** the key value ever landing in `config.json`. Model-free (no network/real ASR).
- `typecheck:renderer`, `ipc-contract` (7/7), and the Python unittest suite green locally. The new T2 needs the freshly-built backend bundle (new CLI commands), which CI builds — it validates there.

Credit: @Vassista (#349).

---
## Summary by cubic
Adds an OpenAI-compatible cloud transcription engine (`openai-asr`) so you can send audio to any `/audio/transcriptions` endpoint, with safe, encrypted key storage and a simple settings UI. Also fixes diarised ordering for text-only endpoints and lets cloud ASR run without local model imports.

- **New Features**
  - New `openai-asr` engine alongside Parakeet and Whisper; Python posts audio to `/audio/transcriptions` with `verbose_json` then `text` fallback and same-origin redirect protection.
  - API key stored via safeStorage, encrypted at rest in `.openai-asr-api-key`, injected as `STENOAI_OAI_API_KEY`; never in `config.json` or argv.
  - Settings > Transcription adds a “Cloud API” option (privacy confirm) with fields for API base URL and model, plus a key input that only shows `api_key_set`.
  - App wiring: env is injected for live sidecar and batch jobs; IPC/preload add `get-openai-asr-config`, `set-openai-asr-config`, `set-openai-asr-key`; diagnostics redact `set-openai-asr-config` args.
  - Tests: new `e2e/specs/cloud-asr-config.t2.spec.ts` verifies config round-trip and encrypted key storage.

- **Bug Fixes**
  - Cloud ASR no longer requires Parakeet/Whisper imports; runs with pure HTTP even if local backends are unavailable.
  - Diarised transcript: correct speaker ordering for text-only endpoints and omit fake `[00:00]` timestamps when no real timing exists.
  - Config hardening: reject blank API URL/model; UI clears now reset to defaults. Additional unit tests cover these cases.

<sup>Written for commit da1fd68f9a11b2986b1dfd8fb21505c70f3bf2dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/diagnostics-filter.js
+++ b/app/diagnostics-filter.js
@@ -75,6 +75,10 @@ const ARGS_ECHO_REDACTORS = {
   'set-remote-ollama-url': redactRest,
   'test-remote-ollama': redactRest,
   'set-cloud-api-url': redactRest,
+  // --api-url may embed credentials or a private org hostname (same class as
+  // set-cloud-api-url). The ASR key never travels via argv — it's env/
+  // safeStorage only — so there's nothing key-shaped to redact here.
+  'set-openai-asr-config': redactRest,
   // device_id + a user-assigned device label (e.g. "Valentin's AirPods") —
   // same PII class as set-user-name.
   'set-microphone': redactRest,
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -271,6 +271,22 @@ function install({ ipcMain }) {
       engine: process.env.STENOAI_E2E_MOCK_ENGINE || 'parakeet',
     }),
 
+    // OpenAI-compatible ASR config. Shape-only for first paint; the real
+    // set/get round-trip + key storage is covered by cloud-asr-config.t2.
+    'get-openai-asr-config': async () => ({
+      success: true,
+      api_url: 'https://api.openai.com/v1',
+      api_key_set: false,
+      model: 'whisper-1',
+    }),
+    'set-openai-asr-config': async () => ({
+      success: true,
+      api_url: 'https://api.openai.com/v1',
+      api_key_set: false,
+      model: 'whisper-1',
+    }),
+    'set-openai-asr-key': async () => ({ success: true, api_key_set: true }),
+
     // Default not-installed keeps most T1 specs on their routes; the pill-dock
     // T1 sets STENOAI_E2E_MOCK_PARAKEET_INSTALLED=1 so App.tsx's first-run
     // setup gate doesn't redirect it to /setup before it can hit Record.
--- a/app/main.js
+++ b/app/main.js
@@ -3709,7 +3709,7 @@ function spawnLiveTranscribe(sessionName) {
     liveTranscribeSessionName = null;
     liveTranscribeStdoutBuf = '';
   }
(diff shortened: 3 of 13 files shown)


Optic00:
Heads-up: my review of this PR was finalised just as it merged, so the findings went into issues instead - #403 carries two I'd treat as release-blocking for the feature (meetings over ~13 min fail against the 25 MB upload limit; unvalidated 200 responses can be mistaken for silence and delete the only audio when Save recordings is off), and #404 covers the privacy-docs contradiction. The safeStorage key handling here is exemplary - the issues are about the request path, not the credential story.

PR #409, opened 2026-07-23
fix(release): make the Discord announcer robust to notes format

ruzin:
## Why

The **v0.6.3** Discord announcement shipped broken — no "A bunch of great features:" section and a generic "Steno v0.6.3 is out." summary (screenshot in thread). Root causes:

1. The workflow fed the announcer `%(contents:body)` — the tag **body**, which excludes line 1 (git's tag *subject* = the summary). So the summary was never seen → generic fallback.
2. The feature parser only matched `- **Name** — desc` bullets. v0.6.3's notes used `### Feature` headers + plain `- description` bullets → **zero** features matched → the whole section vanished.

## Fix (release-infra only)

- **Workflow:** add a `text_full` output (`%(contents)`, subject + body) and point `RELEASE_NOTES` at it so the summary line survives. The GitHub Release body still uses body-only `text`.
- **Announcer:** fall back to synthesising one feature per `### Section` (title = name, bullet text = description) when no bold bullets exist — so an empty features section can never ship again. Also strip a redundant leading `vX.Y.Z —` from the summary and omit the summary rather than printing `Steno vX is out.` when the notes carry none.

## Verified

`DISCORD_DRY_RUN` against both note styles:
- **v0.6.3** (section-header): now renders summary + 4 feature bullets + contributor line.
- **v0.6.0** (bold-bullet): unchanged — bold-bullet features + combined `This release includes work from @… 🙏`.

Note: v0.6.3's post already went out; this fixes future releases (I can manually re-post a corrected v0.6.3 announcement if you want).

---
## Summary by cubic
Fix the Discord release announcer so summaries and feature lists render correctly. We now pass full tag contents and parse mixed note formats (bold bullets + section headers) without double-counting.

- **Bug Fixes**
  - Workflow: add `text_full` from `%(contents)` and point `RELEASE_NOTES` at it to keep the summary line; GitHub Release body stays body-only.
  - Announcer: single-pass parser handles bold bullets and `### Section` + plain bullets together; a section becomes one feature only if it has no bold bullets; skip Upgrade/Contributors.
  - Announcer: strip a leading "vX.Y.Z —"/"Steno vX.Y.Z —" in the summary and omit it entirely if the notes provide none.

<sup>Written for commit 653f07111a16ffa4c4c59ba46720a3bcf2f93b8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/409?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -204,10 +204,19 @@ jobs:
         if [ -z "$MSG" ]; then
           MSG=$(git for-each-ref --format='%(contents:subject)' "refs/tags/${{ github.ref_name }}")
         fi
+        # Full contents (subject + body) for the Discord announcer: the tag
+        # SUBJECT is the release's one-line summary, and %(contents:body) drops
+        # it — which left the announcement with a generic fallback summary. The
+        # GitHub Release body still uses `text` (body only); the announcer uses
+        # `text_full` so its summary line survives.
+        FULL=$(git for-each-ref --format='%(contents)' "refs/tags/${{ github.ref_name }}")
         {
           echo 'text<<STENO_RELEASE_EOF'
           echo "$MSG"
           echo 'STENO_RELEASE_EOF'
+          echo 'text_full<<STENO_RELEASE_FULL_EOF'
+          echo "$FULL"
+          echo 'STENO_RELEASE_FULL_EOF'
         } >> "$GITHUB_OUTPUT"
 
     - name: Download all build artifacts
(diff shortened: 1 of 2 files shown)


ruzin:
Fixed the P2 in the latest commit. Replaced the all-or-nothing fallback with a single pass that handles both shapes **and their mix**: each `- **Name** — desc` bullet becomes a feature (with or without an enclosing section), a section with only plain bullets becomes one section-level feature, and a section that has a bold bullet ignores its plain bullets (no double-counting).

Verified via `DISCORD_DRY_RUN` on three inputs — v0.6.3 (section-header), v0.6.0 (bold-bullet), and a mixed release (`### New` bold bullet + `### Fixes` plain bullets): all three now render every feature.

PR #416, opened 2026-07-24
feat(audio): drop the Screen Recording requirement for system audio, min macOS 14.4 (#116 Part 2)

### LINKED ISSUE #116: Gate auto-detect to macOS 14+; reconsider app's minimum macOS
## Context

The Granola-style auto-detect feature ships with two code paths in `mic-monitor/mic_monitor.swift`:

- **macOS 14+ (Sonoma, Sequoia, Tahoe)**: full per-app attribution via `kAudioHardwarePropertyProcessObjectList` + `kAudioProcessPropertyPID`. Notification reads "Meeting detected — Safari" / "Voice Memos" / etc.
- **macOS 12.3 – 13.x (Monterey, Ventura)**: device-level fallback only knows "mic is in use" without attribution. Notification reads "Meeting detected — an app", and the per-bundle-id debounce collapses to a single sentinel key (so consecutive different-app meetings within the 60s window get silently suppressed).

The fallback is functional but the UX is bad enough that those users would probably ignore the notification anyway. It's also ~15 lines of dead-weight code with edge cases that don't apply to the rest of the codebase.

### 1. Gate auto-detect to macOS 14+ (small, low-risk)

- In `setupAutoMeetingDetector()` (`app/main.js`), check `process.getSystemVersion()`. On `<14`, log and skip spawning mic-monitor entirely.
- Remove the `@available(macOS 14.0, *)` fallback branch from `mic-monitor/mic_monitor.swift` (`defaultInputDevice()`, `deviceIsRunningSomewhere()`, the `FALLBACK_PID` sentinel can all go).
- One-line user-facing note: "Auto-detect requires macOS 14 Sonoma or later."

~30 LOC change. Safe — current 12/13 users get no auto-detect notifications instead of bad ones. Doesn't affect anything else.

### 2. (Separate, larger) Consider raising `LSMinimumSystemVersion`

Currently `LSMinimumSystemVersion: 12.3.0` in `app/package.json`. If raised to **14.4.0**:

- **Drop `NSScreenCaptureUsageDescription` from Info.plist entirely.** That permission is only needed for the pre-14.4 system-audio path (ScreenCaptureKit). On macOS 14.4+ the proper CoreAudio Tap API doesn't require it.
- Privacy / onboarding win — one fewer scary prompt during first-run, and "no screen access required" becomes honest marketing copy.
- Simpler system-audio capture code paths (drop one of the two branches).
- Tradeoff: excludes users on Monterey (12.x), Ventura (13.x), and early Sonoma (14.0–14.3).

Apple's typical support pattern is "current + previous two majors" — as of May 2026 that's Tahoe (26), Sequoia (15), Sonoma (14). macOS 12 is already past that.

Worth pulling PostHog (or whatever telemetry exists) for the actual install-base distribution before committing.

## Recommendation

(1) is uncontroversial — small, low-risk, only affects an opt-in feature. Land as a fast-follow PR after the auto-detect feature ships.

(2) is a product/business call that needs install-base data. Park as discussion under this issue (or split into its own).

Optic00:
Closes #116 (Part 2 - Part 1 was the auto-detect gate in #390). As agreed on Discord: raise the minimum macOS and drop the Screen Recording requirement for system audio.

## What was actually wrong

The CoreAudio Process Tap never needed Screen Recording. electron-audio-loopback's `enable-loopback-audio` handler pulls a REAL screen source via `desktopCapturer.getSources({types:['screen']})` and returns it as the (immediately discarded) video track - that call is the only thing that trips the Screen Recording TCC, and it's why the empirical test with screen recording denied failed ("It doesn't run"). Verified on a real Mac (macOS 26, screen recording explicitly denied): with the handler below, a 440 Hz test tone recorded onto the system channel at -25 dB RMS, and macOS prompted only for Microphone + System Audio Recording.

## Changes

- **Own display-media handler (main.js)**, replacing the package's: answers `getDisplayMedia` with the requesting frame as the throwaway video track plus `audio: 'loopback'`. Two hard constraints discovered while building it:
  - an audio-only response throws (`Video was requested, but no video stream was provided`) because the renderer requests `video: true`, and the one-shot callback is consumed by the throw;
  - Electron invokes the handler without awaiting it, so any throw inside it is an unhandled rejection that crashes the whole app - this is exactly the crash landmine the old `screenPermissionOk` gate worked around. The new handler wraps the callback in try/catch, fixing that class of crash structurally.
  - The override is deliberately platform-neutral: same `audio:'loopback'` semantics on Windows (WASAPI), same landmine removed there. A new `ipc-contract.test.js` guard pins the handler shape (no getSources, try/catch present).
- **Screen-permission machinery removed**: the renderer gate in `useSystemAudioCapture`, the Settings grant/relaunch wizard in GeneralTab, and the `request-screen-recording-permission` / `open-screen-recording-settings` / `relaunch-app` IPC surface (the wizard was relaunch's only consumer).
- **Mic-only notification** now fires when an enabled loopback acquisition genuinely fails (the new failure mode: System Audio Recording denied). macOS-only, as before.
- **`LSMinimumSystemVersion` 12.3 -> 14.4** and `NSScreenCaptureUsageDescription` dropped. README troubleshooting updated.

## Migration / release notes

- **The DMG now requires macOS 14.4+** (CoreAudio Process Taps). Users on 12.3-14.3 keep their installed version but won't be offered a working update path on their OS - worth a line in the release notes.
- Existing users lose nothing else: mic + System Audio Recording are the only permissions used. Already-granted Screen Recording permission simply becomes unused.
- Known limitation (unchanged from today's behavior for other permissions): if a user revokes System Audio Recording later, there's no Settings surface detecting it - they get the "Recording mic-only" notification at record time.

## Verification

- typecheck + lint clean (identical warning count to main), `ipc-contract.test.js` 8/8, GeneralTab vitest 15/15, full T1 e2e 43/43, `recording-lifecycle.t2` 2/2 (drives the real renderer capture path through the new handler).
- On-device end-to-end proof: recording with Screen Recording denied captures system audio (tone at -41 dB RMS on the system channel in the final integrated run; -25 dB RMS in the isolated handler experiment).
- Windows: not run locally; the handler change is platform-neutral and CI's Windows T2 + pipeline lanes cover the recording path.

---
## Summary by cubic
Drops the Screen Recording requirement for system audio by replacing `electron-audio-loopback`’s display-media handler and raises the minimum macOS to 14.4. This simplifies permissions, removes a crash path, and keeps Windows behavior unchanged. Part 2 of #116.

- **New Features**
  - Custom `enable-loopback-audio` handler uses `session.setDisplayMediaRequestHandler` and returns `{ video: request.frame, audio: 'loopback' }`, wrapped in try/catch to avoid unhandled rejections and Screen Recording TCC.
  - Removed Screen Recording UI and IPC (`request/open/relaunch`) and the renderer gate; Settings now only has the toggle. Tests updated.
  - Mic-only notification now triggers only on a real permission denial (getDisplayMedia `NotAllowedError`); other failures fall back silently. Copy updated to match “Screen & System Audio Recording.”
  - Added an IPC contract test to pin the handler shape and ensure no `desktopCapturer.getSources` call.

- **Migration**
  - Minimum macOS is now 14.4+ (`LSMinimumSystemVersion` raised; `NSScreenCaptureUsageDescription` removed). Users on 12.3–14.3 stay on their current version.
  - Only Microphone and System Audio Recording permissions are used; existing Screen Recording grants are ignored.
  - README and troubleshooting updated to reference System Audio Recording only.

<sup>Written for commit c57d6af7c8f0410d9256d2773ad0c58de3cb6352. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ Have questions or suggestions? [Join our Discord](https://discord.gg/DZ6vcQnxxu)
 
 ## Installation
 
-Download the latest release (**Apple Silicon Mac, macOS 12 Monterey or later**):
+Download the latest release (**Apple Silicon Mac, macOS 14.4 or later**):
 
 - [Apple Silicon (M1-M5)](https://github.com/ruzin/stenoai/releases/latest/download/stenoAI-macos-arm64.dmg)
 
@@ -283,9 +283,9 @@ log stream --predicate 'eventMessage CONTAINS "ollama" OR process CONTAINS "Sten
 ### Common Issues
 
 - **Update didn't install**: Auto-updates are applied on next quit. Quit via the **Steno → Quit** menu (not just closing the window), then reopen.
-- **No system audio / no `[Others]` speaker labels**: macOS needs **Screen Recording** permission. Go to **System Settings → Privacy & Security → Screen & System Audio Recording**, enable Steno, and relaunch the app.
+- **No system audio / no `[Others]` speaker labels**: On macOS, allow Steno to record system audio in **System Settings → Privacy & Security → Screen & System Audio Recording**. Screen Recording access is not required.
 - **`stenoai://` deep link doesn't start recording**: Make sure Steno has launched at least once after install so the URL scheme is registered. If it still fails, check the terminal log for `Protocol handler registration` output.
-- **Recording stops early**: Check microphone permissions, Screen Recording permission (if using system audio), and available disk space.
+- **Recording stops early**: Check microphone permission, System Audio Recording permission (if recording system audio), and available disk space.
 - **"Processing failed"**: Usually an Ollama service or model issue — check the terminal logs.
 - **Empty transcripts**: Whisper couldn't detect speech — verify audio input levels.
 - **Slow processing**: Normal for longer recordings; Ollama is CPU-intensive. If summaries are unusually slow, switch to a lighter model in Settings → AI (Gemma 4 E2B is the lightest/fastest).
--- a/app/ipc-contract.test.js
+++ b/app/ipc-contract.test.js
@@ -122,6 +122,24 @@ test('main.js has no duplicate ipcMain registrations (Electron throws on the 2nd
   assert.deepStrictEqual(dups, [], `duplicate ipcMain registration(s): ${dups.join(', ')}`);
 });
 
+test('custom loopback handler avoids screen capture and catches callback failures', () => {
+  const initMainAt = MAIN.indexOf("initMain({ forceCoreAudioTap: process.platform === 'darwin' });");
+  const removePackageHandlerAt = MAIN.indexOf("ipcMain.removeHandler('enable-loopback-audio');");
+  const windowsSetupAt = MAIN.indexOf('// Windows taskbar identity.');
+  assert.ok(initMainAt >= 0, 'expected electron-audio-loopback initMain call');
+  assert.ok(
+    removePackageHandlerAt > initMainAt && removePackageHandlerAt < windowsSetupAt,
+    'custom enable-loopback-audio handler must replace the package handler immediately after initMain',
+  );
+
+  const handler = MAIN.slice(removePackageHandlerAt, windowsSetupAt);
+  assert.match(handler, /ipcMain\.handle\('enable-loopback-audio',\s*\(\)\s*=>\s*{/);
+  assert.match(handler, /setDisplayMediaRequestHandler\(\(request,\s*callback\)\s*=>\s*{/);
+  assert.match(handler, /try\s*{\s*callback\(\{\s*video:\s*request\.frame,\s*audio:\s*'loopback'\s*}\);/s);
+  assert.match(handler, /catch\s*\(err\)\s*{\s*console\.error\(/s);
+  assert.doesNotMatch(handler, /desktopCapturer|getSources/);
+});
+
 test('every renderer-callable channel (preload invoke/send) is registered in main.js', () => {
   const registered = new Set(mainRegistrations());
   const missing = [...preloadInvokeChannels()].filter(
--- a/app/main.js
+++ b/app/main.js
@@ -7012,7 +7012,7 @@ ipcMain.handle('show-system-audio-mic-only-notification', async () => {
     if (!(await notificationsEnabled())) return { success: true, shown: false };
     const notif = new Notification({
       title: 'Recording mic-only',
-      body: 'System audio could not be captured. Check Steno’s System Audio Recording access in System Settings.',
+      body: 'System audio could not be captured. Check Steno’s Screen & System Audio Recording access in System Settings.',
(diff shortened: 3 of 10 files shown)


ruzin:
Maintainer review — **core change is sound, removal is clean** (grep'd the head tree: zero dangling refs to the removed screen-permission symbols; the 14.4 gating via `isCoreAudioTapSupported()` is pre-existing and degrades gracefully).

Before merge:
1. **Rebase** — CONFLICTING, but mechanical: the conflicting `screenPermission` code on main is exactly what this PR deletes (no new consumer introduced).
2. **P3 (cheap)** — the mic-only notification body says 'System Audio Recording access'; the real pane (and this PR's README) is **'Screen & System Audio Recording'**. Match it.
3. **P2 (recommended)** — `useSystemAudioCapture.ts` fires the permission-guidance notification for *any* loopback failure (IPC error, no-audio-track throw), which misleads when permission isn't the cause. Classify to an access-denied result; log the rest.
4. Confirm **Windows T2 + @pipeline** green (the display-media handler override is platform-neutral and changes the Windows path too), and keep the on-device macOS 14.4 QA proof.

Optic00:
Addressed:
1. **Rebased** onto `main` — as you noted, the conflicting `screenPermission` code on main is exactly what this PR deletes; grep confirms zero dangling refs. Now `MERGEABLE`. The `GeneralTab.test.tsx` mock type was reconciled to the real ipc shape `{ supported, osVersion, platform? }` (dropped `screenPermission`, kept #390's `platform` gate). Verified with a local `tsc -p renderer/tsconfig.json` (clean).
2. **P3 copy** — the mic-only toast now says "Screen & System Audio Recording access", matching the real pane + the README.
3. **P2** — `useSystemAudioCapture.ts` now fires the permission-guidance toast **only** on a `getDisplayMedia` `NotAllowedError` (a genuine access denial). IPC failures, an absent audio track, and unsupported hosts fall back to mic-only silently. This also resolves cubic's P3 (an unsupported host no longer reaches the toast, since its failure isn't a `NotAllowedError`).

One assumption worth a sanity check: this classifies macOS System-Audio-Recording denial as surfacing via `NotAllowedError`. If your on-device testing ever shows denial coming back as an empty audio track instead, we'd widen the classification — flagging so it's a conscious choice. Windows T2/@pipeline: confirming green on the re-run.

PR #420, opened 2026-07-24
feat: add configurable locally invoked CLI provider

evgeny-golubev:
## Description

Adds **Locally invoked CLI** as an AI provider for Summarisation & Chat. Users configure a display name, executable with arguments, and timeout; Steno sends each complete prompt over standard input and reads the final answer from standard output.

The adapter supports standard summaries, custom reports, title generation, meeting chat, and global chat without coupling Steno to a named CLI provider. The UI no longer presents Codex or Claude as selectable providers. Documentation instead includes API-key/business-auth examples for Codex CLI, Claude CLI, and Gemini CLI.

Security and configuration behavior:

- The command is saved in Electron Safe Storage, not `config.json`.
- Command configuration is restricted to **Settings → AI → Summarisation & Chat** and requires an explicit native confirmation when changed.
- Command fields are not accepted through `stenoai://` links or the backend CLI; legacy command/provider keys are removed from `config.json`.
- Commands run directly without an implicit shell, receive prompts only through stdin, and start in an empty temporary directory.
- Steno-owned cloud and organisation credentials are removed from the child environment, command arguments are not logged verbatim, stdout is bounded, and the configurable timeout is capped below the outer inactivity watchdog.
- Steno does **not** claim to isolate a configured command. Its tools, filesystem/environment access, network access, and retention behavior remain under the user's control and responsibility. The UI and documentation explicitly call out compliance ownership and prompt-injection risk from meeting content.

The output/error contract is enforced: **Test command** requires a non-empty standard Steno summary with `## Summary`, `## Key Topics`, `## Key Points`, and `## Action Items` sections—not merely exit code 0. CLI failures now propagate as chat errors instead of being stored as successful `[Error: …]` answers.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- [x] Tested on macOS

Additional validation after merging upstream `main`:

- Python CLI/config/Chinese migration suites: 115 passed, 7 skipped
- Electron/Node unit tests: 277 passed
- Renderer unit tests: 122 passed
- Renderer TypeScript typecheck passed
- Production renderer build passed
- Locally invoked CLI T1 Settings E2E passed
- AI provider T2 encrypted-storage/backend E2E passed
- JavaScript syntax, Python compilation, formatting, and diff checks passed

## Additional Notes

The documented examples intentionally use provider-approved API-key or business authentication. Consumer subscription login and bundled subscription rate limits are not presented as a supported product-integration path.

“Locally invoked” describes where the process starts, not where meeting data necessarily remains. A configured command may send transcripts, summaries, or chat context to third parties and may execute enabled tools; the user owns provider-term compliance, participant consent, and resulting data access.


ruzin:
@evgeny-golubev have you checked this against terms of service? also do you want to join our discord https://discord.gg/DZ6vcQnxxu - we've been chatting about it. all the contributors

evgeny-golubev:
@ruzin PR updated. Now it's about a configurable, locally invoked CLI provider instead of a predefined list.

Optic00:
Thanks for the work here, @evgeny-golubev, and for reworking it after ruzin's question - the generic version is clearly the better of the two designs.

We've talked it over and we're going to pass on this one. It isn't a code-quality call, the branch is in decent shape. The blocker is terms of service.

The mechanism is provider-neutral and the docs do show API-key and business-auth examples, so this isn't us assuming bad faith on your part. The problem is that we can't see the difference. The field takes any command, and nothing on our side distinguishes a run authenticated by an API key from one that simply inherits a personal Codex or Claude session that's already sitting in the user's home directory. Realistically, that second case is a good part of why people would ask for this. Those consumer terms don't cover an app piping third-party meeting transcripts through them, and the other people in the meeting never agreed to it either.

That's also the difference to the Cloud API provider, which is the comparison I'd expect you to make. A key entered there was issued by the provider for programmatic use, so the sanctioned path is the only path it can take. A configured executable can carry whatever credential happens to be lying around. Shipping this as a first-class entry in Settings puts our name on a path whose compliance we can neither check nor support, and we'd be the ones the provider comes to.

There's a smaller second reason: ~2.8k added lines and a user-configured executable inside Steno's trust boundary, spread across main.js and the config layer, is a lot of permanent surface for a niche path. That's maintenance and support load we don't want to take on for this.

If the goal is just "use a model that isn't local Ollama", Settings -> AI -> Cloud API already covers it, including a custom endpoint with your own API key.

So I'm closing this. Genuinely though: thanks for the time you put in, and for taking the ToS question seriously when it came up. Please don't read it as a signal about future contributions.

Optic00:
ruzin posted about this in Discord just before I put the above up - same decision, this is just the version with the reasoning written out. And to be explicit, since the two read differently: none of it is a judgement on the quality of your work.

On the direction he sketched there: the version of this we're actually interested in is Steno being driveable by an agent rather than Steno driving a CLI - an agent SDK path or a Steno skill, talking to Steno from the outside. If that appeals to you, it's a far better place to put the effort than this branch, and worth sketching out before anyone writes code.

PR #423, opened 2026-07-24
fix(update): follow redirects in the manual update check

ruzin:
Ahead of the planned org transfer. The manual 'Check for updates' (`checkForUpdates()`) used a raw `https.request` with no redirect handling — after a repo transfer, GitHub's API returns **301** to the new owner/repo, which Node doesn't follow, so the check would silently fail ("Failed to parse update data") for every installed app.

Now follows 301/302/307/308 (max 5 hops, relative-Location safe) and hoists the endpoint into `UPDATE_CHECK_URL`. Background auto-update (electron-updater) already follows redirects; this brings the manual check to parity.

Verified the follow-logic against a local 301→200 harness (follows to the new path and parses). Shipping this **before** the transfer so more installs already tolerate the redirect.

Note: no e2e added — it's an internal network helper not reachable from the T1/T2 harness; verified by inspection + the redirect harness.

---
## Summary by cubic
Fixes the manual "Check for updates" to follow redirects and handle bad headers safely, so it keeps working after a repo rename/transfer. Centralizes the update URL and tightens timeout handling.

- **Bug Fixes**
  - Follows 301/302/307/308 (max 5) and supports relative `Location`; drains response on redirects.
  - Guards redirect URL parsing and invalid update URLs to prevent crashes and return clearer errors.
  - Uses a single 10s timeout budget across the entire redirect chain.

<sup>Written for commit 1004f16637b4a87a9c7bdd04687a62d91f8370f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/423?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -8515,6 +8515,10 @@ const UPDATE_CHECK_URL = 'https://api.github.com/repos/ruzin/stenoai/releases/la
 // check for every already-installed app.
 async function checkForUpdates() {
   const MAX_REDIRECTS = 5;
+  // One 10s budget shared across the whole redirect chain, so a slow chain of
+  // hops can't keep the check hanging for redirects × 10s (each hop used to
+  // reset its own timeout).
+  const deadline = Date.now() + 10000;
 
   function fetchLatest(urlStr, redirectsLeft) {
     return new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


ruzin:
Addressed both:
- **P2 (shared timeout)** — the whole chain now shares a single 10s deadline (`Math.max(1, deadline - Date.now())` per hop), so a slow redirect chain can't stretch to redirects × 10s. ✅
- **P1 (malformed Location)** — wrapped the redirect `new URL()` in try/catch to fail normally. Worth noting: with a valid base URL the WHATWG parser resolves almost any junk `Location` as a relative ref rather than throwing, so the crash is more theoretical than it reads — but the guard is free, so it's in. ✅

PR #425, opened 2026-07-25
feat(update): auto-install downloaded updates when the app is idle

ruzin:
Keeps the existing 'update available' notification, and **additionally** installs a downloaded update + relaunches when the app is genuinely idle — so users never have to click.

## Behaviour
On `update-downloaded` (notification unchanged), a 60s checker + `powerMonitor` resume/lock events install & relaunch **only when every guard passes**:
- setting **on** (`auto_install_when_idle`, **default true**) + update staged, and
- **not** recording / system-audio-capturing, **not** processing (empty queue), **no** live-transcribe sidecar, **no** in-flight AI stream, and
- machine idle **≥ 10 min** (`getSystemIdleTime`).

The recording/processing/live/stream gates make the one catastrophic case — restarting mid-meeting and losing a recording — impossible. `autoInstallOnAppQuit` stays as the safe fallback.

## Safety-critical logic is a pure, exhaustively-tested predicate
`app/update-idle-gate.js` → `isSafeToAutoInstall(state)` (all-AND, defensive type guards; fails safe on any missing/garbled state). `update-idle-gate.test.js`: 12 cases — all-clear, threshold boundary, and every single blocker → false.

## Settings
New **'Install updates automatically'** toggle in General → System (default on), wired end-to-end through the existing settings plumbing.

## New user-visible behaviour ⚠️
This is a **new self-restart** for all users (gated so it can't interrupt work). Flagging for an explicit call on the default-on decision.

## Tests / gates
typecheck green; `test:unit` 203/203 (incl. 12 gate tests); GeneralTab 17/17; Python config round-trip; e2e `settings-roundtrip.t2` extended to assert the toggle persists. Install-and-relaunch itself isn't e2e-tested (it'd kill the process) — covered by the pure-function tests. Inert under E2E.

---
## Summary by cubic
Auto-installs downloaded updates when the app is truly idle and then relaunches, keeping the existing “update available” notification. Adds a default-on “Install updates automatically” setting with a strict, fail-closed safety gate to avoid interruptions.

- **New Features**
  - Installs only when safe: not recording/processing, queue empty, no live transcription or AI streams, no reprocess/report/title jobs, and machine idle ≥10 min; evaluated on download, every 60s, and on resume/lock-screen.
  - New “Install updates automatically” toggle in Settings → General → System (default on); applies immediately and persists in config.
  - Safety gate is a pure `isSafeToAutoInstall` with explicit fail-closed checks and unit tests; re-checks after a brief autosave settle before relaunch; E2E inert; existing on-quit install remains as fallback.

<sup>Written for commit 50e1d406212497a0125ef8c8d458a42eb942332a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/425?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ruzin:
Addressed all of cubic's findings (good catches on the safety path):
- **P0** — gate is now fully **fail-closed**: every activity flag must be explicitly `false` (garbled/undefined blocks). +matrix tests.
- **P1** — the safety gate is **re-checked after the 500ms autosave settle**, immediately before `quitAndInstall`; if work started during the window it aborts and re-arms the checker.
- **P1** — folded **reprocess / report-generation / title-regeneration** into the gate via `activeReprocessJobs` (`otherJobsActive`), so those jobs now block the relaunch.
- P2/P3 — moved the config tests to their own class, made the mock-IPC toggle stateful, fixed the test name.
64 gate tests + 255 unit + contract + typecheck + python config all green.

PR #426, opened 2026-07-25
fix(export): save the note that's on screen as PDF, not always the Standard one

Optic00:
Found while testing a build: with a generated template report open, **"Save
notes as PDF…" exported the Standard structured note** instead of the report on
screen.

## Root cause

`saveNotesPdf` called `buildNotesHtml(noteSections)` and never passed
`activeReport` — the builder's input type had no slot for one. That was a
deliberate scoping decision (*"the branded template is structured-note
shaped"*), but it left the two adjacent menu actions disagreeing: #318 taught
**Copy notes** to follow the screen, and the PDF path was never given the same
treatment.

## Fix

A report is free-form markdown, so it is serialised with the **same renderer the
detail view uses** (`react-markdown` via `renderToStaticMarkup`) and dropped into
the branded shell. Reusing the view's renderer is the point: a second markdown
implementation inside `notesPdf.ts` would drift from what the user is looking at.

- `buildNotesHtml` takes an optional pre-rendered report; when present it
  **replaces** the Standard sections, mirroring `buildNotesCopyText`'s
  short-circuit.
- The branded shell moved into `renderDocument()` so both paths share chrome,
  fonts and layout and cannot drift.
- CSS for the element types markdown brings that the structured note never
  emits: body headings, ordered lists, emphasis, code, quotes, tables. Without
  it they inherited the structured note's list rules and lost their numbering.
- **Also fixed:** `canExportNotesPdf` now counts an open report on its own. A
  transcript-only note (auto-summarise off) with a generated report had the
  action disabled outright — the report could not be exported at all.

The report HTML is injected unescaped by necessity. `notesPdf.ts` documents why
that is safe: `react-markdown` emits no raw HTML from its input unless
`rehype-raw` is enabled (it is not), and the document's CSP
(`default-src 'none'`) remains the second line of defence.

## Testing

- Four unit cases in `notesPdf.test.ts` — **three failed before the fix**, so the
  bug was pinned by tests before any code changed.
- A `notes-pdf-export.t1` case driving the real UI: open the seeded report,
  export, then assert the written file carries the **rendered** report, the
  template name, no Standard sections, and no leaked `<think>` reasoning.
- That e2e case was verified to fail without the fix.

Typecheck clean, lint 0 errors, unit 255 node:test + 124 vitest, T1 suite 50/50.

---
## Summary by cubic
Fixes “Save notes as PDF…” exporting the Standard note when a template report is open. The PDF now matches what’s on screen and can export generated reports even when the structured note is empty.

- **Bug Fixes**
  - Report markdown is serialized with the same renderer as the detail view (`react-markdown` via `renderToStaticMarkup`) so the PDF matches the UI.
  - `buildNotesHtml` accepts an optional pre-rendered report and replaces the Standard sections when present.
  - Moved the branded shell to shared `renderDocument()` and added CSS for markdown elements (headings, ordered lists, code, tables) to keep numbering and layout correct.
  - `canExportNotesPdf` now treats an open report as exportable on its own, enabling export for transcript-only notes.

<sup>Written for commit 6ce64dee8feb413aa43afa1617e7a254d76a1466. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/renderer/src/lib/notesPdf.test.ts
+++ b/app/renderer/src/lib/notesPdf.test.ts
@@ -125,3 +125,61 @@ describe('escapeHtml', () => {
     expect(escapeHtml(`&<>"'`)).toBe('&amp;&lt;&gt;&quot;&#39;');
   });
 });
+
+/**
+ * A generated template report is free-form markdown, not the Standard note's
+ * structured fields — so when one is on screen the PDF must carry THAT, the way
+ * "Copy notes" already does (#318). The caller serialises the markdown with the
+ * same renderer the detail view uses and hands the HTML in; this builder only
+ * places it inside the branded shell.
+ */
+describe('buildNotesHtml with an open template report', () => {
+  const report = {
+    templateName: 'Shareable summary',
+    contentHtml: '<h2>Outcome</h2>\n<p>We ship on <strong>Friday</strong>.</p>',
+  };
+
+  test('renders the report instead of the Standard sections', () => {
+    const html = buildNotesHtml(full, report);
+    // The report's content is present…
+    expect(html).toContain('<p>We ship on <strong>Friday</strong>.</p>');
+    // …labelled with the template it came from…
+    expect(html).toContain('>Shareable summary</h2>');
+    // …and the Standard note's sections are NOT along for the ride.
+    expect(html).not.toContain('>Key Topics</h2>');
+    expect(html).not.toContain('Ship v2 in July');
+    expect(html).not.toContain('Ben: draft the announcement');
+  });
+
+  test('keeps the brand chrome, title and meta line', () => {
+    const html = buildNotesHtml(full, report);
+    expect(html.startsWith('<!doctype html>')).toBe(true);
+    expect(html).toContain('<h1>Weekly sync</h1>');
+    expect(html).toContain('Mon, Jun 23, 2026 · 45m');
+    expect(html).toContain('@font-face');
+  });
+
+  test('renders a report even when the Standard note is empty', () => {
+    // A transcript-only note (auto-summarise off) can still have a generated
+    // report — the export must not be gated on the Standard sections.
+    const empty: NotesPdfInput = {
+      name: 'Transcript only',
+      discussionAreas: [],
+      keyPoints: [],
+      actionItems: [],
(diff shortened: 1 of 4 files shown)


Optic00:
### Merge order note: this overlaps #448

Flagging an overlap rather than proposing a decision - the order is yours.

This PR and #448 (Share menu) touch the same four files, and two of the overlaps are semantic rather than textual:

**1. `canExportNotesPdf`.** This PR makes it report-aware:

```
activeReport ? Boolean(stripReasoning(activeReport.content).trim()) : hasNotesContent(noteSections)
```

#448 adds two new entries (`Save notes as .md…`, `Share notes as .md…`) that hang off that same predicate, and its body flags the resulting gap as a known one: their markdown builder carries an open report, so on a transcript-only note with a report open, `Copy notes` works while those two entries are disabled. **This PR fixes that gap for free.** If it lands first, that caveat on #448 simply disappears instead of being patched separately.

**2. `buildNotesHtml(noteSections, reportForPdf)`.** #448 introduces a second call site for the branded HTML - the `Share notes as PDF…` entry, which hands the same document to the macOS share sheet. A naive merge would leave that call site passing the Standard note while `Save notes as PDF…` passes the report, so saving and sharing the same note would produce different PDFs. It is a one-line fix on the #448 side, but only if someone knows to look.

**3. `e2e/specs/notes-pdf-export.t1.spec.ts`.** Both PRs rewrite it. #448 moved the PDF action out of the `…` popover into the Share menu, so every selector in that spec changed there.

**Suggested order: this one first, then rebase #448.** It is the smaller and older change, it resolves #448's open caveat rather than colliding with it, and the reverse order would mean pulling this PR through #448's menu restructuring. Happy to do the #448 rebase as soon as this lands.

PR #431, opened 2026-07-26
chore(release): v0.6.5

ruzin:
Release prep for **v0.6.5**: README "What's New" (rolling 4), `docs/changelog.mdx` Update block, and version bump 0.6.4 → 0.6.5.

Headline changes shipping (since v0.6.4):
- System audio without the Screen Recording permission — **now requires macOS 14.4+** (#416)
- Automatic idle install of downloaded updates (#425)
- Re-transcribe an existing recording (#392); undo / soft-delete for note deletion (#391)
- Chinese Simplified/Traditional transcription on the Whisper engine (#102)
- Low-memory model warning (#389); Keychain prompt fix (#363); redirect-safe update check (#423); nav security hardening (#387)

Docs-and-version-only; no runtime code changes in this PR.

---
## Summary by cubic
Release prep for v0.6.5: bump version and update README “What’s New” and `docs/changelog.mdx`. Docs-only; no runtime code changes in this PR.

- **New Features**
  - System audio without Screen Recording; requires macOS 14.4+.
  - Automatic idle install of updates; never during a recording; can be turned off in Settings.
  - Re-transcribe existing recordings; undo/soft-delete note deletions.
  - Chinese transcription (Simplified and Traditional) on the Whisper engine.
  - Low-memory warning for large local models.

- **Bug Fixes**
  - Keychain “Safe Storage” prompt no longer repeats on launch.
  - Auto-update checks: follow redirects and avoid offering updates on macOS <14.4.
  - Settings links land on General.
  - Removed stray AI reasoning tags in notes.
  - Fewer false “meeting detected” prompts.

<sup>Written for commit d45ffcbb84f7d09820fcaa1ed9f4f9d7e7d1b9b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/431?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,10 @@
 If you're looking for a hosted desktop recording API, consider checking out [Recall.ai](https://www.recall.ai/product/desktop-recording-sdk?utm_source=github&utm_medium=sponsorship&utm_campaign=ruzin-stenoai), an API that records Zoom, Google Meet, Microsoft Teams, in-person meetings, and more.
 
 ## 📢 What's New
-- **2026-07-24** 🔎 Settings search — press ⌘K on the Settings screen to search every setting and jump straight to the right tab.
-- **2026-07-24** 🔔 Cleaner notifications — meeting and note alerts now show as Steno's own in-app toast instead of native OS notifications.
-- **2026-07-24** ✨ Letter avatars — notes in the Previous list show their initial instead of a generic file icon.
-- **2026-07-24** 📅 Connected calendar — Calendar settings now show the account you've connected.
+- **2026-07-26** 🎙️ System audio without Screen Recording — record both sides of a call with no Screen Recording permission. Now requires macOS 14.4 or later.
+- **2026-07-26** ⬇️ Automatic updates — Steno installs a downloaded update while you're idle and relaunches, never mid-recording. Turn it off in Settings to keep the manual prompt.
+- **2026-07-26** ↩️ Undo delete + re-transcribe — restore a note right after deleting it, and re-transcribe an existing recording from its detail view.
+- **2026-07-26** 🌏 Chinese transcription — pick Simplified or Traditional Chinese when the Whisper engine is selected.
 
 
 ## Features
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stenoai",
-  "version": "0.6.4",
+  "version": "0.6.5",
   "description": "AI powered intelligence layer for confidential workflows",
   "main": "main.js",
   "homepage": "https://github.com/ruzin/stenoai",
--- a/docs/changelog.mdx
+++ b/docs/changelog.mdx
@@ -21,6 +21,7 @@ description: "What's new in Steno -- recent releases and notable changes."
 - Opening Settings from a bare link now lands on General instead of stranding you on a stale tab.
 - The AI's internal reasoning tags no longer leak into your notes in a few remaining cases.
 - Fewer false "meeting detected" prompts.
+- Macs below macOS 14.4 are no longer offered an update they can't install; a manual check explains that a newer macOS is required instead.
 </Update>
 
 <Update label="v0.6.4" description="July 24, 2026">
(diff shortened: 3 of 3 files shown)


Optic00 (dismissed):
Checked the changelog and README copy against `git log v0.6.4..main`, and every
entry maps to a merged PR in the range: #416, #425, #392, #391, #102, #389,
#421, #363, #423, #415, #370, #390. Nothing in the range is misdescribed, the
"What's New" marquee is correctly trimmed to four, and the version bump is
consistent. The copy itself reads well.

Two things that are not in this PR but gate the tag it leads to.

**1. Auto-update will push this build onto macOS 12/13 installs (#432).**

This is the release-blocking one. #416 raised `LSMinimumSystemVersion` to 14.4,
but that floor is enforced by Launch Services at *startup* — nothing in the
update path checks it. A v0.6.4 install on macOS 12/13 will download v0.6.5,
install it during idle (#425), and then have an app that no longer opens.

It cannot be fixed inside the app: a build only ever runs on an OS that already
meets its own floor, so an in-app guard could never fire, and the machines at
risk are running the previous build. The manifest is what reaches them —
electron-updater checks `minimumSystemVersion` in `latest-mac.yml` before it
reports an update or downloads anything, and v0.6.3/v0.6.4 both ship
`electron-updater ^6.8.3`. Fix in #436.

One trap worth knowing before anyone patches this by hand: electron-updater
compares that field against `os.release()`, which on macOS is the **Darwin**
version. `minimumSystemVersion: 14.4.0` is a silent no-op, because macOS 12
reports Darwin `21.6.0`. It needs `23.4.0`. Evidence is in #432; #436 has the
workflow change and the unit-tested translation.

**2. The release body template still says macOS 12 (#434).**

`.github/workflows/build-release.yml` has `- macOS 12 (Monterey) or later` in
its `### Requirements` block, so every release page from this tag onward would
restate the old floor next to the download. #434 fixes that plus the website
FAQ, which #430 left behind.

**Optional, same area:** the template already carries an "Intel Mac users"
paragraph telling that group which version to stay on. Once #432 lands, macOS
12/13 users are in exactly the same position — stranded on v0.6.4 with
auto-update correctly refusing to move them — and an equivalent line would
answer the support question before it is asked.

**Nit, take it or leave it:** the PR body lists "nav security hardening (#387)"
as shipping, but there is no changelog entry for it. Fine to leave out
deliberately; just flagging the mismatch between body and changelog.

Optic00:
Suggested cut line for this tag, in the interest of shipping it soon and safely: **#431 + #436 + #434, and nothing else.**

- **#436** is the one hard blocker. Verified end to end now, not just by reading the library: a real electron-updater 6.8.3 cycle at version `0.6.4` against the manifest the workflow produces refuses the update on macOS 12/13/14.3 and offers it on 14.4+, while the manifest as `main` produces it today offers it to macOS 12 — [evidence](https://github.com/ruzin/stenoai/pull/436#issuecomment-5083750120), including why the obvious `14.4.0` value is inert.
- **#434** is docs, but it fixes the `### Requirements` block in the release template, which prints "macOS 12 (Monterey) or later" on the release page this tag creates.

Everything else we have open — #428, #422, #426, #398, #368 — is green and reviewed but not required for a safe v0.6.5. They add change surface to a release whose whole point is a raised OS floor, and they lose nothing by going into 0.6.6 right after. I also have an About-tab copy fix ready locally and am deliberately holding it back for the same reason: it touches the update error path that #435 and #436 just changed twice in one day.

No opinion on the release timing itself beyond that — the branch gate (`release/v0.6.5`) before the tag is the remaining step on the checklist.

Optic00 (approved):
Re-approving after `d45ffcb` (my earlier approval was dismissed by that commit). Diff since then is the single changelog line, and it reads correctly.

One dependency worth stating explicitly: "Macs below macOS 14.4 are no longer offered an update they can't install" is only true for the machines at risk once #436 is in. On its own, #435's runtime gate ships inside the build that must not be installed, while the affected clients are running v0.6.4 with no gate at all. So this line and #436 should land in the same tag.

PR #435, opened 2026-07-26
fix(update): don't auto-update onto macOS below the 14.4 launch floor (#432)

### LINKED ISSUE #432: Auto-update can install the macOS 14.4 build onto macOS 12/13, which then won't launch
## Summary

`main` now raises the app-wide floor to macOS 14.4 (#416), but nothing in the update path gates on the OS version. When the next release is tagged, an existing install on macOS 12/13 will download and install a build that Launch Services then refuses to start.

This has **not shipped yet** — `git tag --contains cd085ed` (the #416 commit) returns no tag — so it is still fixable before the tag rather than after. That is the reason for filing it now.

## Why the gate is missing

The floor is set through `mac.extendInfo`, not through electron-builder's own `mac.minimumSystemVersion` option:

```jsonc
// app/package.json:166-169
"extendInfo": {
  ...
  "LSMinimumSystemVersion": "14.4.0"
}
```

`extendInfo` writes the key into `Info.plist`, which is what stops the app from launching. It does not reach the generated `latest-mac.yml`, so the update manifest carries no `minimumSystemVersion` for electron-updater to act on.

There is no manual check either — no `minimumSystemVersion`, no `isUpdateSupported`, and no `process.getSystemVersion()` comparison anywhere in the update path in `app/main.js`. What is there:

- `app/main.js:6250` — `autoUpdater.autoDownload = true`
- `app/main.js:6251` — `autoUpdater.autoInstallOnAppQuit = true`
- `app/main.js:6223` — `autoUpdater.quitAndInstall(true, true)`

`getDownloadUrl()` (`app/main.js:9367`) gates on `process.arch` for the Intel/arm64 split, but there is no equivalent for the OS version.

#425 ("auto-install downloaded updates when the app is idle", `35952b0`) is in the same untagged range, so the install no longer waits for the user to quit.

## Impact

A user on macOS 12 or 13 running a current release gets the update installed for them, and afterwards the app will not start. Recovery means manually downloading and reinstalling an older DMG — not something most users will work out on their own. The size of the affected group is whatever share of installs is still below 14.4; telemetry would tell us.

## Proposed fix

Set `mac.minimumSystemVersion` in the electron-builder config so the floor is written into `latest-mac.yml`, where electron-updater can honour it, in addition to the existing `Info.plist` key.

**Caveat, stated honestly:** I have not verified this against the electron-updater 6.8.3 source — I could not run the check locally. The gap itself is verified from this repo; the specific mechanism is a proposal and should be confirmed before it is relied on. If the manifest field turns out not to be honoured in 6.8.3, an explicit `process.getSystemVersion()` guard before `checkForUpdates()` would achieve the same thing.

## Suggested verification

Whichever fix is chosen, it is worth confirming end to end rather than by inspection:

1. Build a release candidate and check that the generated `latest-mac.yml` actually contains `minimumSystemVersion`.
2. On a macOS 13 VM with a pre-14.4 build installed, point the updater at that manifest and confirm it reports no update rather than downloading one.

## Context

Found while reviewing #430 (docs alignment for the same 14.4 floor). Unrelated to that PR, which is docs-only — filing separately.

ruzin:
Closes #432.

## Problem
#416 raised the app-wide floor to macOS 14.4 (`Info.plist` `LSMinimumSystemVersion`), but the update path had **no OS gate**. An existing install on macOS 12/13 would auto-download and — with idle auto-install (#425) — **auto-install a build that Launch Services then refuses to start**. An active brick, not passive abandonment.

## Fix — two layers
1. **Manifest** — `build.mac.minimumSystemVersion` so electron-builder writes `minimumSystemVersion` into `latest-mac.yml`, which electron-updater 6.8.3's `isUpdateSupported` already honours before downloading (verified in `node_modules/electron-updater/out/AppUpdater.js:348,365-375`).
2. **Runtime guard** — new pure `app/update-os-gate.js` (`isOSUpdateEligible`) gates `setupAutoUpdater()` (early return → nothing downloads → `update-downloaded` never fires → the #425 idle checker never arms) and the manual `check-for-updates` kick. Minor-version aware (blocks 14.0–14.3, not just <14), **fail-open** on a parse hiccup (manifest is the backstop). Darwin-only — Windows untouched.

Under-floor Macs also get an honest **manual-check** experience: the About tab now says "vX requires a newer version of macOS" and suppresses the "View release" download nudge (`osUpdateEligible` threaded through IPC).

## Invariants preserved
`process.arch` Intel/arm64 gate, #423 redirect logic, and the #425 idle gate are all untouched; the new guard sits upstream. The floor is single-sourced (`MIN_MACOS_FOR_AUTOUPDATE`) and a unit test pins `minimumSystemVersion` + `LSMinimumSystemVersion` to it so they can't drift.

## Tests
- `app/update-os-gate.test.js` — 18 cases (boundary/garbled/non-string/fail-open + the config-floor consistency assertion).
- `settings-about.t1` — new spec: under-floor Mac shows the "requires a newer macOS" notice and no "View release".
- Gates green locally: 273 node:test, 120 vitest, renderer typecheck.

## Verification honesty
The pure gate is runtime-exercised + unit-tested, and the About-tab notice is driven by a real-Electron T1. The full end-to-end — a signed build on a **macOS 13 VM** confirming electron-updater reports no-update from `latest-mac.yml` — can't run in CI/here; the runtime guard is the correct-by-inspection second layer. Flagging that manual VM check as a follow-up.

Reviewed in-context (senior QA + senior Eng lenses): no critical/high; medium/low folded in (single-source floor, wired the notice, hardened install-update).

---
## Summary by cubic
Prevents auto-update from downloading or installing on macOS versions below 14.4, avoiding installs that won’t launch. Adds a manifest floor and a runtime gate, and clarifies the About tab when an update requires a newer macOS.

- **Bug Fixes**
  - Set `build.mac.minimumSystemVersion: "14.4.0"` so `electron-builder` writes the floor into `latest-mac.yml` and `electron-updater` blocks before download.
  - Added `app/update-os-gate.js` with `isOSUpdateEligible` and `MIN_MACOS_FOR_AUTOUPDATE`; gates `setupAutoUpdater`, `install-update`, and manual `check-for-updates` on macOS, fail-open on parse; other platforms unchanged.
  - Threaded `osUpdateEligible` through IPC; About tab shows “vX requires a newer version of macOS” and hides “View release” when blocked.
  - Single-sourced the floor and added tests (unit coverage for the gate + an About-tab e2e).

<sup>Written for commit 163fc386c0bd62d27418289fc48948540eaf743a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -588,14 +588,33 @@ function install({ ipcMain }) {
     // Read-only display poll for the About tab's "Check for Updates" button
     // (settings-about.t1). Fully hermetic — no real GitHub call under mock
     // IPC, so this is the only source of truth for that flow in T1.
-    'check-for-updates': {
-      success: true,
-      updateAvailable: false,
-      currentVersion: '0.0.0-e2e',
-      latestVersion: '0.0.0-e2e',
-      releaseUrl: '',
-      releaseName: '',
-      downloadUrl: null,
+    'check-for-updates': () => {
+      // STENOAI_E2E_SEED_UPDATE_BLOCKED_OS=1 simulates an under-floor Mac: an
+      // update exists on GitHub but this OS is below the 14.4 launch floor, so
+      // osUpdateEligible is false and the About tab must explain that instead of
+      // offering a broken install/download nudge (#432, settings-about.t1).
+      if (process.env.STENOAI_E2E_SEED_UPDATE_BLOCKED_OS === '1') {
+        return {
+          success: true,
+          updateAvailable: true,
+          currentVersion: '0.0.0-e2e',
+          latestVersion: '9.9.9',
+          releaseUrl: 'https://github.com/ruzin/stenoai/releases/latest',
+          releaseName: 'Version 9.9.9',
+          downloadUrl: null,
+          osUpdateEligible: false,
+        };
+      }
+      return {
+        success: true,
+        updateAvailable: false,
+        currentVersion: '0.0.0-e2e',
+        latestVersion: '0.0.0-e2e',
+        releaseUrl: '',
+        releaseName: '',
+        downloadUrl: null,
+        osUpdateEligible: true,
+      };
     },
     // AboutTab's mount-time re-seed effect. Without an explicit stub, both
     // fields fall through to the permissive default (undefined, not null),
--- a/app/main.js
+++ b/app/main.js
@@ -57,7 +57,7 @@ const { createTeardownRegistry } = require('./teardown');
 const { registerFoldersIpc } = require('./folders-ipc');
 const { registerSettingsIpc } = require('./settings-ipc');
 const { isSafeToAutoInstall } = require('./update-idle-gate');
-const { isOSUpdateEligible } = require('./update-os-gate');
+const { isOSUpdateEligible, MIN_MACOS_FOR_AUTOUPDATE } = require('./update-os-gate');
 const processingLog = require('./processing-log');
 const { isMeetingApp, allowsDeviceLevelFallback, isMacos14Plus } = require('./meeting-detect');
 const { sweepOrphanedLiveSnapshots } = require('./live-snapshot-sweep');
@@ -6237,22 +6237,20 @@ function startIdleAutoInstallChecker() {
   if (idleAutoInstallInterval.unref) idleAutoInstallInterval.unref();
 }
 
-// Auto-update OS floor. MUST stay in lockstep with the Info.plist floor
-// (`build.mac.minimumSystemVersion` / `extendInfo.LSMinimumSystemVersion` in
-// app/package.json). The DMG won't LAUNCH below this, so we must never
-// download or install it onto an under-floor Mac (#432).
-const MIN_MACOS_FOR_AUTOUPDATE = '14.4.0';
+// Current OS version for the update gate, '' if it can't be read (→ fail open).
+function currentSystemVersionSafe() {
+  try { return process.getSystemVersion(); } catch (_) { return ''; }
+}
 
 // Whether electron-updater may run on this OS. See update-os-gate.js — darwin
-// below the floor is blocked (so nothing downloads and the idle installer has
-// nothing to apply); non-darwin and unparseable versions are eligible, with the
-// manifest `minimumSystemVersion` in latest-mac.yml as the authoritative backstop.
+// below the 14.4 floor (MIN_MACOS_FOR_AUTOUPDATE) is blocked (so nothing
+// downloads and the idle installer has nothing to apply); non-darwin and
+// unparseable versions are eligible, with the manifest `minimumSystemVersion`
+// in latest-mac.yml as the authoritative backstop.
 function autoUpdateOSEligible() {
-  let osVersion = '';
-  try { osVersion = process.getSystemVersion(); } catch (_) {}
   return isOSUpdateEligible({
     platform: process.platform,
-    osVersion,
+    osVersion: currentSystemVersionSafe(),
     minVersion: MIN_MACOS_FOR_AUTOUPDATE,
   });
 }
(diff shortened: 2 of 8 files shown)


Optic00:
**Heads-up before the tag: the manifest half of this does not take effect, so
the installs #432 is about are still unprotected.**

I built it locally to check rather than argue from the docs. electron-builder
26.8.1, `mac.minimumSystemVersion: 14.4.0` set exactly as merged here, `zip`
target so the manifest is produced:

```
$ /usr/libexec/PlistBuddy -c "Print :LSMinimumSystemVersion" dist/mac-arm64/Steno.app/Contents/Info.plist
14.4.0

$ cat dist/latest-mac.yml
version: 0.6.5
files:
  - url: Steno-0.6.5-arm64-mac.zip
    sha512: E9ikA5ED7PAb6mvhhmiTr96GHsNdnOmd6sHLoPE+NQd7ASsDu6FTYW0NuOLz3jkFeVwQCgIYLcGbsAcN9FCtVQ==
    size: 419573490
path: Steno-0.6.5-arm64-mac.zip
sha512: E9ikA5ED7PAb6mvhhmiTr96GHsNdnOmd6sHLoPE+NQd7ASsDu6FTYW0NuOLz3jkFeVwQCgIYLcGbsAcN9FCtVQ==
releaseDate: '2026-07-26T12:24:52.849Z'

$ grep -c minimumSystemVersion dist/latest-mac.yml
0
```

The option only reaches the plist: `app-builder-lib/out/macPackager.js:462-464`
assigns it to `appPlist.LSMinimumSystemVersion` and nothing else reads it. The
manifest is built from `releaseInfo` (`out/publish/updateInfoBuilder.js:132-152`),
and `ReleaseInfo` in `app-builder-lib/scheme.json` is `additionalProperties: false`
with no such property, so there is no config path that produces it.

**And there is a second trap underneath it.** electron-updater compares that
field against `os.release()` — the *Darwin kernel* version on macOS, not the
product version (`AppUpdater.js:364-378`). So even once the key is written,
`14.4.0` is a no-op: macOS 12 reports Darwin `21.6.0`, which is not lower.
Driving the real 6.8.3 function with a stubbed `os.release()`:

| Darwin (macOS) | `14.4.0` | `23.4.0` |
|---|---|---|
| 21.6.0 (macOS 12) | update offered | **blocked** |
| 22.6.0 (macOS 13) | update offered | **blocked** |
| 23.3.0 (macOS 14.3) | update offered | **blocked** |
| 23.4.0 (macOS 14.4) | update offered | update offered |

**Why the runtime gate cannot cover for it.** `isOSUpdateEligible` only ever runs
inside a build that already meets the floor — Launch Services would not have
started it otherwise. The machines that get bricked are running v0.6.4, which
has no gate and will never execute this code before deciding to update. That is
also why the documented fail-open ("the manifest gate still catches a genuinely
old OS") is currently load-bearing on a layer that isn't there.

And nothing downstream catches it either: Squirrel.Mac verifies the signature
and then swaps the bundle without reading `LSMinimumSystemVersion`
([SQRLInstaller.m](https://github.com/Squirrel/Squirrel.Mac/blob/0e5d146ba13101a1302d59ea6e6e0b3cace4ae38/Squirrel/SQRLInstaller.m#L479-L519)),
so the first thing that enforces the floor is the relaunch, which fails. The
About-tab change is v0.6.5's manual UI and never runs in the automatic path.

#436 completes it: the release workflow writes `minimumSystemVersion` into
`latest-mac.yml` itself, derived from your `MIN_MACOS_FOR_AUTOUPDATE` through a
`darwinFloorForMacos()` added to `update-os-gate.js`, so the two floors cannot
drift. It runs in the build job where `npm ci` has already run, so the patched
manifest is re-parsed with js-yaml to prove the key landed at the root and
nothing else changed; the release job refuses to publish a manifest without it.
I also corrected the comments in `update-os-gate.js` that describe the manifest
as the authoritative backstop.

Worth reopening #432 until that lands — as it stands, tagging v0.6.5 still
pushes the 14.4-only build onto macOS 12/13.

PR #436, opened 2026-07-26
fix(release): write the macOS floor into latest-mac.yml, in Darwin form (#432)

Optic00:
## What

Completes #432. #435 landed the runtime gate and set
`build.mac.minimumSystemVersion`, but the manifest half of it does not take
effect, so the installs #432 is about are still unprotected and tagging v0.6.5
would still push the 14.4-only build onto macOS 12/13.

Rebased onto `main`, and folded into #435's module rather than sitting beside
it: `darwinFloorForMacos()` now lives in `app/update-os-gate.js` and derives
from `MIN_MACOS_FOR_AUTOUPDATE`, so the runtime floor and the manifest floor
cannot drift.

## Why the manifest half doesn't work today

**1. electron-builder never writes the field.** `mac.minimumSystemVersion` only
reaches Info.plist (`app-builder-lib/out/macPackager.js:462-464`); the manifest
is built from `releaseInfo` (`out/publish/updateInfoBuilder.js:132-152`) and
`ReleaseInfo` in `scheme.json` is `additionalProperties: false`. Verified by
building it — electron-builder 26.8.1, `mac.minimumSystemVersion: 14.4.0` set,
`zip` target:

```
Info.plist  → LSMinimumSystemVersion = 14.4.0
latest-mac.yml → grep -c minimumSystemVersion = 0
```

**2. The value has to be the Darwin version.** electron-updater compares
`minimumSystemVersion` against `os.release()`, which on macOS is the kernel
version (`AppUpdater.js:364-378`). So `14.4.0` is inert even once written —
macOS 12 reports Darwin `21.6.0`, which is not lower. Driving the real 6.8.3
function with a stubbed `os.release()` against the patched manifest this PR
produces:

| Darwin (macOS) | `14.4.0` | `23.4.0` (this PR) |
|---|---|---|
| 21.6.0 (macOS 12) | update offered | **blocked** |
| 22.6.0 (macOS 13) | update offered | **blocked** |
| 23.3.0 (macOS 14.3) | update offered | **blocked** |
| 23.4.0 (macOS 14.4) | update offered | update offered |
| 25.5.0 (macOS 26.5) | update offered | update offered |

**3. The runtime gate cannot cover for it.** `isOSUpdateEligible` only ever runs
inside a build that already meets the floor — Launch Services would not have
started it otherwise. The machines that get bricked are running v0.6.4, which
has no gate. That is also why the gate's documented fail-open needs the manifest
to exist: the comments in `update-os-gate.js` are corrected here to say which
layer protects whom.

## Changes

- **`app/update-os-gate.js`** — adds `darwinFloorForMacos()` (a table, not
  `major + 9`: that breaks at macOS 26 = Darwin 25; an unmapped major throws so
  a floor bump fails the release rather than shipping a floor that gates
  nobody), reusing the existing strict `parseVersion`. Corrects the header
  comments about the manifest layer.
- **`.github/workflows/build-release.yml`**
  - `build-macos`: writes `minimumSystemVersion` into `dist/latest-mac.yml`
    after the DMG build, where `npm ci` has run — so the patched manifest is
    **re-parsed with js-yaml** to prove the key landed at the root with the
    right value and that `version` / `files` / `path` / `sha512` /
    `releaseDate` are untouched. Anything unexpected throws.
  - `release`: re-derives the expected value and refuses to publish unless the
    manifest carries exactly `minimumSystemVersion: 23.4.0`, so a removed or
    half-applied step cannot ship an ungated feed.
- **`app/update-os-gate.test.js`** — 4 tests on top of #435's 18, pinning
  `MIN_MACOS_FOR_AUTOUPDATE` → `23.4.0` end to end.

## Verification

- `npm run test:unit`: 277 node:test + 120 vitest, all passing. Typecheck clean.
- Both workflow steps extracted from the YAML and executed against the manifest
  a real `electron-builder` run had just produced, plus the published v0.6.4
  one. Failure branches exercised: key already present, YAML document marker,
  unmapped macOS major, manifest without a trailing newline, ungated manifest
  reaching the release job, and a manifest carrying the *product* floor instead
  of the Darwin one — each fails the job instead of publishing.
- Reviewed by Codex (read-only, cross-family) over four rounds, the last one
  adversarial on the two load-bearing claims above. Adopted from review: strict
  version parsing, replacing textual validation with a real parse/re-parse, and
  the release backstop asserting the value rather than only its presence.

## Not in scope

- **Windows** (`latest.yml`) is deliberately untouched — the same field is
  compared against the NT version there, so a macOS floor would wrongly block
  Windows updates.
- The release body template still says "macOS 12 (Monterey) or later"; that is
  #434.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -156,6 +156,67 @@ jobs:
         APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.APPLE_APP_SPECIFIC_PASSWORD }}
         APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
 
+    # Write the macOS floor into the update manifest (#432). The runtime gate in
+    # app/update-os-gate.js cannot protect the installs that are actually at
+    # risk: it only runs inside a build that already meets the floor, while the
+    # machines that would be bricked are running the PREVIOUS release. Only
+    # latest-mac.yml reaches them — electron-updater checks its
+    # `minimumSystemVersion` before it reports an update or downloads one.
+    #
+    # electron-builder does NOT write that field. `mac.minimumSystemVersion`
+    # only reaches Info.plist (app-builder-lib macPackager.js) and `ReleaseInfo`
+    # in its schema is additionalProperties:false. Verified against 26.8.1: a
+    # build with mac.minimumSystemVersion set emits the plist key and a
+    # latest-mac.yml with zero occurrences of minimumSystemVersion.
+    #
+    # The value MUST be the DARWIN version — electron-updater compares against
+    # os.release(), which on macOS is the kernel version. The product floor
+    # 14.4.0 would be a silent no-op, since macOS 12 reports Darwin 21.6.0.
+    # darwinFloorForMacos() does the translation from the same constant the
+    # runtime gate uses, so the two can never disagree.
+    #
+    # Runs here rather than in the release job because `npm ci` has run, so the
+    # patched manifest can be re-parsed with a real YAML parser instead of being
+    # eyeballed. A missing js-yaml (electron-builder's, hoisted) fails the step —
+    # the right direction: never publish an ungated manifest.
+    - name: Write the macOS floor into the update manifest
+      working-directory: app
+      run: |
+        node -e "
+          const fs = require('fs');
+          const yaml = require('js-yaml');
+          const { darwinFloorForMacos, MIN_MACOS_FOR_AUTOUPDATE } = require('./update-os-gate');
+          const file = 'dist/latest-mac.yml';
+
+          const floor = darwinFloorForMacos(MIN_MACOS_FOR_AUTOUPDATE);
+          const before = fs.readFileSync(file, 'utf8');
+          const parsed = yaml.load(before);
+          if (!parsed || typeof parsed !== 'object') {
+            throw new Error('latest-mac.yml did not parse as a mapping');
+          }
+          if (parsed.minimumSystemVersion !== undefined) {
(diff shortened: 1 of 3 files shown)


Optic00:
**End-to-end check, since a decision this consequential shouldn't rest on reading the library's source.**

I drove a **real electron-updater 6.8.3 instance** — app version `0.6.4`, a full `checkForUpdates()` cycle, `autoDownload` on — against an HTTP feed serving the manifest **this PR's workflow step actually produced** (from a real `electron-builder --mac zip --arm64` run), with `os.release()` stubbed to the Darwin version of each macOS release:

| macOS | Darwin | with this PR (`23.4.0`) | manifest as it ships today | with the product floor `14.4.0` |
|---|---|---|---|---|
| 12 | 21.6.0 | **refused** (`update-not-available`) | update offered | update offered |
| 13 | 22.6.0 | **refused** | update offered | update offered |
| 14.3 | 23.3.0 | **refused** | — | — |
| 14.4 | 23.4.0 | offered, download starts | — | — |
| 26.5 | 25.5.0 | offered, download starts | — | — |

The two right-hand columns are the counterfactuals worth having on the record: the manifest as `main` produces it today offers the 14.4-only build to macOS 12/13, and so does the obvious `minimumSystemVersion: 14.4.0`, because electron-updater compares that field against `os.release()` — the Darwin kernel version, where macOS 12 reports `21.6.0`. Only the Darwin value refuses.

On the 14.4+ rows the updater proceeds to fetch the artifact and then errors on my stub feed, which is the point: it got past the gate and started downloading.

Harness: a minimal Electron app + a local static feed, no mocking of electron-updater itself — only `os.release()` is stubbed, which is the one thing that can't be varied on a single machine. Happy to attach it if useful.

ruzin (approved):

Issue #337, opened 2026-07-14
RFC: UI localization (i18n) - multi-language support, built to extend


Optic00:
# RFC: UI localization (i18n) - multi-language support, built to extend

## Problem

Steno's interface is English-only. Every user-facing string is hardcoded - across the ~67 React `.tsx` files in the renderer and the Electron-native chrome in `app/main.js` (app menu, tray menu, system notifications, native dialogs). Non-English users get an English UI even when their meeting content (transcript, summary) is already in their own language.

There's clear latent demand: past spoken-language requests (#45, #70/#71/#72) show a non-English user base, and German is a concrete first ask from the community.

**Scope guard - this RFC is UI chrome only.** The *content* language (summaries, reports) is a separate, already-solved concern: `output_language` in `src/config.py` + `src/report_store.py`, plus transcript language detection (#283). This proposal does not touch any of that. Reusing the name `get-language` is off-limits here - it already means transcription/output language (`app/preload.js`, `app/main.js`).

## Proposal

Introduce an i18n layer so the interface can render in multiple languages. English stays the source language. **German lands first** - it drives the scaffolding end-to-end and can be proof-read by native speakers on the team - but this is explicitly *not* an EN/DE-only feature.

The initial target set mirrors the content languages Steno's summary/output path already handles (**French, Spanish, Dutch, Portuguese** - the same European set the language detector covers), with the community-requested **Polish (#71), Norwegian (#70), and Slovak (#72)** as natural fast-follows. Beyond validating German end-to-end, no locale is gated on the maintainers doing the translation - the scaffolding lands once, and each language's JSON can arrive independently (see *Incremental rollout* and the community-contribution note below). The architecture takes further locales without rework.

### Library

`react-i18next` in the renderer (`useTranslation` / `t()`, `<Trans>` only for copy with embedded React nodes), backed by the `i18next` core. A **second, separate** i18next instance runs in the main process for menu/tray/notification/dialog strings.

To be precise: these are two independent instances in two runtimes - nothing synchronizes automatically. They share the same JSON resource files and are coordinated by an explicit protocol (below), not by magic. Configure `fallbackLng: 'en'` and await init before first use in each runtime.

[...]

PR #440, opened 2026-07-26
fix(about): say what actually failed when an update errors, in one sentence

Optic00:
## The bug

The About tab pasted electron-updater's raw error text behind a hardcoded `Update download failed:` prefix. Two things went wrong with that.

It named the wrong phase. A *check* that never got off the ground was reported as a failed *download*, and so was a failure to install. The prefix was hardcoded, so it could not be anything else.

And it was developer text. A dev build with no update feed produced:

> Update download failed: ENOENT: no such file or directory, open '/Applications/Steno.app/Contents/Resources/app-update.yml'

A stack-trace fragment with a bundle path, in an interface that otherwise speaks in sentences.

There was a second, quieter problem: a failed cycle left its banner sitting under a fresh "You're on the latest version", so the tab gave two contradictory answers to the same question, and a remount brought the settled failure back.

## What this changes

A new pure module, `app/update-error-copy.js`, maps an updater error to one sentence. It follows three rules:

- **Name the phase that actually failed.** `updateErrorPhase()` decides from live state in `main.js`. A staged update on its own means nothing here: a periodic check can fail long after a download succeeded, and calling that a failed download is the original bug.
- **Say whether the user has to do anything.** Most of these resolve on the next scheduled check, and saying so is the difference between a warning and a chore.
- **Mark whether a later success disproves the failure** (`sticky`). A dropped connection is disproved by the next good check; a full disk, a missing update feed or a permission problem is not, and clearing those on an unrelated success would hide a condition that is still true.

Sticky failures now survive *finding* a version and clear only once bytes actually arrive (`download-progress`), because that is what disproves them. `main.js` owns that decision and announces it on a new `update-error-cleared` channel, so the About tab follows main rather than reimplementing the rule.

The raw text still goes to the debug log, which is where it is worth having.

Copy is phase- and platform-aware where it has to be: a permission error while fetching says the update could not be *saved*, not *installed*, and the "move Steno to your Applications folder" hint is macOS-only, since Windows has no such folder.

## Tests

- `app/update-error-copy.test.js` — 21 cases: phase decision, branch precedence (an `EACCES` on `app-update.yml` is a permission problem, not a missing feed; an HTTP status is not a connection problem), sticky semantics, the platform/phase copy, and a guard that no developer-shaped string (`net::`, errno, a path) reaches the user.
- `e2e/specs/settings-about.t1.spec.ts` — the existing About specs plus a race regression: a slow status reply must not restore a failure the check just settled. **Verified to fail without the fix.** It needs a first-call-only delay in `app/e2e-mock-ipc.js`, which answers with the state as of the request, as `main.js` does.

Locally: node:test 298 pass, vitest 120 pass, renderer typecheck clean, lint 0 errors, T1 52/52.

## Notes for review

- Rebased onto `v0.6.5` (`9b5f590`), so it sits on top of both #435 and #436. It deliberately did not go into 0.6.5: it touches the update error path that those two changed twice in one day.
- One known limit is documented in the code rather than papered over: electron-updater's `error` event does not say which attempt it belongs to, so a periodic check that fails *while* a download is genuinely running is still attributed to the download. That case was already reported as a download failure before this module existed, and closing it needs per-attempt correlation the library does not expose.

---
## Summary by cubic
Make update errors on the About tab clear and accurate: show a one‑sentence message that names the failed phase (check, download, install), clear stale banners after a successful check, and clear the banner when an update finishes downloading. This replaces raw `electron-updater` text and fixes conflicting “latest version” vs. error states.

- **Bug Fixes**
  - Added `app/update-error-copy.js` to convert `electron-updater` errors into one sentence; raw text stays in debug logs.
  - Phase-aware labeling from `main.js` (check/download/install), including transport errors during install — not mislabeled as downloads.
  - Sticky semantics: clear non‑sticky errors on a clean check; keep sticky ones (disk space, permissions, no update feed) until bytes arrive, clearing on `download-progress` or `update-downloaded`; emit `update-error-cleared`.
  - About tab now clears the banner on `update-downloaded`, listens for `update-error` and `update-error-cleared`, and drops stale `getStatus()` replies via a simple sequence guard so a settled failure can’t reappear.

<sup>Written for commit 8223eee8071a8428dc506444e562357fcb628471. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -583,6 +583,21 @@ function install({ ipcMain }) {
       ? 'mlx-community/parakeet-tdt-0.6b-v3'
       : 'istupakov/parakeet-tdt-0.6b-v3-onnx';
 
+  // Mirrors main.js: a successful check that finds no update settles an earlier
+  // FAILED check, and it does so in the state the About tab rehydrates from —
+  // so the T1 spec asserts the contract (check clears it for good), not just a
+  // local setState. Seeded by STENOAI_E2E_SEED_UPDATE_ERROR.
+  let seededUpdateError =
+    process.env.STENOAI_E2E_SEED_UPDATE_ERROR === '1'
+      ? "Steno couldn't reach the update server. Check your connection — it will try again later."
+      : null;
+
+  // Only the FIRST get-update-status is delayed, so the About tab's mount-time
+  // request is still in flight when the check that settles the error completes,
+  // and its stale reply lands last. Delaying every call would let the check's
+  // own re-read arrive after it and paper over the bug being guarded against.
+  let slowUpdateStatusCallsLeft = process.env.STENOAI_E2E_SLOW_UPDATE_STATUS === '1' ? 1 : 0;
+
   const DEFAULTS = {
     'get-app-version': { success: true, version: '0.0.0-e2e', name: 'Steno' },
     // Read-only display poll for the About tab's "Check for Updates" button
@@ -605,6 +620,10 @@ function install({ ipcMain }) {
           osUpdateEligible: false,
         };
       }
+      // Up to date, and that settles a previously failed check — main.js
+      // clears the persisted error in exactly this branch, so the mock does
+      // too and the spec can assert it stays gone across a remount.
+      seededUpdateError = null;
       return {
         success: true,
         updateAvailable: false,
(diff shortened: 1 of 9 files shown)


Optic00:
Thanks — three of the four are in `8223eee`, one I'm pushing back on.

**Fixed: the banner surviving `update-downloaded`.** Correct, and it was reachable. `main.js` clears both kinds of failure there but sends only `update-downloaded`, so a mounted About tab kept its own copy. In practice `download-progress` usually cleared it first, but on a resumed or cached transfer that event is the first one the tab sees. The listener now clears with it.

**Fixed: a transport error during install labelled a download.** Also right, and it is the same mislabelling this module exists to stop, one level down — applying a staged update touches no network, so promising a retry of a transfer would be wrong. `NETWORK_RE` + install now returns the install sentence, sticky, matching the general install case. Two tests cover it.

**Fixed, as wording: the "one sentence" claim.** The module promised more than it delivers — several branches legitimately produce two, because rule 2 (say whether the user has to do anything) needs its own clause and reads better split. Rather than compress good copy to satisfy a claim, the claim is now accurate: prose, with the guarantee that no developer text survives — no errno, no `net::` code, no path. That guarantee is enforced by a test.

**Not changing: the mock and `update-error-cleared`.** This one is incorrect. The E2E mock replaces `ipcMain` handlers; renderer subscriptions go through the preload's generic `subscribe()` and need no mock entry, so a new event channel cannot break a mount. Empirically as well: four T1 specs mount the About tab and the suite is 52/52 locally and green in CI on this PR, including the three About specs and the new race regression.

Re-verified after the changes: node:test 300, vitest 120, typecheck clean, lint 0 errors, T1 52/52.

Optic00:
Note on ordering: #337's implementation is now up as #441, and it touches four of the same
files as this PR. A test merge conflicts in `app/package.json` (both extend `test:unit`) and
`app/renderer/src/routes/settings/AboutTab.tsx`.

**This one should go first.** It is smaller, older, already reviewed, and fixes a visible bug.
#441 will then rebase on top and extract this PR's new strings — the About-tab error copy and
`app/update-error-copy.js` — as part of that pass. Nothing to change here.

The reverse order would mean reworking this PR onto `t()` afterwards, which would throw away
the review it has already had.

PR #442, opened 2026-07-27
fix(recording): notification timing/focus + duplicate-note fixes

ruzin:
Fixes four related auto-detect/notification UX bugs. Spec: `.specs/2026-07-27-recording-notification-fixes.md` (built via /assemble → ironman → hulk).

## Bugs fixed
1. **Focus-steal** — tapping "Take Notes" no longer yanks the window forward; `requestAutoRecord` drops `mainWindow.show()/focus()` and records in the background pill-dock.
2. **"Note ready" timing** — a `notesGenerated` flag (set only on `STREAM_COMPLETE`) is threaded through `processing-complete` for both the pipeline and reprocess paths; "Note ready" now fires only when notes actually exist.
3. **"Summarise?" too early** — the mic-idle prompt is reframed to "Wrap up" (stops, no summarise wording, no focus); the summarise decision moves to a post-transcription **"Transcript ready — generate notes?"** prompt whose "Generate notes" action runs reprocess in the background → then "Note ready". Engine-agnostic (shared post-stop path).
4. **Duplicate note** — `get-queue-status` surfaces the session's real-note key (`liveSummaryFile`); `useMeetings` drops the synthetic `__live__/…` row once the real note exists → one recording, one entry.

## New IPC (4-file contract synced)
- `show-transcript-ready-notification` (invoke) + `generate-notes-requested` (M→R), across main/preload/ipc.ts/e2e-mock.

## Review (in-context, senior QA + Eng + a confirming pass)
- QA caught a **HIGH** regression (H1: `liveSummaryFile` precedence dropped a fresh recording's row during back-to-back meetings) — fixed (flip precedence).
- A **MEDIUM** (M2: append/continue fired the wrong prompt) — fixed; the confirming pass found the first M2 fix was a no-op (`notes_generated` is never written `true`), re-fixed with the correct `!== false` semantics + a real-value test.
- Engine review: no critical/high; design endorsed.

## Tests
- Pure helpers extracted + unit-tested: `classifyCompletionNotification` + `meetingAlreadyHasNotes` (Bug 2/3), `liveRowRedundant` (Bug 4).
- `notifications.t2` extended to gate the new transcript-ready notification (real backend).
- Gates: typecheck clean, 277 node:test + 138 vitest, IPC contract 8/8, 50/50 T1, notifications.t2 green.

## Verify
Verified at runtime where drivable (T1 app boot + recording/pill-dock flows; notifications.t2 handler+gate). The model-bearing recording→transcription→notification-timing end-to-end isn't drivable in dev (needs mic+model) — noted as the one un-driven slice.

## Follow-ups (non-blocking)
- A T2 exercising get-queue-status during record+process overlap (would regression-guard H1).
- Optional: rename the `auto-summarise-requested` channel (now means "wrap up") — kept for now to avoid 4-file churn, documented in code.

---
## Summary by cubic
Fixes auto-detect recording focus and notification timing (now visibility-aware), and removes duplicate live meeting rows. Recording starts in the background, notifications fire at the right time even when the window is hidden, and each recording shows only one row.

- **Bug Fixes**
  - "Take Notes" no longer steals focus; auto-record runs in the background pill-dock.
  - "Note ready" only fires when notes were generated; threads a `notesGenerated` flag from `STREAM_COMPLETE` through `processing-complete`.
  - Mic-idle prompt is now "Wrap up" (stop only). Body tap just opens the app; the summarization decision moves to a post-transcription "Transcript ready — generate notes?" notification; "Generate notes" runs reprocess in the background.
  - Completion prompts are visibility-aware: they still notify when the window is hidden; disabled window `backgroundThrottling` to keep recorder cadence and silence auto-stop timing consistent.
  - Duplicates removed: `get-queue-status` exposes `liveSummaryFile`, and `useMeetings` drops the synthetic `__live__/…` row once the real note exists.

- **New Features**
  - Added `show-transcript-ready-notification` and `generate-notes-requested` IPC channels to support the new post-transcription flow.

<sup>Written for commit abf92876260766d9a6f7e875b14e742d2ec8f655. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/442?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -429,6 +429,13 @@ function install({ ipcMain }) {
       return { success: true, path: seamPath };
     },
 
+    // Notification handlers — the real ones return { success, shown } after the
+    // notifications_enabled gate; the mock has no OS toast, so it just reports
+    // "shown" so a T1 that drives the transcript-ready branch (#bug2/#bug3) sees
+    // a realistic shape instead of the permissive fallback. (The gate itself is
+    // covered by the T2 spec against the real handler.)
+    'show-transcript-ready-notification': async () => ({ success: true, shown: true }),
+
     // Mirror the real export-note-pdf handler's seam. The mock has no Chromium
     // to rasterise HTML, so instead of a PDF it writes the renderer-built HTML
     // verbatim to STENOAI_E2E_EXPORT_PATH — that lets a T1 spec assert the exact
--- a/app/main.js
+++ b/app/main.js
@@ -4459,14 +4459,25 @@ ipcMain.handle('get-queue-status', async () => {
       (currentRecordingProcess !== null || systemAudioRecordingActive)
         ? (currentRecordingAppendTarget || null)
         : null,
-    // The real note file this recording/processing session produces (the
+    // The real note file the CURRENT recording/processing session produces (the
     // instant-stop placeholder written at stop, then rewritten in place by the
     // pipeline). Lets useMeetings dedupe the synthetic "__live__/…" row against
     // the real note once it exists, so one recording is never shown twice
     // (#bug4). Deterministic from the audio stem, so it's the same key across
-    // the recording → processing phases. Null for Whisper/import (no instant
-    // placeholder) and when idle.
-    liveSummaryFile: currentProcessingJob?.summaryFile || activeSysAudioSummaryFile || null,
+    // this session's recording → processing phases.
+    //
+    // Precedence matters: prefer the ACTIVE recording's key over a PROCESSING
+    // job's. In the supported back-to-back flow (stop meeting A → immediately
+    // start meeting B while A is still processing), `sessionName` reports B (the
+    // recording), so liveSummaryFile must also be B's — otherwise it'd be A's,
+    // A's note is already in the list, and B's live row would be wrongly dropped
+    // (B shows no row at all). At B's own stop, teardown nulls
+    // activeSysAudioSummaryFile before currentProcessingJob is set to B, so the
+    // fallback still yields B and the intended dedup against B's placeholder
+    // fires. Only the Parakeet instant-stop path writes a placeholder, so this
+    // only actually bites there; Whisper/import have no placeholder (dedup is a
+    // no-op) and it's null when idle.
+    liveSummaryFile: activeSysAudioSummaryFile || currentProcessingJob?.summaryFile || null,
   };
 });
 
@@ -6667,6 +6678,13 @@ function showMeetingEndedNotification(appName) {
   // action and a body tap wrap up (low stakes now — it only stops, it doesn't
   // start AI processing), and neither steals focus (background, like
   // requestAutoRecord).
+  //
+  // Gating (deliberate): like the meeting-detected toast, this is an
+  // auto-detect *lifecycle* prompt gated by the auto_detect_meetings toggle,
+  // NOT by notificationsEnabled() — that gate covers the result toasts
+  // (note-ready / transcript-ready / silence). Splitting lifecycle vs result
+  // toasts is intentional so turning off result notifications doesn't strand a
+  // paused auto-detected recording with no way to wrap it up.
   notif.on('action', (_evt, _index) => requestWrapUp());
   notif.on('click', () => requestWrapUp());
   trackNotificationLifecycle(notif, 'meeting_ended');
--- a/app/main.js
+++ b/app/main.js
@@ -1511,15 +1511,11 @@ function createWindow(options = {}) {
       sandbox: true,
       preload: path.join(__dirname, 'preload.js'),
       scrollBounce: true,
-      // Audio capture is fully renderer-owned (getUserMedia + MediaRecorder on a
-      // 1s timeslice + a 1s silence-auto-stop interval, useSystemAudioCapture.ts).
-      // A recording can legitimately run while the window is HIDDEN — the close
-      // handler hides rather than quits, and #bug1 makes auto-detect start
-      // recording without ever showing the window. Chromium background-throttles
-      // timers/rAF in hidden windows by default, which would slow the capture
-      // timeslice and silence-auto-stop cadence. Disable it so capture behaves
-      // the same hidden as visible.
-      backgroundThrottling: false,
+      // backgroundThrottling is left at its default (true) and toggled at
+      // runtime instead — see applyRecordingBackgroundThrottling(). We only
+      // disable throttling WHILE recording (renderer-owned capture timers must
+      // run full-rate even when the window is hidden), and let a
+      // backgrounded-idle app throttle normally so it doesn't burn CPU.
     },
     // Windows/Linux render the Electron application menu as an in-window menu
     // bar (File/Edit/View/…); macOS puts it in the global bar. Hide it off-mac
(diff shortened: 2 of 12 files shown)


Optic00 (changes requested):
## Verdict

Request changes. The design is right and the four bugs are real, but as written the headline fix (bug 3) does not fire in the flow it was built for, and fix 1 quietly moves the whole recording path into a hidden, background-throttled renderer.

## Summary

This fixes four related auto-detect and notification bugs: the focus-steal on "Take Notes", "Note ready" firing when no notes exist, the premature "Summarise?" prompt at meeting end, and a recording showing as two rows on Home. The approach is sound. Threading a real `notesGenerated` signal out of `STREAM_COMPLETE` instead of inferring it is the correct fix for bug 2, the `notes_generated` frontmatter semantics in `meetingAlreadyHasNotes` are exactly right (the backend only ever writes `false` or omits the key, so `!== false` is the only correct test), and pulling the classification into pure, unit-tested helpers is a real improvement over inline logic.

My concern is the interaction between the fixes rather than any one of them. Fix 1 removes the guarantee that the window is visible, and three other pieces of this change still assume it is.

## Findings

### Blocker

**1. The transcript-ready prompt never fires in its own main flow (Parakeet).**
`app/renderer/src/hooks/useRecording.ts:479-485`

The completion handler decides whether to notify by comparing the current route against the finished note's route. There are three branches: on `/meetings/processing` it navigates and does not notify, on any *other* route it notifies, and when the current route already *is* the finished note's route it does nothing at all.

Now trace "Wrap up" on macOS, where Parakeet is the default engine. `requestWrapUp()` calls the renderer's `stopRecording()`, which on the instant-stop path receives `data.summaryFile` and navigates to ``/meetings/${summaryFile}`` (`useRecording.ts:219`). That is precisely `finishedMeetingRoute`. So when processing completes, `currentRoute === finishedMeetingRoute`, the `else if` at line 485 is false, and **no notification fires at all** - neither note-ready nor the new transcript-ready.

The route guard is not new and its original reasoning was sound: if the user is sitting on the note's detail page, a banner is noise. But that reasoning assumed a visible window, and fix 1 in this same PR removes that assumption. After an auto-detected meeting wrapped up from a tray-only Steno, the route says "the user is looking at the note" while the window is hidden and the user is looking at something else entirely.

Net effect: the flow this PR exists to fix (auto-detected meeting ends, user wraps up, gets prompted after transcription) produces no prompt on the default macOS engine. Worth deciding whether the guard should consider `mainWindow.isVisible()` rather than the route alone.

### Important

**2. Recording now starts in a hidden, background-throttled renderer.**
`app/main.js:6717-6723`

Capture is entirely renderer-owned: `useSystemAudioCapture.ts` runs `getUserMedia`/`getDisplayMedia`, a `MediaRecorder` with a 1 s timeslice (`:492`), and a 1 s `setInterval` driving silence auto-stop (`:571`). Dropping `mainWindow.show()` is correct for the focus-steal complaint, but the window's `close` handler preventDefaults and hides (`main.js:1578-1581`), so the normal state for a user who closed Steno is a hidden window, and `backgroundThrottling` is not set in `webPreferences` (`main.js:1508`), so it defaults to `true`.

I have not measured this, and Chromium does exempt capturing pages from some throttling tiers, so I am not claiming it is broken. But the one behaviour this PR changes here is also the one slice the PR notes as un-driven by tests, and the failure mode (silence auto-stop running at a fraction of its intended cadence, or capture degrading in a hidden window) is not something a user would report as "the notification changed". Please confirm on real hardware: close the Steno window so it is tray-only, trigger auto-detect, tap "Take Notes", and check that capture and silence auto-stop behave as they do with the window open. If they do not, `backgroundThrottling: false` on the capture window is the cheap fix.

**3. A body tap on the meeting-ended notification now commits.**
`app/main.js:6689`

Previously the body click deliberately did not commit: it opened Steno so the user could decide, and the removed comment spelled out why. Now both the action and the body tap call `requestWrapUp()`, and the new comment justifies it as "low stakes now - it only stops, it doesn't start AI processing". That is not accurate: stopping hands the recording to the shared pipeline, which transcribes it, and with auto-summarize on it summarises too.

So a stray tap on a notification body ends a meeting that was only auto-paused on mic-idle, and because `requestWrapUp()` also no longer shows the window, nothing appears on screen to tell the user it happened. If mic-idle was a false positive and the meeting is still running, they have lost the recording from that point on. Suggest keeping the split that was there: action wraps up, body tap opens the app.

**4. The live-row dedup drops the wrong row in the back-to-back flow it documents.**
`app/main.js:4480`, `app/renderer/src/hooks/useMeetings.ts:87`

The precedence comment argues that when B starts while A is still processing, `sessionName` reports B so `liveSummaryFile` must be B's too. The ordering does not deliver that. `start-recording-ui` sets `currentRecordingSessionName = B` at `main.js:5846`, and the comment immediately below it states the design: the active flag is flipped "so the queue handler reports hasRecording=true on the very next poll, which is what cues the renderer hook to fire startCapture". `activeSysAudioSummaryFile` is only assigned later, inside `open-system-audio-file`, after the renderer has acquired mic and loopback.

For at least one full poll interval plus device acquisition, `get-queue-status` therefore returns `sessionName: B` with `liveSummaryFile` falling through to `currentProcessingJob.summaryFile`, which is A. A's instant-stop placeholder is already in the list, so `liveRowRedundant` returns true and **B's live row is dropped** for that window. That is the same "recording invisible on Home" symptom from the other direction. Gating the fallback on there being no active recording would close it.

### Nit

**5. The no-`meetingData` fallback loses the M2 append fix.**
`app/main.js:5464-5477`, `app/renderer/src/lib/completionNotification.ts:56`

When the `list-meetings` lookup fails, `processing-complete` is sent without `meetingData`, so `meetingAlreadyHasNotes` returns false and an append into a note that already has notes classifies as transcript-ready again. The docstring calls this out as deliberate (prefer falling back to the transient signal over defaulting a transcript-only note to "has notes"), and I agree that is the safer direction of the two, so this is a note rather than a request. Mentioning it only so the trade-off is visible outside the docstring.

### Praise

The `meetingAlreadyHasNotes` docstring is the good kind of comment: it records that the first fix was a no-op because `notes_generated` is never written `true`, and the test file exercises the values the call site actually receives rather than the ones the type allows. That is what stopped the second attempt from being wrong in the same way.

## Note on review method

Reviewed at `18b073c`. CI was 11/11 green (cubic skipped on its monthly limit, not a failure). Findings 1, 3 and 4 were also reached independently by a cross-family reviewer; 1 and 4 I confirmed by reading the paths through myself before including them here. Finding 2 is unverified by design, as flagged above.

ruzin:
Thanks @Optic00 — all three were spot-on. Fixed in abf9287:

**Blocker (transcript-ready never fires in the Parakeet wrap-up flow).** Correct — the route guard assumed a visible window, which bug 1 removed, and `stopRecording()` navigates to the note's own route, so `currentRoute === finishedMeetingRoute` and nothing fired. The guard is now **visibility-aware**: extracted `completionRouteAction()` (pure, unit-tested) that only suppresses when on the note's route **and the window is actually visible** (`document.visibilityState`); otherwise it notifies. So the tray-only wrap-up flow now fires the prompt. Added a unit test whose headline case is exactly this (on the note's route + hidden → notify), plus processing-hidden → notify and the visible cases.

**Important (hidden, background-throttled capture).** Agreed — I couldn't measure it here, so I took the cheap fix you suggested: `backgroundThrottling: false` on the main window's `webPreferences`. Capture is renderer-owned (MediaRecorder 1s timeslice + 1s silence-auto-stop interval) and bug 1 makes hidden-window recording the default auto-detect path, so this keeps timer cadence identical hidden vs visible rather than relying on Chromium's capture-exemption.

**Important (body tap now commits).** Fixed — restored the split you called out: the **Wrap up action** stops; a **body tap only opens Steno** (show/focus) so the user decides. You're right that stopping isn't low-stakes (it hands off to the pipeline + summarises), and with wrap-up no longer showing the window a stray body-commit would be an invisible data loss on a false-positive mic-idle.

Gates green after the fixes: typecheck, 277 node:test + 143 vitest, 50/50 T1 (re-verified app boot + recording flows with the new webPref). CI re-running.

ruzin:
Heads up @Optic00 — pushed a **second refinement round** (8338276) after product feedback; the notification/focus model changed materially since your review, so this wants a fresh look:

- **Take Notes now focuses + opens the live note** (reverses the earlier no-focus): tapping is an explicit intent to write, so `requestAutoRecord` restores show()/focus() and the renderer navigates to /recording. The passive toast still never focuses.
- **Meeting end auto-stops — the "Wrap up" notification is gone.** Mic-release reliably means the call ended (software-mute keeps the stream open, per the MEETING_END_DEBOUNCE_MS comment), so mic-idle → auto-pause + a 20s grace (auto-resume cancels it on a device switch) → auto-stop. `showMeetingEndedNotification` deleted; `requestWrapUp` → `autoStopEndedMeeting` (timer-fired).
- **Summarise runs in the background again** (no focus): "Note ready" on completion is the come-in moment.
- **Accurate "Finishing up" copy**: only says "generating notes" when auto-summarize is on.
- **Dynamic backgroundThrottling** (addresses your throttling flag properly): replaced the static `backgroundThrottling:false` webPref with `applyRecordingBackgroundThrottling()` — throttling is disabled only WHILE recording, so a backgrounded-idle app throttles normally instead of burning CPU.

Gates green: typecheck, 277 node:test + 143 vitest, 50/50 T1 re-verified. Separately filed #462 for the recording-time CPU (ScriptProcessorNode → AudioWorklet) — out of scope here.

PR #451, opened 2026-07-31
fix(analytics): classify summarization STREAM_ERROR reasons

### LINKED ISSUE #450: summarization_completed failures are 86% of tracked errors, but the real cause is invisible — STREAM_ERROR message never reaches analytics
## Problem
Per the July product analytics review, `error_occurred` with `stage=summarization, reason=subprocess_exit_1` accounts for 44 of 51 tagged errors (86%) in the last 60 days — the dominant tracked failure mode in the whole pipeline.

## Root cause
`simple_recorder.py::process_streaming()` ([L1218-1241](https://github.com/ruzin/stenoai/blob/main/simple_recorder.py#L1218)) deliberately catches any exception from `summarizer.summarize_transcript_streaming()`, prints `STREAM_ERROR:<message>` to **stdout**, and calls `sys.exit(1)`.

On the Electron side, `main.js` builds its error from **stderr** only ([L4477](https://github.com/ruzin/stenoai/blob/main/app/main.js#L4477)), and `classifyErrorReason()` just regexes the generic "exited with code N" out of that — so every summarization failure collapses into `subprocess_exit_1` regardless of what actually went wrong. The real message *is* captured locally via `forwardDiagnosticStdout` → `sendDebugLog` ([L5685](https://github.com/ruzin/stenoai/blob/main/app/main.js#L5685)), but never reaches telemetry.

## Likely candidates (unconfirmed — this is exactly what the fix would reveal)
Based on `src/summarizer.py`'s raise sites, and given 92.5% of users are on local Ollama (`ai_provider_selected`):
- `"Failed to start Ollama service"`
- `"Failed to ensure model <X> is available"`
- `"Reduce step returned empty result"` — possibly long-meeting map-reduce/context overflow (referenced in an existing main.js comment)

## Proposed fix
1. **Immediate:** classify the `STREAM_ERROR:` message into a small set of fixed, PII-free reason buckets (e.g. `ollama_not_running`, `model_unavailable`, `empty_summary`, `api_error`) and attach that classification — never the raw message — as a property on the `error_occurred` event when `stage=summarization`. This should follow the existing pattern in `app/analytics-helpers.js` (`classifyErrorReason`), which deliberately keeps every tracked reason to a fixed enum rather than free text, since raw summarizer error text can carry transcript fragments or API response bodies.
2. **Follow-up:** once real error categories are visible in PostHog, fix whichever specific cause turns out to dominate.

## Acceptance criteria
- `error_occurred` for `stage=summarization` failures carries enough classified detail to distinguish "Ollama not running" from "model missing" from "empty reduce result" without reading a user's local debug log.
- No sensitive content (transcript text, file paths, API response bodies) leaks into the tracked property — matches the existing fixed-enum pattern used by `classifyErrorReason`, not a raw-string passthrough.

## Reference
Steno AI July Product Analytics Review, finding ACT·2.

WilliamDrewett:
## Description

Summarization failures accounted for 86% of tracked `error_occurred` events, but all collapsed into an uninformative `subprocess_exit_1` -- the real failure reason (Ollama not running, model missing, empty map-reduce result, etc.), printed by the Python backend as `STREAM_ERROR:<message>` on stdout, was captured for the local debug log but never reached analytics. This adds `classifySummarizationStreamError()` to bucket that message into a small, fixed, PII-free enum and attaches it as a new `summarization_reason` property on `error_occurred` when the failure happens during the summarization stage. No user-facing behavior changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Unit tests added/passing (`node --test app/analytics-helpers.test.js`, `node --test app/*.test.js` -- 279/279 pass)
- [ ] Tested locally with `npm start`
- [ ] Tested on macOS
- [x] No breaking changes to existing functionality (analytics-only change)

## Additional Notes

Fixes https://github.com/ruzin/stenoai/issues/450. Only wires up the `processing_queue`/summarization path, which drives the cited 86% figure -- `reprocess`/`generate-report`/chat streaming don't fire `error_occurred` at all today, which is a separate, unrelated instrumentation gap left out of scope here. Root-causing which bucket actually dominates is the issue's own stated follow-up, pending real PostHog data.

---
## Summary by cubic
Classifies summarization `STREAM_ERROR` messages into a small PII-safe enum and adds `summarization_reason` to `error_occurred` so analytics show real failure causes instead of an opaque `subprocess_exit_1`. No user-facing changes.

- **Bug Fixes**
  - Capture `STREAM_ERROR:` during summarization and map via `classifySummarizationStreamError()`; guard the write so only the summarization stage is recorded.
  - Extend classification to mid-stream Ollama failures (`Connection refused`/`ConnectError`) and out-of-memory model loads (`insufficient_memory`).
  - Add `summarization_reason` to `error_occurred` only for the summarization stage.
  - Buckets: `ollama_not_running`, `model_unavailable`, `insufficient_memory`, `context_overflow`, `empty_summary`, `auth_rejected`, `not_configured`, `dependency_missing`, `api_error`, `unknown` (never forwards raw text).

- **Tests**
  - Added unit tests plus a drift-guard that asserts classifier patterns match `src/summarizer.py` strings.

<sup>Written for commit f8b0cd661c351c4893e3348cb4457f0014ec08ea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app/analytics-helpers.js
+++ b/app/analytics-helpers.js
@@ -140,13 +140,37 @@ function classifyErrorReason(error) {
 // Bedrock's raw response body), whose text is unpredictable and can echo
 // request content or infra details -- those simply won't match any pattern
 // below and correctly fall through to 'unknown', same as classifyErrorReason.
+//
+// The connection-refused half of ollama_not_running and the
+// insufficient_memory branch match text that ORIGINATES from the Ollama
+// server / the underlying httpx client on a bare re-raise, not a literal
+// authored in summarizer.py -- unlike every other branch here, those two
+// are exempt from the drift-guard test in analytics-helpers.test.js (there
+// is no summarizer.py string for them to grep). "Connection refused"/
+// "ConnectError" mirrors the same detection already used for the
+// remote-Ollama status check in simple_recorder.py.
 function classifySummarizationStreamError(message) {
   const msg = String(message || '');
   if (/Failed to start Ollama service/i.test(msg)) return 'ollama_not_running';
+  // A stream can die mid-generation even after the pre-flight readiness
+  // check in _ensure_ollama_ready already passed (Ollama killed/crashed
+  // during generation) -- functionally the same "Ollama isn't there" case.
+  if (/Connection refused|ConnectError|ECONNREFUSED/i.test(msg)) return 'ollama_not_running';
   if (/Failed to ensure model .* is available|could not find model/i.test(msg)) return 'model_unavailable';
+  // Ollama's own server error when a model doesn't fit in available RAM,
+  // e.g. "model requires more system memory (10.5 GiB) than is available
+  // (8.0 GiB)" -- distinct from model_unavailable (the model IS installed).
+  if (/requires more system memory/i.test(msg)) return 'insufficient_memory';
   if (/too long to summarize even after chunking/i.test(msg)) return 'context_overflow';
-  if (/returned empty result|returned an empty result after retry|returned empty response|returned an empty (summary|report)/i.test(msg)) return 'empty_summary';
-  if (/rejected the (request|api key)/i.test(msg)) return 'auth_expired';
+  // Deliberately matches just "returned (an) empty result/response/summary/
+  // report" with no trailing anchor (e.g. NOT "...after retry") -- anchoring
+  // to what follows would make this brittle to a copy-edit of the rest of
+  // the sentence, same drift risk the guard test below exists to catch.
+  if (/returned (an )?empty result|returned empty response|returned an empty (summary|report)/i.test(msg)) return 'empty_summary';
+  // Neutral name (not auth_expired): the adapter's 401/403 really is a
+  // session expiry, but Bedrock's 403 is a missing IAM permission -- a
+  // rejected credential either way, not necessarily an expired one.
+  if (/rejected the (request|api key)/i.test(msg)) return 'auth_rejected';
   if (/is not configured/i.test(msg)) return 'not_configured';
   if (/is not installed|package is required/i.test(msg)) return 'dependency_missing';
   if (/failed after all retries|HTTP \d+/i.test(msg)) return 'api_error';
--- a/app/analytics-helpers.js
+++ b/app/analytics-helpers.js
@@ -131,6 +131,28 @@ function classifyErrorReason(error) {
   return 'unknown';
 }
 
+// Coarse, PII-free classification of the STREAM_ERROR:<message> line
+// simple_recorder.py's process_streaming() prints on a summarization
+// failure (see src/summarizer.py's raise sites). Fixed enum output only --
+// NEVER forwards the raw message to PostHog. Most raise sites in
+// summarizer.py are static/templated text (safe to pattern-match by name),
+// but several are bare re-raises of third-party SDK/HTTP exceptions (and
+// Bedrock's raw response body), whose text is unpredictable and can echo
+// request content or infra details -- those simply won't match any pattern
+// below and correctly fall through to 'unknown', same as classifyErrorReason.
+function classifySummarizationStreamError(message) {
+  const msg = String(message || '');
+  if (/Failed to start Ollama service/i.test(msg)) return 'ollama_not_running';
+  if (/Failed to ensure model .* is available|could not find model/i.test(msg)) return 'model_unavailable';
+  if (/too long to summarize even after chunking/i.test(msg)) return 'context_overflow';
+  if (/returned empty result|returned an empty result after retry|returned empty response|returned an empty (summary|report)/i.test(msg)) return 'empty_summary';
+  if (/rejected the (request|api key)/i.test(msg)) return 'auth_expired';
+  if (/is not configured/i.test(msg)) return 'not_configured';
+  if (/is not installed|package is required/i.test(msg)) return 'dependency_missing';
+  if (/failed after all retries|HTTP \d+/i.test(msg)) return 'api_error';
+  return 'unknown';
+}
+
 // This module lives in the app/ directory alongside main.js and is always
 // required from there via a relative path, so __dirname here is identical to
 // main.js's own -- the app's install root, whether that's an asar-packaged
@@ -368,6 +390,7 @@ module.exports = {
   sanitizeTrackProperties,
   calendarMeetingProvider,
   classifyErrorReason,
+  classifySummarizationStreamError,
   captureSanitizedException,
   redactLocalPaths,
   sanitizeModelForAnalytics,
--- a/app/analytics-helpers.test.js
+++ b/app/analytics-helpers.test.js
@@ -11,6 +11,7 @@ const {
   sanitizeTrackProperties,
   calendarMeetingProvider,
   classifyErrorReason,
+  classifySummarizationStreamError,
   captureSanitizedException,
   redactLocalPaths,
   sanitizeModelForAnalytics,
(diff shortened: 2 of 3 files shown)


Optic00 (commented):
## Verdict

Comment. The change is correct, well tested and privacy-clean, and I found no blockers. The suggestions below are mostly about whether the new metric will actually answer the question in #450, not about whether this code is right.

## Summary

`processNextInQueue` now captures the `STREAM_ERROR:<message>` line the backend prints on a summarization failure and attaches a classified `summarization_reason` to `error_occurred`, so the 86% of tracked errors that currently collapse into `subprocess_exit_1` become distinguishable. Classification is a fixed enum, the raw message is never forwarded, and the local debug log is unchanged because the new branch still calls `forwardDiagnosticStdout`. No user-facing behavior changes.

I reviewed 6dd8b8e. All 12 CI checks are green. A second pass came from a different model family, and its findings are folded in below.

## Findings

### Important

**1. `unknown` may still swallow the dominant case (`app/analytics-helpers.js:143`)**

`ollama_not_running` and `model_unavailable` only fire for the two messages raised by `_ensure_ollama_ready()` (`src/summarizer.py:752`, `758`), which is a pre-flight check before the stream starts. The interesting failure for a long meeting is Ollama dying or refusing *during* the stream, or the model not fitting in memory. Both surface as a bare `ollama`/`httpx` exception, so they land in `unknown` along with every cloud failure.

Given the issue's own note that 92.5% of users are on local Ollama, there is a real chance the new property reads mostly `unknown` and the follow-up in #450 stays blocked for another release cycle of data. Two more patterns would cover it cheaply: the connection-refused shape, and Ollama's "requires more system memory" text. Both are Ollama's own fixed strings, so they carry no more PII risk than the ones already matched.

**2. The enum is coupled to Python string literals with nothing pointing back (`app/analytics-helpers.js:143`, `app/analytics-helpers.test.js:218`)**

Every pattern matches a literal in `src/summarizer.py`, but the tests assert against their own copies of those strings. If someone rewords `"Reduce step returned empty result"` on the Python side, the whole JS suite stays green and the bucket silently degrades to `unknown` with no signal anywhere. A one-line comment at each raise site in `summarizer.py`, or a test that greps `summarizer.py` for the matched substrings, closes that for very little effort.

### Nit

**3. A pre-summarization `STREAM_ERROR` would be reported as the summarization reason (`app/main.js:5224`, `5343`, `5462`)**

The captured value is written unguarded and only read under the stage check, so the correctness of the property rests on a backend invariant: no `STREAM_ERROR` line before `TRANSCRIPTION_COMPLETE`. That invariant does hold today (`process_streaming` prints its only `STREAM_ERROR` at `simple_recorder.py:1313`, well after the marker at `1180`), so this is not a live bug. But if a future transcription-stage `STREAM_ERROR` is ever added, a summarization failure that dies *without* its own message, e.g. the `metal::malloc` SIGABRT your own comment mentions, would report the stale earlier reason. Guarding the write instead of the read, or clearing the variable when the stage flips, removes the class of bug in one line.

**4. `auth_expired` overstates a Bedrock 403 (`app/analytics-helpers.js:149`)**

`"Bedrock rejected the API key (403). Verify the key has bedrock:InvokeModel access..."` is a missing-IAM-permission error, not an expired session, but it buckets as `auth_expired`. A user reading that in PostHog would chase the wrong cause. Either a neutral name like `auth_rejected` or splitting 401 from 403 would fix it.

**5. `after retry` makes one pattern needlessly brittle (`app/analytics-helpers.js:148`)**

`returned an empty result after retry` would match on `returned an empty result` alone. Dropping the tail costs nothing and survives a copy edit to the second half of that sentence. Same theme as finding 2.

**6. Two of the `api_error` inputs are unreachable (`src/summarizer.py:800`, `854`, `1040`)**

`_openai_chat` re-raises the original SDK exception on the final attempt (`src/summarizer.py:796-798`), so `"OpenAI chat failed after all retries"` is never actually raised. Same shape for the adapter and Bedrock variants. Not introduced here and harmless, but it means `api_error` will in practice come almost entirely from the `HTTP \d+` half of that pattern.

### Praise

The reasoning for letting bare re-raises of third-party exceptions fall through to `unknown`, rather than pattern-matching them into a false-confidence bucket, is the right call, and I like that there is a test asserting exactly that rather than only a comment. Keeping `forwardDiagnosticStdout` on the new branch, so the local debug log stays byte-for-byte identical, is the other detail that made this easy to review.

Optic00:
One follow-up on the review, mostly about timing.

I'd like to get this in soon rather than let it sit, so I want to settle the two Important points first. If the coverage gap stays open, the first release carrying this likely reports mostly `unknown`, and the real follow-up in #450 waits another cycle for data that doesn't say much. That's the only reason I'm nudging instead of approving as is.

Do you want to add the two extra Ollama patterns and the drift guard yourself? Happy either way. If you'd rather I did it, I'll open a PR against your branch rather than push to it: a push from me makes my own approval stop counting for the newest commit, which is what has had #384 sitting at BLOCKED since the 24th.

No rush, just say which you prefer.

WilliamDrewett:
Thanks for reviewing, I can do the things mentioned np

WilliamDrewett:
Pushed f8b0cd6 addressing both Important findings, plus the two applicable nits:

- **#1 (unknown swallowing the dominant case):** added patterns for Ollama dying/refusing mid-stream (`Connection refused`/`ConnectError`, matching the same detection already used in `simple_recorder.py`'s remote-Ollama check) and out-of-memory model loads (`requires more system memory` → new `insufficient_memory` bucket), both distinct from the pre-flight-only checks.
- **#2 (drift guard):** added a test that greps `src/summarizer.py` for the literal substrings each pattern depends on, so a future rewording fails the JS suite instead of silently degrading to `unknown`.
- **Nit #3:** guarded the `STREAM_ERROR` capture on the write (`processingStage === 'summarization'`), not just at read time.
- **Nit #4:** renamed `auth_expired` → `auth_rejected` (Bedrock's 403 is a missing IAM permission, not an expired session).
- **Nit #5:** dropped the brittle `...after retry` anchor from the empty-summary regex.

Left #6 (unreachable `api_error` string literals in `summarizer.py`) alone since it's pre-existing dead code, not introduced here, and out of scope for this PR.

All 281 unit tests pass (`node --test app/*.test.js`), including 3 new/updated cases. Ready for another look whenever you have time.

Optic00 (approved):
## Verdict

Approve. Both Important findings are properly addressed, the three nits you took are done cleanly, and skipping #6 was the right call. One small gap in the new drift guard is worth a follow-up commit but doesn't need to hold this up.

## What I checked

Reviewed f8b0cd6 against 6dd8b8e. All 11 CI checks green (cubic and Mintlify skipped).

**#1, coverage:** the mid-stream cases are now covered, and separating `insufficient_memory` from `model_unavailable` is a better call than what I suggested, since "the model is installed but doesn't fit" is a genuinely different user problem with a different fix. Matching Ollama's own `requires more system memory` wording keeps it as PII-free as the rest.

**#2, drift guard:** the grep test does what it needs to, and documenting *why* the two runtime-text branches are exempt, in both the test and the classifier, is better than silently omitting them.

**#3:** guarding the write rather than the read is the stronger version, and leaving `forwardDiagnosticStdout` outside the guard keeps the debug log byte-for-byte unchanged. Exactly right.

**#4, #5:** both good. The comment explaining why the empty-summary pattern deliberately has no trailing anchor will stop someone helpfully "fixing" it later.

## Findings

### Nit

**1. Three branches are still outside the drift guard (`app/analytics-helpers.test.js:281`)**

`expectedSubstrings` misses the literals behind three matching branches, so the exact silent degradation the test exists to prevent can still happen to them:

- `returned an empty summary` and `returned an empty report` (`src/summarizer.py:1568`, `1571`, `1575`)
- `HTTP ` from the `HTTP \d+` half of `api_error` (`src/summarizer.py:1035`)

Three more strings in the array closes it. Related and weaker: the guard checks `Failed to ensure model` but not the ` is available` tail the pattern also depends on, so a reword of the second half degrades that branch too.

**2. `Connection refused` is provider-blind (`app/analytics-helpers.js:152`)**

The branch returns `ollama_not_running` for any connection failure, including a cloud provider being unreachable from the user's network. Right the overwhelming majority of the time given the 92.5% local split, and I'd rather have this than nothing, but the bucket name will read as more certain than it is for the cloud minority. Only worth doing something about if the cloud share grows.

### Praise

Turning finding #1 into an extra bucket instead of stretching an existing one, and writing down the reasoning for each pattern choice in comments rather than just changing the regex, makes this materially easier to maintain than what I asked for. Nice turnaround too.

Issue #443, opened 2026-07-27
download-ollama.sh: a failed download is saved as the artifact, so CI fails three steps later with a misleading error


Optic00:
All four downloads in `scripts/download-ollama.sh` use `curl -L` without `--fail`, so an HTTP
error is written to disk as if it were the artifact. Only the macOS ffmpeg branch validates
what it got; the Windows and Linux ffmpeg branches and the Ollama download do not.

## What it looks like when it bites

`Build backend bundle (Windows)` on #441 failed like this:

```
100      9 100      9   0      0     69      0
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.
unzip:  cannot find zipfile directory in one of ffmpeg.zip or
        ffmpeg.zip.zip, and cannot find ffmpeg.zip.ZIP, period.
##[error]Process completed with exit code 9
```

Nine bytes — the length of `Not Found`. The GitHub release redirect returned an error page,
curl happily saved it as `ffmpeg.zip`, and `unzip` produced an error that says nothing about a
failed download. The asset was fine: the same URL returns 200 and a 168 MB zip on retry, and
the job passed on a re-run. So the underlying event was transient; what made it expensive was
that the failure surfaced three steps away from its cause.

Run: https://github.com/ruzin/stenoai/actions/runs/30274798683/job/90006009060

## Where the gap is

| Download | `--fail` | validated after extract |
|---|---|---|
| ffmpeg, macOS (`:26`) | no | **yes** — asserts arm64 Mach-O and pins `ffmpeg version 7.1` |
| ffmpeg, Linux (`:57`) | no | no |
| ffmpeg, Windows (`:70`) | no | no |
| Ollama, all platforms (`:115`) | no | no |

The macOS branch already carries a comment explaining exactly this failure mode ("Truncated/
corrupt download or wrong-format extract"). The guard just never made it to the other
branches.

## Suggested fix

Two small things, no behaviour change on the happy path:

1. `curl --fail --retry 3 --retry-delay 2 -L` for all four. `--fail` turns an HTTP error into
   a non-zero exit instead of a saved error page, and the retries absorb exactly the class of
   blip seen above.
2. Give the Windows and Linux ffmpeg branches, and the Ollama extract, the same
   post-conditions the macOS branch has — at minimum that the expected binary exists and runs.
   `ffmpeg.exe -version` is enough to catch a truncated or wrong-format archive.

[...]

PR #454, opened 2026-07-31
fix(download): add --fail --retry to curl calls in download-ollama.sh (#443)

MeiSiristhebest:
## Summary

Refs #443.

`scripts/download-ollama.sh` downloaded four artifacts with `curl -L` and no `--fail`, so a
transient HTTP error (e.g. a CDN 404 returning a 9-byte "Not Found" page) was written to disk
as if it were the artifact. The failure then surfaced three CI steps later as a misleading
"corrupt archive" error.

### 1. Fail-fast curl flags (all four downloads)

Added `--fail --retry 3 --retry-delay 2 --retry-all-errors` to all four `curl` calls
(macOS/Linux/Windows ffmpeg and the Ollama download). `--fail` turns an HTTP error into a
non-zero exit instead of saving an error page; the retries (including `--retry-all-errors`,
which also retries on HTTP 4xx/5xx) absorb the exact class of transient CDN blip described in
the issue.

### 2. Post-download binary checks (all platforms)

Added `assert_binary()` helper that verifies the extracted file exists and is not implausibly
small. Called after extraction on macOS (alongside the existing arch/version checks), Linux,
and Windows for both ffmpeg and Ollama. This closes the asymmetry where only the Darwin branch
had post-download verification.

The Ollama download now also locates the binary via `find` instead of assuming a fixed path,
and fails loudly if the extract produced nothing.

## Verification

- Script re-parsed locally with `bash -n` (syntax OK).
- `assert_binary` tested with three cases: valid file (pass), too-small file (fail), missing
  file (fail).
- Change is additive: all four call sites now have curl flags + post-download checks.

### DIFF
--- a/scripts/download-ollama.sh
+++ b/scripts/download-ollama.sh
@@ -7,6 +7,30 @@ set -e
 OLLAMA_VERSION="v0.31.1"
 BIN_DIR="$(cd "$(dirname "$0")/.." && pwd)/bin"
 
+# Assert an extracted artifact exists and is not implausibly small, then report it.
+# `curl --fail` only rejects an HTTP error status: a transfer that dies mid-flight
+# still exits 200 with a partial file, and `find ... -exec mv` exits 0 even when it
+# matched nothing. Both leave a bad or missing binary that only surfaces much later
+# in PyInstaller or at runtime. Size is the one post-condition that works on every
+# platform -- the Windows binary cannot be executed on the runner that produced it.
+assert_binary() {
+    local path="$1" min_bytes="$2" size
+    if [ ! -f "$path" ]; then
+        echo "Expected $path after extract, but it is missing" >&2
+        exit 1
+    fi
+    size=$(wc -c < "$path" | tr -d '[:space:]')
+    if [ "$size" -lt "$min_bytes" ]; then
+        echo "$path is only ${size} bytes (expected at least ${min_bytes}); download or extract was truncated" >&2
+        exit 1
+    fi
+}
+
+# Conservative floors: the real binaries are tens of MB, while an HTML error page
+# or a truncated transfer is orders of magnitude below these.
+MIN_FFMPEG_BYTES=5000000
+MIN_OLLAMA_BYTES=1000000
+
 # --- Download ffmpeg ---
 echo "=== Downloading ffmpeg ==="
 case "$(uname -s)" in
@@ -29,6 +53,7 @@ case "$(uname -s)" in
         unzip -o ffmpeg.zip ffmpeg
         rm ffmpeg.zip
         chmod +x ffmpeg
+        assert_binary ffmpeg "$MIN_FFMPEG_BYTES"
         # Validate the binary before bundling it. Two distinct failure modes:
         #  1. Wrong architecture. An x86_64 ffmpeg runs fine HERE under Rosetta and
         #     would sail through the -version check, then crash on a Rosetta-less
@@ -59,6 +84,7 @@ case "$(uname -s)" in
         tar -xf ffmpeg.tar.xz --strip-components=1 --wildcards '*/ffmpeg'
         rm ffmpeg.tar.xz
         chmod +x ffmpeg
+        assert_binary ffmpeg "$MIN_FFMPEG_BYTES"
         echo "ffmpeg downloaded"
         cd - > /dev/null
         ;;
@@ -73,6 +99,7 @@ case "$(uname -s)" in
         unzip -o ffmpeg.zip -d ffmpeg-extract > /dev/null
         find ffmpeg-extract -name 'ffmpeg.exe' -exec mv {} . \;
         rm -rf ffmpeg-extract ffmpeg.zip
+        assert_binary ffmpeg.exe "$MIN_FFMPEG_BYTES"
         echo "ffmpeg.exe downloaded"
         cd - > /dev/null
         ;;
@@ -122,6 +149,16 @@ else
 fi
 
 rm "$OLLAMA_FILE"
+
+# The Ollama archives differ in layout between platforms (some nest the binary under
+# bin/), so locate it rather than assuming a path, and fail if the extract produced
+# nothing at all.
+OLLAMA_BIN="$(find . -maxdepth 3 -type f \( -name ollama -o -name ollama.exe \) | head -n 1)"
+if [ -z "$OLLAMA_BIN" ]; then
+    echo "No ollama binary found under $BIN_DIR after extracting $OLLAMA_FILE" >&2
+    exit 1
+fi
+assert_binary "$OLLAMA_BIN" "$MIN_OLLAMA_BYTES"
 
 echo "Ollama downloaded to $BIN_DIR"
 ls -la "$BIN_DIR"
--- a/scripts/download-ollama.sh
+++ b/scripts/download-ollama.sh
@@ -141,10 +141,12 @@ cd "$BIN_DIR"
 # Download
 curl --fail --retry 3 --retry-delay 2 --retry-all-errors -L "$OLLAMA_URL" -o "$OLLAMA_FILE"
 
-# Remove stale ollama binaries from a previous run so the find below only
+# Backup existing ollama binaries so the find below only
 # matches the freshly extracted download. Same scope as the lookup below
-# (-maxdepth 3, both names): any path the lookup can hit is cleared first.
-find . -maxdepth 3 -type f \( -name ollama -o -name ollama.exe \) -delete
+# (-maxdepth 3, both names): any path the lookup can hit is backed up first.
+# Using mv instead of delete preserves the last-known-good binary if the new
+# download or extraction fails, allowing easy manual recovery.
+find . -maxdepth 3 -type f \( -name ollama -o -name ollama.exe \) -exec mv {} {}.bak \;
 
 # Extract based on file type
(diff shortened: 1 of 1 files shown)


Optic00 (commented):
## Verdict

Comment. The change is correct and I'd take it as-is, but it covers the first of the two fixes #443 asks for, so `Closes #443` currently overstates what merging it does.

## Summary

Routes all four downloads in `scripts/download-ollama.sh` through `curl --fail --retry 3 --retry-delay 2 -L`. That turns an HTTP error into a non-zero exit on the line that caused it, instead of saving the error page as the artifact and failing three steps later in `unzip`. Exactly the right first move, and thank you for picking it up.

I reviewed 6e1607d.

## Findings

### Important

**1. The retries don't cover the incident the issue describes (`scripts/download-ollama.sh`)**

`--retry` only re-attempts what curl considers transient: timeouts, 408, 429 and 5xx. A 404 is not in that set. The failure in #443 was a GitHub release redirect returning `Not Found` for an asset that was fine on the next run, so it is precisely a 404 that this won't retry.

Measured against the real URL:

```
curl --fail --retry 3 --retry-delay 2 -L <404>   ->  exits after 0s, no retry
curl --fail --retry 3 --retry-delay 2 --retry-all-errors -L <404>   ->  retries
```

So after this change that build still goes red, just with a message that names the cause. That is a genuine improvement and worth having on its own. If you want the blip to heal itself rather than only be legible, `--retry-all-errors` is the missing flag.

**2. The second half of the issue is still open (`scripts/download-ollama.sh`)**

#443 asks for two things, and the post-conditions aren't here yet: the Linux and Windows ffmpeg branches and the Ollama extract still have no check that the extract produced anything. The Windows one earns its keep twice over, because

```
find ffmpeg-extract -name 'ffmpeg.exe' -exec mv {} . \;
```

exits 0 even when it matched nothing, so a wrong-shaped archive leaves `bin/` empty and only surfaces later in PyInstaller or at runtime. The macOS branch has carried this guard all along; the others just never got it.

A size floor is enough and works everywhere, including for the Windows binary that can't be executed on the runner that produced it. Something like a small helper next to the existing macOS checks, asserting the file exists and is over a plausible minimum, applied to `ffmpeg`, `ffmpeg.exe` and `ollama`/`ollama.exe`.

Happy either way on this: add it here, or narrow this PR to `Refs #443` and leave the guards for a follow-up. I'd just rather the issue not close while half of it is still true.

### Praise

The flag choice is right and applied consistently to all four call sites rather than only the one that broke, which is the difference between fixing an incident and fixing a class of incident. It also happens to be blocking a PR of mine right now, where the Windows bundle job is red on exactly this download step, so this is more useful than its four-line diff suggests.

Optic00 (commented):
Thanks for the quick turnaround, and for picking `--retry-all-errors` specifically. That
was the load-bearing part: plain `--retry` only re-tries what curl considers transient
(timeouts, 408, 429, 5xx), and the incident in #443 was a 404 on a release redirect for an
asset that was fine seconds later. I measured that against the real URL before raising it -
with `--retry` alone curl gives up in 0 s, with `--retry-all-errors` it spends ~6 s over
three attempts. So this now fixes the reported failure.

One piece of the original comment is still open, and I would like to see it before this
goes in.

**The post-download checks are still macOS-only.** After extracting, the Darwin branch
asserts the binary is arm64 and that `./ffmpeg -version` reports 7.1.x, then fails loudly
if not. The Linux and Windows branches extract and move on with no check at all. That
asymmetry predates your PR, but this PR is the one hardening these downloads, so it seems
like the right place to close it rather than leaving Windows the only path that can bundle
a bad artifact silently.

`--fail` does not cover this case. It rejects an HTTP error status, but a transfer that
dies mid-flight still returns 200 with a partial file, and a truncated zip fails later in
`unzip` with "End-of-central-directory signature not found" - an error that says nothing
about the download and costs someone a log dive to reach "just re-run it". That is exactly
how #443 presented.

It does not need to be the macOS treatment. The Windows binary cannot be executed on the
runner that produced it anyway, so a cheap post-condition is enough: assert the extracted
file exists and is not implausibly small. A 9-byte error page and a missing file both fail;
a real 20 MB+ binary passes. Something like a small helper called after each extract, taking
a path and a minimum size, would cover all four downloads uniformly.

Happy to look again once that is in.

MeiSiristhebest:
Thanks for the thorough review, @Optic00.

**Post-download checks are now on all platforms.** Added `assert_binary()` which verifies the extracted file exists and is not implausibly small (conservative 5 MB floor for ffmpeg, 1 MB for ollama). It's called after extraction on macOS, Linux, and Windows — the macOS branch still keeps its existing `file`/`-version` checks on top.

`--fail` only rejects an HTTP error status, but a transfer that dies mid-flight still exits 200 with a partial file. `assert_binary` catches that case via the size floor. The Windows binary can't be executed on the runner that produced it, so size is the one post-condition that works universally.

The Ollama download now also uses `find` to locate the extracted binary (archives nest it differently per platform) and fails loudly if the extract produced nothing.

Verified locally: `bash -n` passes; `assert_binary` tested with valid file (pass), too-small file (fail), and missing file (fail).

Optic00 (dismissed):
This does exactly what was needed, and then some. Approving.

`assert_binary` on all three platforms closes the gap I raised, and the reasoning in your
comment is right: `--fail` only rejects an HTTP status, and a transfer that dies mid-flight
still exits 200 with a partial file. Catching that `find ... -exec mv` also exits 0 when it
matched nothing is a failure mode I had not spotted - that one would have produced a missing
binary with a green build.

Verified on my side: the four `curl` calls now carry `--retry-all-errors`, which is the flag
that actually covers the 404 in #443 (plain `--retry` gives up on it immediately - measured
against the real URL: 0 s versus ~6 s over three attempts). macOS keeps its `file`/`-version`
checks on top of the new size floor, so nothing regressed there.

One note for the record, not a change request: cubic flagged that `find ... | head -n 1`
validates only the first match while PyInstaller bundles the whole `bin/` tree, so in
principle an archive with two `ollama` entries could pass validation on one and ship
another. I checked both real layouts here (`bin/ollama` and
`dist/stenoai/_internal/ollama/ollama`) and each has exactly one match, and the worst case
would be the assertion not firing - i.e. the state before this PR, not a new failure. If a
future Ollama archive ever nests differently, asserting exactly one match would be the
robust form. Not worth another round now.

Thanks for two quick, careful turnarounds on this.

Optic00:
Correction to my approval note: you had already fixed the `find` point in `b393010` before I wrote it - `rm -f ollama ollama.exe` before the extract plus `-print -quit` is the right shape, and it removes the stale-file half of the concern entirely. My "not worth another round" was aimed at a problem you had already closed. Approval stands against that commit.

Optic00:
I approved too early - holding that until one more thing is fixed, and I have turned
auto-merge back off so this does not land in the meantime. Apologies for the back and forth;
this one is on me for not reading cubic's second comment before approving.

cubic's P2 on line 146 is correct, and it defeats the exact guarantee this PR is adding.

`rm -f ollama ollama.exe` only clears candidates in `$BIN_DIR` itself, but the `find` below
searches `-maxdepth 3` - and as your own comment notes, some platform archives nest the
binary under `bin/`. So a stale `$BIN_DIR/bin/ollama` from an earlier run survives the
cleanup, gets found, passes `assert_binary`, and the script prints "Ollama downloaded" for a
binary this run never downloaded.

The sequence, to be concrete about how narrow it is:

1. an earlier run leaves `$BIN_DIR/bin/ollama`, at least 1 MB
2. a later run downloads successfully and extracts without error, but the archive contents or
   layout changed so it yields no candidate at the expected place - `tar`/`unzip` still exit 0
3. `find` matches the old file, the size floor accepts it, and the build reports success

There is a milder version too: even when the new archive *does* supply a binary, `find`'s
ordering is not guaranteed, so it can validate the stale one instead of the fresh one.

**A fresh CI checkout never hits this** - it only bites local repeat runs, which is where a
contributor would be misled about their own build. So it is not a blocker in the sense of
"this makes CI wrong", but it does hollow out the post-condition on exactly the machines
where someone would trust it.

Two shapes that fix it, whichever you prefer:

- clear every candidate in the lookup scope before extracting, e.g.
  `find . -maxdepth 3 -type f \( -name ollama -o -name ollama.exe \) -delete`, or
- extract into a clean temporary directory and move the binary into place from there, which
  also makes the "did this archive supply it" question unambiguous

Sorry to send you round a third time. This is the last thing from me on this PR - everything
else, including the `--retry-all-errors` and the cross-platform `assert_binary`, is right and
I am not looking for more.

Optic00:
One thing to fix before I merge, and it is a one-liner: the file mode dropped from `100755`
to `100644` somewhere in this branch, so `scripts/download-ollama.sh` is no longer
executable.

CI does not catch it, because all four workflows that call the script run
`chmod +x scripts/download-ollama.sh` first. But `README.md:231` tells a new contributor to
run `./scripts/download-ollama.sh` directly, and on a fresh clone that now fails with
`Permission denied`.

```
git update-index --chmod=+x scripts/download-ollama.sh
```

plus a push is all it needs.

Everything else here reads right to me. The `mv` to `.bak` before the extract closes cubic's
P2 without the recoverability gap that deleting first would have left, the ordering holds in
all four failure paths, and the size floors are the correct post-condition given the Windows
binary cannot be executed on the runner that produced it. Sorry again for the approve and
un-approve earlier; this is the last thing from me.

Issue #359, opened 2026-07-21
Speaker diarization + cross-recording speaker ID


valentinweyer:
# Speaker diarization + cross-recording speaker ID

## Problem

Steno currently attributes speech per-channel only: everything on the mic side is "You," everything on system audio is "Others." Two real cases fall through that:

- **Multiple speakers sharing one side of a call**: an in-person conversation captured on a single mic, or several remote participants mixed into system audio, all get lumped into one label each.
- **No memory across recordings**: even if a channel is correctly split into distinct speakers, there's no way to know that "Speaker 2" in today's meeting is the same person as "Speaker 2" in last week's, so named participants never accumulate across a user's history.

Solving this properly needs both acoustic diarization (splitting a channel into distinct speakers) and a safe way to attach names to those speakers over time.

## What's already built (not a proposal, a working, tested branch)

This is fully implemented and validated locally. Opening this issue first because it's a large unsolicited change and I'd rather align on approach before a big PR lands in your inbox, but wanted to be upfront that the "proposal" is really "here's a finished thing, does the approach look right to you." **Happy to give a live showcase/screen-share of it running against real recordings**, whichever is more useful than reading this.

### Acoustic diarization

A macOS-only Swift/CoreML sidecar (`diarize-sidecar/`) wrapping FluidAudio's Sortformer diarizer, invoked from Python (`src.transcriber._run_steno_diarize`). Never from Electron, since the batch pipeline is entirely Python-orchestrated.

- **Per-channel**: mic and system-audio channels diarized independently, so multiple speakers sharing one side of a call get labelled "You" / "Speaker 2" / "Speaker 3" instead of lumped together. The dominant-by-duration cluster on each channel keeps the legacy "You"/"Others" label; other clusters are numbered by first chronological appearance across both channels. Any failure (missing binary, timeout, bad output, single-cluster result) falls back to today's exact channel-only behavior, so this can never fail a meeting.
- **Known model limit**: the underlying model (NVIDIA's `diar_streaming_sortformer_4spk-v2.1`, converted to CoreML by FluidAudio) is architecturally capped at 4 speakers per diarization pass. A channel with more than 4 distinct speakers will have some of them conflated rather than separated. Worth flagging plainly since it's a real ceiling on the model, not something the integration can work around.
[...]

PR #455, opened 2026-07-31
feat(diarize): per-channel acoustic speaker diarization (Sortformer)

valentinweyer:
## Description

Adds acoustic speaker diarization for both stereo (mic/system) and mono recordings, splitting multiple speakers sharing one side of a call or a single mic into distinct labeled turns instead of the current flat "You"/"Others" split. This picks up issue #359, scoped down to diarization only. A separate follow-up PR covers human-confirmed cross-recording speaker identification (naming a "Speaker 2" from a past meeting), which depends on this one and is a substantially larger review surface (persistent identity storage, matching quality caveats) that deserves its own slower pass.

Diarization runs through a new Swift/CoreML sidecar (`diarize-sidecar/`) wrapping FluidAudio's Sortformer diarizer, invoked from Python (`src.transcriber._run_steno_diarize`), never from Electron, since the batch pipeline is entirely Python-orchestrated. It is macOS-only and gated in `stenoai.spec` on both the platform and the sidecar binary existing, so a checkout that skips the Swift build (or Windows/Linux) falls back to the existing channel-only "You"/"Others" labeling automatically. Diarization can never fail a meeting: any failure (missing binary, timeout, bad output, a single-cluster result) degrades to the legacy behavior byte-for-byte.

Also adds a live per-stage progress indicator ("Identifying speakers" with an elapsed-time counter) so a long diarization pass on a multi-hour recording no longer sits on a static "Analyzing transcript" spinner with no feedback, and a heartbeat mechanism so Electron's inactivity watchdog does not kill a long-running diarization call.

## Type of Change

- [x] New feature

## Testing

- 538 backend unit tests passing (`python -m unittest discover tests`), including new coverage for the sidecar's JSON parsing (real-world quirks: FluidAudio/CoreML warning text before, between, or after the payload; a dominant-speaker gate to avoid spawning phantom speakers from misdiarization blips) and the progress/heartbeat plumbing.
- `ruff check` clean on every file this PR touches.
- Renderer `tsc --noEmit` clean, `eslint` clean (0 errors) on the new Processing stage.
- Full T1 e2e tier (53 specs) passing, including a new `processing-stages.t1.spec.ts` covering the stage-transition logic end to end.
- `speaker-diarization.t2.spec.ts`: synthesizes real speech via macOS `say` so Parakeet/whisper.cpp produce real ASR segments, points the sidecar at a fixture script, and asserts the saved transcript's per-channel labeling and cross-channel speaker numbering.
- Beyond the automated suite, validated directly against real multi-hour, multi-speaker recordings on real hardware: built the sidecar, ran it on real audio, and confirmed the Python parser handles the actual production output (including the FluidAudio warning-prefix case) correctly end to end. A full DMG build was tested from a cold launch (no background services running) to confirm the full record/transcribe/diarize/summarize pipeline works in the packaged app.

## Additional Notes

Sortformer has a fixed 4-speaker-slot architecture (no speaker-count hint can be passed to it). This is a known limitation flagged in #359 and its discussion; the identity follow-up PR's discussion covers a possible "who is in this meeting" field as a future lever, which is out of scope here.

The rendered transcript UI (chat bubbles) itself is unchanged by this PR, and it already benefits directly: when a channel has two diarized speakers, e.g. two people sharing one mic with no separate system-audio channel, both show up as two visually distinct bubbles (the dominant speaker as "You", green/right; the other as grey/left), where today they would be merged into one "You" bubble. That per-channel split is the actual feature.

Its limit is at three-plus: the bubble styling only has two visual buckets ("You" vs everyone else), so if diarization finds a third or fourth distinct speaker, e.g. three people sharing one mic, or extra speakers split across both channels, the transcript text still correctly carries distinct "Speaker 2"/"Speaker 3"/"Speaker 4" labels, but the bubbles do not visually differentiate them from each other, or from "Others" — all render identically grey/left, unless a speaker has been given a confirmed real name (the identity follow-up PR). That richer per-speaker label data is already useful today regardless (readable in the raw transcript text/export, and it's what the identity PR's suggestion engine keys off of); giving 3+ distinct speakers their own visual treatment in the bubble UI is a reasonable follow-up, not something this PR does.

---
## Summary by cubic
Adds acoustic speaker diarization for stereo and mono recordings on macOS via `diarize-sidecar` (Sortformer), labeling turns as “You” and “Speaker N” instead of channel-only. Adds an “Identifying speakers” stage with an elapsed timer; diarization is optional and always falls back to legacy labels if unavailable or low-confidence.

- **New Features**
  - macOS-only `diarize-sidecar` (Swift/CoreML) using `FluidAudio` Sortformer; invoked from Python and bundled when present, otherwise auto-fallback to “You”/“Others”.
  - Per-channel diarization on stereo and mono: the dominant speaker per channel is “You”; additional speakers are labeled “Speaker N” by first appearance across channels; transcript parser and UI render these labels (up to four voices per channel).
  - Live diarization progress via `PROGRESS:diarize:*` and a heartbeat; new “Identifying speakers” stage with an elapsed time counter; progress lines are persisted in the processing log.

- **Bug Fixes**
  - Diarization now counts as “on” only when multiple speaker labels are present, so empty second channels no longer drop labeled transcripts.
  - Long ASR sentences are split and reassigned at the word level to match diarizer turns, improving speaker alignment.
  - Fixed stage transitions and progress handling: ignore `PROGRESS:diarize:*` for summarization, clear stale sub-labels on transitions, and persist diarize progress markers.
  - Processing-stage tests stabilized by waiting for the session header before emitting events and launching with fakeAudio on CI runners without audio devices.
  - UI now treats only exact “You” (or no marker) as self; “Speaker N” and “Others” render on the left.

<sup>Written for commit 05679be9b3538d3048e42029d12f801aff8ce3b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/stenolabs/stenoai/pull/455?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -199,3 +199,8 @@ ollama.pid
 
 # Local design reference library (cloned, not vendored)
 .reference/
+
+# Swift Package Manager build artifacts for the diarize-sidecar helper
+# (fetched FluidAudio checkout + compiled binaries) — scripts/build-diarize-sidecar.sh
+# regenerates this; only diarize-sidecar/{Package.swift,Package.resolved,Sources/} are tracked.
+diarize-sidecar/.build/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@ The app is a thin Electron shell over a PyInstaller-bundled Python CLI. There is
 - **Python CLI (`simple_recorder.py`, ~2.9k lines, ~60 click commands)** is the single entry point bundled by `stenoai.spec`. Sub-modules in `src/`: `audio_recorder` (sounddevice), `transcriber` (pywhispercpp), `summarizer` (Ollama HTTP client), `ollama_manager` (lifecycle of the bundled `ollama serve`), `config` (JSON-backed user settings + model registry), `folders`, `models`, `whisper_models`.
 - **State across CLI invocations** is persisted to `recorder_state.json` and similar small JSON files — there is no daemon. Long-running recordings are a `record` subprocess kept alive by the Electron main process.
 - **User data lives in `~/Library/Application Support/stenoai/`** (`recordings/`, `transcripts/`, `output/`), resolved via `src.config.get_data_dirs()`. Repo-root `recordings/`/`transcripts/`/`output/` dirs are dev-only scratch.
-- **Bundled binaries (`bin/`)**: Ollama + ffmpeg, downloaded by `scripts/download-ollama.sh`. PyInstaller copies them into `dist/stenoai/ollama/` and `dist/stenoai/ffmpeg`. Electron then re-bundles `dist/stenoai/` as an `extraResource`.
+- **Bundled binaries (`bin/`)**: Ollama + ffmpeg, downloaded by `scripts/download-ollama.sh`. PyInstaller copies them into `dist/stenoai/ollama/` and `dist/stenoai/ffmpeg`. Electron then re-bundles `dist/stenoai/` as an `extraResource`. `bin/steno-diarize` (macOS only) is a separate Swift/CoreML sidecar built by `scripts/build-diarize-sidecar.sh` — see "Speaker diarization" below.
 - **Deep links**: app registers the `stenoai://` URL scheme. Handler logic is in `app/main.js` near `SHORTCUT_PROTOCOL`. Used by macOS Shortcuts: `stenoai://record/start?name=...` and `stenoai://record/stop`.
 
 ## Development Commands
@@ -38,6 +38,28 @@ The app is a thin Electron shell over a PyInstaller-bundled Python CLI. There is
 The Electron build pulls the bundled backend from `../dist/stenoai` via `extraResources`, so the PyInstaller step (`pyinstaller stenoai.spec --noconfirm`) must succeed *before* `npm run build` — otherwise the packaged app will be missing `stenoai`, `ollama`, and `ffmpeg`. The same applies in dev: `getBackendPath()` falls back to `dist/stenoai/stenoai`, so a fresh checkout needs the backend built once before the app can record or transcribe.
 
 For setup from a clean checkout, see `CONTRIBUTING.md` and `README.md`.
+
+### Speaker diarization (macOS only)
+Per-channel acoustic speaker diarization (splitting multiple speakers sharing
+one side of a call — e.g. two people around one mic, or multiple remote
+participants on system audio) runs through `bin/steno-diarize`, a Swift/
+CoreML sidecar (`diarize-sidecar/`) wrapping FluidAudio's Sortformer
+diarizer, invoked from Python (`src.transcriber._run_steno_diarize`) — never
+from Electron, since the batch pipeline is entirely Python-orchestrated.
+Build it *before* `pyinstaller stenoai.spec`, same as `download-ollama.sh`:
+
+```
+scripts/build-diarize-sidecar.sh   # outputs bin/steno-diarize
+scripts/download-ollama.sh
+pyinstaller stenoai.spec --noconfirm
+```
+
+`stenoai.spec` bundles `bin/steno-diarize` only when it exists and only on
+macOS (`_IS_DARWIN`), so skipping the build step is safe — the app falls
+back to the legacy channel-only "You"/"Others" labeling whenever the binary
+or a diarization run is unavailable (missing binary, timeout, bad output);
+this never fails a meeting. Windows/Linux never get acoustic diarization —
+`_resolve_steno_diarize()` returns `None` immediately off-darwin.
 
 ### End-to-end tests (Playwright)
 The e2e suite drives the **real Electron app** (real window, real clicks) to catch
@@ -97,7 +119,14 @@ overrides an agent's own test-level defaults.
     `setup-check.t2` (the setup-wizard allGood + checks contract) (all model-free,
     run in `t2-macos` /
     `t2-windows`); `transcription-pipeline.t2` and `honest-failure.t2` (tagged
-    `@pipeline`, run in `t2-pipeline-macos` / `t2-pipeline-windows`). Engine selection
+    `@pipeline`, run in `t2-pipeline-macos` / `t2-pipeline-windows`), and
+    `speaker-diarization.t2` (also `@pipeline`, macOS-only — skips loudly off-darwin
+    and when macOS's `say` TTS is unavailable — synthesizes real speech via `say` so
+    Parakeet/whisper.cpp produce real ASR segments, points
+    `STENOAI_DIARIZE_SIDECAR_PATH` at a fixture script returning fixed 2-speaker JSON,
+    and asserts the saved transcript's per-channel "You"/"Speaker 2"/"Others" labeling
+    and cross-channel numbering; the real `steno-diarize`/Sortformer binary itself was
+    validated directly against real recordings rather than in CI). Engine selection
     for `@pipeline` specs is shared via `e2e/fixtures/engine.ts`; model-free T2 setup
     helpers (deterministic recording config + seeded meeting summaries) live in
     `e2e/fixtures/user-config.ts`. The core-loop specs drive the preload IPC bridge and
--- a/app/main.js
+++ b/app/main.js
@@ -7098,6 +7098,13 @@ function logPipelineStdoutLine(line, source) {
     processingLog.logLine(source, l);
     return;
   }
+  if (l.startsWith('PROGRESS:diarize:')) {
+    // Only :start/:done markers exist on this pipeline (rare, at most twice
+    // per channel) -- no per-chunk flood risk, so no throttle needed unlike
+    // HEARTBEAT above.
+    processingLog.logLine(source, l);
+    return;
+  }
   if (
     l.startsWith('TRANSCRIPTION_COMPLETE') ||
     l.startsWith('TRANSCRIPTION_FAILED') ||
--- a/app/renderer/src/components/TranscriptPanel.tsx
+++ b/app/renderer/src/components/TranscriptPanel.tsx
@@ -133,8 +133,10 @@ function TranscriptRow({ segment, highlight }: { segment: Segment; highlight: st
   // showing their content as un-bubbled plain text (or worse, on the
   // "Others" side) reads as wrong. Granola takes the same charitable
   // default — when in doubt, attribute to the mic owner. Explicit
-  // `Others` markers still render as grey/left.
-  const isYou = segment.speaker !== 'Others';
+  // `Others` markers and acoustically-diarized `Speaker N` markers
+  // (transcriber.py's _resolve_speaker_placeholders) both render as
+  // grey/left — only an exact "You" (or no marker at all) is "self".
+  const isYou = segment.speaker === 'You' || segment.speaker == null;
   return (
     <div className={cn('flex flex-col gap-0.5 px-1 py-0.5', isYou ? 'items-end' : 'items-start')}>
       {segment.timestamp && (
--- a/app/renderer/src/lib/transcriptSegments.test.ts
+++ b/app/renderer/src/lib/transcriptSegments.test.ts
@@ -40,6 +40,18 @@ describe('parseTranscript — diarised', () => {
     expect(segs[0].text).toContain('line two');
     expect(segs[1].timestamp).toBe('00:05');
   });
+
+  test('parses acoustically-diarized "Speaker N" labels alongside You/Others', () => {
+    const segs = parseTranscript(
+      '[00:00] [You] Hello\n\n[00:05] [Speaker 2] Hi there\n\n[00:10] [Others] Welcome',
+      true,
+    );
+    expect(segs).toEqual([
+      { speaker: 'You', text: 'Hello', timestamp: '00:00' },
+      { speaker: 'Speaker 2', text: 'Hi there', timestamp: '00:05' },
+      { speaker: 'Others', text: 'Welcome', timestamp: '00:10' },
+    ]);
+  });
 });
 
 describe('parseTranscript — non-diarised', () => {
--- a/app/renderer/src/lib/transcriptSegments.ts
+++ b/app/renderer/src/lib/transcriptSegments.ts
@@ -3,7 +3,10 @@
 // unit-tested in isolation, and so the live dock could share it later.
 
 export interface Segment {
-  speaker: 'You' | 'Others' | null;
+  /** 'You' / 'Others' for the channel-only fallback, or 'Speaker N' for an
+   *  acoustically-diarized secondary speaker on a channel (see
+   *  transcriber.py's _resolve_speaker_placeholders). */
+  speaker: string | null;
   text: string;
   /** `MM:SS` / `H:MM:SS` offset parsed from a diarised line's leading
    *  `[MM:SS]` marker (transcriber.py writes it). Absent on older transcripts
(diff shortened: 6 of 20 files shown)


Optic00 (commented):
I ran this branch's sidecar against an independent diarization benchmark rather than reading
the diff alone. Measured on one M5 Air, scored with dscore at a 0.25 s collar against the AMI
reference RTTMs from `BUTSpeechFIT/AMI-diarization-setup`, on the **test** split (16 meetings,
9.06 h, `Array1-01`, one distant room microphone) - the closest public stand-in for what this
PR targets, several people sharing one channel.

## It works, and the margin is large

| system | DER | JER |
|---|---|---|
| all remote speech under one label (today's system-audio channel) | 71.09 | 89.82 |
| **this branch's sidecar** | **29.24** | **39.46** |

Throughput ~105x realtime, about 35 s for a one-hour meeting. This is strong evidence that
the PR materially improves diarization on the case it targets. (FluidAudio's own CLI scored
33.55 on the same material, but that run used a different library version and a likely
different code path, so I would not read anything into the gap either way.)

To be precise about the baseline: it is a one-label system scored on single-channel audio,
so it stands in for the channel where Steno is blind today, not for the whole app - the mic
channel already gets "You" right by construction.

## Two claims in the code comments I could not reproduce

**The `.highContextV2` switch made no measurable difference here.** I rebuilt the sidecar with
`sortformerHighContextMinDuration` raised so `.default` is always chosen, and reran all 16
meetings: DER identical to two decimals, wall-clock identical too.

AMI meetings are 14-40 minutes, so this does not contradict the PR body's "~6.6x faster" if
that was measured on the multi-hour recordings you tested against. If the gain only appears
past some length, that would explain why I could not reproduce it on shorter material. Not
asking for a change.

**Forcing `.cpuAndNeuralEngine` costs a little more than the comment implies.** With
`STENOAI_DIARIZE_COMPUTE_UNITS=all`: DER 28.89 (vs 29.24) and RTFx 145 (vs 105), so GPU is
marginally more accurate and ~1.4x faster. For the live-recording path the power-first
default still looks right, and the env-var escape hatch is already there. Just recording the
numbers, since the comment reasons about the trade without them.

## The phantom-speaker case, which came out better than I expected

`EN2002c` has 3 reference speakers and both this sidecar and upstream report 4 - the model
fills its slots whether the people exist or not. I also built a 2-speaker fixture (28.8 s,
two macOS `say` voices, distinctness verified spectrally first) where 0.48 s at the first
speaker change goes to a third cluster holding 1.7 % of speech.
`CHANNEL_DOMINANCE_THRESHOLD = 0.92` does not catch it, since the dominant cluster there
holds only 57 %.

**It does not reach the user.** I drove `transcribe_diarised` on that file through the real
Parakeet path: the output carries exactly `[You]` and `[Speaker 2]`, both correct, and no
`Speaker 3`. The phantom overlaps only a 1.44 s ASR segment, stays under
`LONG_SENTENCE_SPLIT_THRESHOLD_S`, is assigned whole to the nearest real cluster, then
dropped for carrying no text. So the dominance gate is not the only thing protecting you, and
the design holds under a case built specifically to break it. I had drafted a suggestion for
an extra minimum-cluster guard and dropped it - it would buy nothing observable and would
misattribute the participant who says "approved" once in an hour.

One unrelated thing visible in that same run: a 5.92 s sentence crossing a real speaker
boundary triggers word-level splitting, and the trailing full stop lands past the boundary,
so it opens the next turn (`[Speaker 2] ... around the hiring plan` / `[You] . Sure, let me
pull that up.`). Cosmetic, and only for sentences past the 5 s threshold that straddle a
turn change.

## Scope: shipping diarization before identity looks right to me

Today a three-person call renders every remote utterance as one undifferentiated `Others`.
Anonymous `Speaker 1/2/3` is the normal first rung, turn structure is what makes such a
transcript readable, and the fallback means any failure lands exactly where users are now. I
do not think this needs the identity PR to be worth merging.

One caveat worth stating now rather than after: anonymous labels pay off **in the transcript**.
In summaries, search and action items, "Speaker 2 owns the follow-up" has no stable referent.
That is an argument for not letting the identity PR sit, not against merging this one.
Related: the summariser now carries speaker attributions it did not have before, and a
confidently wrong attribution is a failure class `Others` could not produce. Nothing I
measured suggests it is likely, but it is what I would watch first in front of users.

Since the cap will shape that follow-up, I put the measurements in **#359** rather than here
- you flagged the 4-speaker limit there yourself, and the short version is that a supplied
participant count is worth real DER, which makes it a cheap field on a form that is going to
ask about speakers anyway. Nothing in it argues against this PR's backend choice; if
anything it vindicates it for a pipeline that has to estimate.

## Caveats

**No German, and no non-English validation at all** - my German fixtures turned out to be
degenerate (one synthetic voice recorded four times), so they measure nothing. AMI is
meeting-room audio, not VoIP system audio. I would not extrapolate 29.24 to either.

## Verified in passing

The stdout contamination handling is right, and worth saying because it looks like
over-engineering until you hit it: my run produced exactly the `E5RT encountered an STL
exception. msg = unordered_map::at: key not found.` line ahead of the JSON, with clean
stderr. Scanning every `[` with a real decoder and keeping the last segment-shaped array is
what makes that survivable.

Also: keep the `Package.resolved` pin. The newer FluidAudio I compared against scored worse
on this material, so whoever lifts it should measure first.

Happy to share the run tables, the control runs, or the harness. Nice piece of work - the
fallback discipline in particular is what makes this reviewable as a single change.

Optic00:
One concrete follow-up to my review above, which I would rather name precisely than leave as
"the cap should be documented somewhere".

**This PR makes an existing FAQ answer untrue.** `docs/faq.mdx`, under "Can Steno record
in-person meetings?":

> In-person recordings are mic-only, so they have no `[You]` / `[Others]` speaker labels.

After this merges that is the exact case that gains speaker labels - the mono path is the
headline of the feature. Left as is, the docs tell the in-person user the thing they are
looking for does not exist.

It is also the natural home for the four-slot limit, which currently lives only in the PR's
Additional Notes. A replacement that covers both, worded per channel since that is the real
constraint:

> Yes. Steno records from your Mac's microphone, which will pick up voices in the room. For
> best results, place your laptop centrally. Steno separates up to four distinct voices per
> recording channel, so an in-person meeting of four or fewer people gets per-speaker labels;
> beyond that, additional speakers are merged into the four it detects.

Two reasons for putting it in this PR rather than an issue: it ships with the feature that
causes it, and "up to four per channel" is a good deal less restrictive than a bare "four
speakers" - for a call, that is four people on the far side plus you, so most users never
meet the limit and should not be told otherwise.

Not a blocker from me either way. If you would rather keep this PR code-only, say so and I
will open it as a docs issue against the merge instead.

Optic00:
Heads-up on CI, since I do not think this one has been looked at: `T2 pipeline (transcribe +
summarize)` is red on the current head (`f31f2c9`), and it is your own new spec. It failed
three times including both retries, so it is not a flake. Windows is green only because the
spec skips there on `process.platform !== 'darwin'`.

**The line that fails**, `e2e/specs/speaker-diarization.t2.spec.ts:67`:

```
Expected: < 0.5053125
Received:   1.5
```

```js
const tailSeconds = Math.max(1.5, Math.min(3.0, micBoundarySeconds * 0.6));
expect(tailSeconds).toBeLessThan(micBoundarySeconds);
```

The floor and the invariant contradict each other: `Math.max(1.5, ...)` guarantees
`tailSeconds >= 1.5`, so the assertion cannot hold for any `micBoundarySeconds <= 1.5`. On
your machine the boundary is comfortably above that, which is why it passes locally.

**But the boundary being 0.5 s is the finding that matters.** `makeStereoSpeechWav` returns
`durA + gapSeconds / 2`, so with the 1 s gap the runner produced `durA = 5.3 ms` of audio for

> Hello there, I hope you are having a wonderful and productive day today.

I ran the same `say -o out.wav --data-format=LEI16@16000` call on my own Mac for comparison:
**3.796 s** for that sentence, 0.961 s for "Thanks a lot." So `say` on the GitHub-hosted
macOS runner is writing essentially nothing, roughly 170 bytes of PCM. My guess is that the
runner image ships the `say` binary without the voice assets, but I have not confirmed that
and it does not really matter for the fix.

What does matter is that `isSayAvailable()` cannot see it: it probes `say -v ?`, which exits
0 on the runner, so the guard stays green while the fixture is empty.

**So fixing the formula alone would make the spec pass without testing anything** - the WAV
would carry no speech, ASR would produce no sentence segments, and there would be nothing
for the per-channel labeling to work on. The honest fix is a duration guard on what `say`
actually wrote (skip loudly below, say, 1 s of audio, the way the spec already skips on a
missing model) plus a `tailSeconds` expression that cannot exceed the boundary, e.g.
`Math.min(micBoundarySeconds * 0.6, ...)` with no floor above it.

None of this touches the substance of the PR, which I still think is the right backend
choice - see the benchmark in my review above.

Optic00:
Heads-up before you rebase this onto current `main`: `processing-stages.t1` goes red, 2 of 3 tests. It is not your spec and not main alone - the spec only exists here, the queue watchdog only exists on main, and they meet for the first time in that merge.

Mechanism, read off the screen rather than inferred: after `processing-complete {success:false}` the label shows `Analyzing transcript` with `data-stage="transcribing"` instead of the error panel. `Processing.tsx` bumps `generation` when `activeSession` flips from null to the session name (or `recording.status` reaches `recording`), and the reset block then calls `setStage('transcribing')` during render. That runs after your error transition and undoes it.

Worth a look beyond the test: a genuinely failed recording whose session identity settles late would show `Analyzing transcript` instead of the failure.

Optic00:
Retargeted the base to `feat/speaker-diarization` — thanks for the quick rebase, it merges cleanly.

One red check left: **T1**, `processing-stages.t1.spec.ts:95`, 3 of 3 attempts in CI. I reproduced it locally on `e129c0a7`, where it trips one assertion earlier (112 rather than 113 — the window is sub-second).

The error panel never renders: after `processing-complete {success: false}` the screen still shows the "Analyzing transcript" stage card.

**Mechanism, and it isn't your code.** `processingComplete` sets `setStage('error')` (`Processing.tsx:225`). The same event also reaches the app-level listener mounted at `App.tsx:89`, which invalidates the queue query (`useRecording.ts:500`). That refetch is the first to observe the recording `openProcessing()` starts at `spec:26` and never stops — the mock reports `hasRecording: true` and `sessionName: 'test-session'` (`e2e-mock-ipc.js:296, 301`). It flips both signals `generation` folds together (`Processing.tsx:124-132`), so the render-phase reset block (`:143-152`) runs one IPC round-trip *after* the error was set and clears it again: `setStage('transcribing')` (`:145`), `setRetryAudioFile(null)` (`:150`). With `retryAudioFile` gone, `canRetry` is false (`:445`), so there is no usable retry button either.

Evidence rather than reading: patching only `:145` to `setStage((s) => (s === 'error' ? s : 'transcribing'))` makes 112 and 113 pass and moves the failure to line 120, the `Try again` click — which is exactly what `:150` predicts.

I diffed the branch against its merge base `4e6fcb52`: inside that handler your change is `clearDiarizeTimer()` + `setChunkProgress(null)` (`:223-224`). `setStage('error')` and the whole generation/reset machinery (`:107-152`) are untouched `main` code. So this is a new spec meeting pre-existing behaviour, not a regression you introduced.

I'd fix it on the spec side rather than in production. The reset exists so a screen that stays mounted across back-to-back recordings repaints fresh (`:107-118`); letting a terminal `error` stage survive the bump would leave a *new* session staring at the previous one's error panel. The cheapest fix is to let the generation settle before emitting anything — for instance have `openProcessing()` wait until the session name is painted in the header (`displayTitle`, `:395`, rendered at `:423`), since that implies the bump has already been consumed. Stopping the recording before the failure emit is also closer to how a user reaches this screen, and still leaves `sessionName` truthy via `currentJob` for `canRetry`. I have not tried either, so treat both as suggestions rather than a tested patch.

Everything else is green: 11 checks, including both Windows jobs and both pipeline runs.

Optic00:
Here's the spec-side fix, tested rather than suggested this time — on `e129c0a7`, all three tests in the file, three repeats each, 9/9 green:

```diff
--- a/e2e/specs/processing-stages.t1.spec.ts
+++ b/e2e/specs/processing-stages.t1.spec.ts
@@ -28,6 +28,11 @@ async function openProcessing(page: Page) {
     window.location.hash = '/meetings/processing';
   });
   await expect(page.getByTestId('processing-stage-label')).toBeVisible();
+  // The queue poll that first reports this recording bumps Processing's
+  // `generation`, whose render-phase reset clears stage + retryAudioFile.
+  // Wait until that has landed (the header switches from 'Note' to the
+  // session name) so a later emit can't race the reset.
+  await expect(page.getByRole('heading', { name: 'test-session' })).toBeVisible();
 }
```

Why it settles the race: the header is `<h1>{displayTitle}</h1>` with `displayTitle = … ?? activeSession ?? 'Note'` (`Processing.tsx:395, 423`). Once the session name is painted, the generation bump has been through a render and the reset can no longer clear anything a later emit sets. No assertion gets weaker — the error panel and the retry button are still asserted exactly as you wrote them.

Yours to take or leave, it's your PR. One weaker observation while I was in there: `:51` emits the first progress line right after the visibility check, which doesn't strictly guarantee the IPC listeners are mounted. I did not see it fail, so I'd leave it alone unless it ever goes flaky.

Optic00:
Correction on the patch I posted above: I verified it on macOS only, and it made things worse. Before it, T1 failed on one test; with it (`2060434f`) tests 1 and 2 fail instead, with the emitted progress event having no effect at all. Please feel free to revert that line — it isn't the fix, and I shouldn't have called it tested when the environment it fails in wasn't part of the test.

The lead I'd chase instead: `openProcessing()` starts a real recording, but all three `launchApp` calls omit `fakeAudio: true`. Every other spec in the suite that records sets it, including `notification-navigation.t1` on main. On a runner with no audio device `getUserMedia` rejects, and the resulting failure path opens a second window — which would make `emit()`'s `BrowserWindow.getAllWindows()[0]` the wrong target from that point on. That is a permanent misroute rather than a race, which is exactly what "five seconds of resending changed nothing" looks like, and it fits the macOS/Ubuntu split.

I am not asserting that yet. I have it running on our own fork right now with `fakeAudio: true` plus a probe that logs the window list at every emit, so one run distinguishes it either way, and I'll report the result here whichever way it comes out.

Optic00:
Result, as promised. It is `fakeAudio`, and it's the only thing missing.

I took your current head (`33369815`, i.e. after your `emitUntil` revert) and changed nothing but the three `launchApp` calls:

```diff
-  const { app, page } = await launchApp({ mockIpc: true });
+  const { app, page } = await launchApp({ mockIpc: true, fakeAudio: true });
```

On the same Ubuntu runner the whole T1 suite goes green — 68 passed, including all three tests in this file:

```
✓ processing-stages.t1.spec.ts:48  walks transcribing -> diarizing -> summarizing -> finalizing …
✓ processing-stages.t1.spec.ts:80  shows a ticking elapsed-time counter throughout diarization …
✓ processing-stages.t1.spec.ts:100 a processing failure swaps to the error panel …
```

Why: `openProcessing()` starts a real recording, and the CI runner has no audio device, so `getUserMedia` rejects and the capture-failure path runs instead. Every other spec in the suite that records sets `fakeAudio` for exactly this reason — the fixture starts Chromium with `--use-fake-device-for-media-stream`. It's the same class of failure the live-transcript specs hit back in July, which is why it only ever showed up in CI and never locally.

Your revert of `emitUntil` was right — with the flag set, the plain `emit` calls are enough. The `openProcessing` wait can stay or go; it is green either way, and it does prevent the error-panel race on its own.

Evidence, in case you want to look: [run on our fork](https://github.com/Optic00/stenoai/actions/runs/30852846978), branch built from your head with that one change.

Optic00:
As agreed, I've pushed the fix onto this branch rather than leaving it waiting: `05679be9`, the three `fakeAudio: true` calls and nothing else. It's an additive commit on top of your head, so it's a plain fast-forward for you — and if you'd rather solve it differently, revert it without asking, no hard feelings.

I've also taken it out of draft. The draft status was my own suggestion from back when this targeted `main`; now that it targets `feat/speaker-diarization`, which exists specifically for this feature, that reason is gone.

Once CI is green I'll merge it into `feat/speaker-diarization` as a merge commit, not a squash, so your individual commits and their authorship stay intact.

That should also clean up #472 for you: with the diarization commits in the base branch, merging `feat/speaker-diarization` into `speaker-identity` will collapse that diff from +12k down to the identity work alone, and it picks up this test fix on the way. The one thing left there afterwards is `speaker-review.t1:177`, which expects a button labelled "Stop sample" and finds "Play sample" — that's a question about your feature's intent rather than test plumbing, so I've left it for you.

PR #453, opened 2026-07-31
docs: reflect Windows 10/11 (alpha) support across docs + README

MeiSiristhebest:
## Summary
Closes #433. The docs site still says Steno is "macOS only" / "no Windows version", while the README (`README.md:197`), the website CTA footer, and CI (`build-windows.yml` + a Windows T2 e2e lane) already ship and smoke-test the Windows alpha.

This makes the docs consistent with that reality using the wording suggested in the issue: **macOS (stable) and Windows 10/11 (alpha)**.

## Changes
- `docs/faq.mdx` — accordion answer now states Windows 10/11 (x64) is supported in alpha; no iPhone/web version yet.
- `docs/pricing.md` — Platform line lists macOS + Windows (alpha).
- `docs/llms.txt` — description + Platform line list macOS + Windows (alpha).
- `README.md` — Future Roadmap entry updated: Windows alpha already ships; remaining work is GA hardening.
- `docs/compare/steno-vs-*.mdx` (7 pages) — Steno Platform column updated from "macOS only" to
  "macOS (stable), Windows 10/11 (alpha)"; Otter "Mobile app" row corrected to "No (desktop only)";
  Granola copy updated to "macOS (stable) with a Windows 10/11 (alpha) build". Third-party competitor
  rows that describe another app's own platform (e.g. MacWhisper "macOS only") are left untouched.

## Notes
- Cosmetic doc-only change; no code or build impact.
- The comparison pages were folded into this PR after all: their wording ("macOS (stable), Windows
  10/11 (alpha)") matches the rest of the docs and the issue's intent. MacWhisper's competitor row
  stays "macOS only" because it describes that app's own platform support, not Steno's.

---
## Summary by cubic
Update docs across FAQ, pricing, `docs/llms.txt`, README, and all comparison pages to reflect macOS (stable) and Windows 10/11 (x64) alpha support, replacing outdated “macOS only” text. Aligns platform messaging site‑wide with the current release and CI, and updates comparison tables/copy to show Windows (alpha) availability.

<sup>Written for commit cab32c5a5b063dca729dc89a3ba0f6f3128fbe3f. Summary will update on new commits.</sup>

[...]

PR #457, opened 2026-08-01
docs: add Windows data location and unsigned installer documentation

MeiSiristhebest:
## Summary

Follow-up to #453. Adds the two documentation items @Optic00 mentioned, all in `docs/faq.mdx`:

1. **Windows data location**: `%APPDATA%\stenoai\` is now documented in five FAQ entries — data storage, backup, and uninstall (first commit), plus the export-notes and disk-space answers (latest commit). Every path mention in the FAQ is now dual-platform.

2. **Unsigned installer note**: a new FAQ entry ("Why does Windows SmartScreen say Steno is an unrecognized app?") explains that the alpha is unsigned and how to bypass SmartScreen. (The README already covered this; the gap was the FAQ.)

## Files changed

- `docs/faq.mdx` — five accordions now include the Windows `%APPDATA%\stenoai\` path alongside the macOS path; added a new SmartScreen accordion.

## Notes

- No code changes, documentation only.
- As @Optic00 said: "yours if you fancy, otherwise I'll pick them up." Picking them up 🙂
- Per review: the intended README change was a no-op against what was already there (SmartScreen is documented in the README since the Windows alpha landed), so this PR is FAQ-only.

---
## Summary by cubic
Add Windows-specific docs: data lives in `%APPDATA%\stenoai\`, now noted in storage, backup, and uninstall FAQs. Add a new FAQ explaining Windows SmartScreen warnings for the unsigned alpha and how to proceed.

<sup>Written for commit bba6b29ab5ffb97fce475aae61b49f8e50287873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ruzin/stenoai/pull/457?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/faq.mdx
+++ b/docs/faq.mdx
@@ -18,7 +18,7 @@ Steno runs on macOS (Apple Silicon, M1 and later -- Intel Macs are not supported
 </Accordion>
 
 <Accordion title="Can I export my notes to Notion, Obsidian, or other tools?">
-Steno saves your notes as plain Markdown files under `~/Library/Application Support/stenoai/output/`. Because they are ordinary files, you can move them into Notion, Obsidian, or any other tool by hand or through your own automation.
+Steno saves your notes as plain Markdown files under `~/Library/Application Support/stenoai/output/` (macOS) or `%APPDATA%\stenoai\output\` (Windows). Because they are ordinary files, you can move them into Notion, Obsidian, or any other tool by hand or through your own automation.
 </Accordion>
 
 <Accordion title="How do I back up or delete my data?">
@@ -64,7 +64,7 @@ Steno's Windows alpha builds are not digitally signed yet. Windows SmartScreen f
 </Accordion>
 
 <Accordion title="How much disk space does Steno need?">
-The default setup uses roughly 7GB: the Parakeet transcription model (~572MB) plus the default Gemma summarization model. Plan for around 8GB free during install. Larger summarization models take more space. Recordings accumulate over time in `~/Library/Application Support/stenoai/recordings/`.
+The default setup uses roughly 7GB: the Parakeet transcription model (~572MB) plus the default Gemma summarization model. Plan for around 8GB free during install. Larger summarization models take more space. Recordings accumulate over time in `~/Library/Application Support/stenoai/recordings/` (macOS) or `%APPDATA%\stenoai\recordings\` (Windows).
 </Accordion>
 
 <Accordion title="How do I update Steno?">
--- a/docs/faq.mdx
+++ b/docs/faq.mdx
@@ -22,7 +22,7 @@ Steno saves your notes as plain Markdown files under `~/Library/Application Supp
 </Accordion>
 
 <Accordion title="How do I back up or delete my data?">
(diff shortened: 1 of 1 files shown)


Optic00 (commented):
First, a correction, because you did this on the strength of something I got wrong.

I told you the unsigned installer was undocumented - "nothing says so". That is not true, and
I should have checked before saying it. The README has covered it since the Windows alpha
landed in June, in three places:

> On first launch, Windows SmartScreen warns because the alpha is unsigned - click **More
> info → Run anyway**.

plus a "Unsigned - SmartScreen warns on first launch; we'll code-sign before 1.0" bullet and
a note that auto-updates are unsigned too. I generalised from one file to the whole project
without grepping for it. Sorry for sending you after something that was already there.

What was actually missing is narrower, and your PR still fixes it: none of that was in
`docs/faq.mdx`, which is where someone hitting the warning would look. So the new accordion
earns its place - it just is not filling the gap I described.

**The `%APPDATA%` half was right**, and it is the more useful of the two: before this PR the path appeared
nowhere in the README or docs, and `%APPDATA%\stenoai\` matches what `get_user_data_dir()`
in `src/config.py` resolves to on Windows (Roaming). Good catch adding it to backup,
uninstall and storage rather than only one of them.

Two things I would still change.

**Two path mentions were missed**, both still macOS-only after this PR:

- line 21: "Steno saves your notes as plain Markdown files under
  `~/Library/Application Support/stenoai/output/`" - this is the one a Windows user reading
  about exporting notes lands on
- line 67: "Recordings accumulate over time in
  `~/Library/Application Support/stenoai/recordings/`" in the disk-space answer

Same treatment as the three you did would finish it.

**"The app is safe to run" is doing less work than the sentence before it.** "Not digitally
signed yet ... SmartScreen flags unsigned executables" already explains the warning, which is
what makes it safe to proceed. Asking for trust after explaining the mechanism reads weaker
than the explanation alone. I would just drop that clause and keep the code-signing promise.

One housekeeping note: the PR description lists README changes ("Added `#### Installing on
Windows` ... with download, SmartScreen bypass, and data location"), but the diff only
contains `docs/faq.mdx`. I assume the README commit turned out to be a no-op against what was
already there. Worth trimming the description so the next reader is not looking for a change
that is not in the diff.

Optic00 (approved):
Both points from my earlier review are covered now: the export-notes answer (line 21) and
the disk-space answer (line 67) both carry the Windows path, so every path mention in
`faq.mdx` is dual-platform. The uninstall answer picking up **Settings -> Apps** was not
something I asked for and it is the right addition.

Checked rather than assumed: `%APPDATA%\stenoai\` matches what `get_user_data_dir()`
resolves to on Windows (`src/config.py:97`, Roaming), and the `output/`, `recordings/`
subpaths line up with `get_data_dirs()`.

Thanks for the two rounds on this one, and for taking the correction about the README in
good grace. Merging.

Issue #234, opened 2026-06-21
Undo / soft-delete for note deletion (recover an accidental delete before the hard unlink)


ruzin:
## Summary

Note deletion is immediate and **irreversible**: `useDeleteMeeting` → the `delete-meeting` IPC does a hard `fs.unlink` of the audio + summary, with no recovery path. #204 (closes #189) made the *feedback* instant (optimistic cache drop, no backend re-scan), but deliberately kept **no snapshot/rollback** so concurrent deletes can't clobber each other. That leaves the safety gap called out in the #189 thread as the real headline: a mis-click permanently destroys the original recording.

Add an **undo / soft-delete** so users get both the instant feedback #204 ships *and* a safety net for the irreversible audio loss.

## Proposed shape

- On delete, **don't hard-unlink immediately**. Either:
  - move the audio + `*_summary.{md,json}` (+ any `*_notes.txt`/transcript) to a trash subdir under the data root, or
  - defer the unlink and keep an in-memory snapshot of the removed row.
- Show an **"Undo"** affordance (toast/snackbar) for ~5–10s after delete. Undo restores the row (re-insert into the list cache; move files back if trashed).
- **Hard-delete on expiry** (timer elapses, or app quit), so nothing lingers indefinitely.
- Keep #204's instant-removal feel: the row still disappears from the list immediately; undo re-inserts it.

## Notes / constraints

- The list cache already supports targeted writes (`setQueryData` on `meetingsKeys.list()`), so re-inserting on undo is cheap and consistent with #204's approach.
- Cross-platform: a trash dir is simplest and survives an app crash mid-window (an in-memory snapshot is lost on crash). Resolve the trash dir from the same data root as recordings/output (`getUserDataDir()` packaged; mind the dev/prod data-dir split tracked in #233).
- Coverage: add a model-free T2 e2e (delete → undo restores the note + files; delete → wait past window → files gone) per the standing e2e rule.

## Context

Split out from #189 (perf: skip the backend re-scan on delete), which shipped in #204. In that thread the owner asked for undo to get its own tracked issue — this is it.

PR #469, opened 2026-08-02
feat(obsidian): one-way Obsidian vault sync (Settings → Integrations)

### LINKED ISSUE #413: Obsidian integration: export / sync notes into a vault folder
ruzin flagged Obsidian (and Zapier, tracked separately) as the integrations that would "push Steno to another category". This issue scopes the Obsidian half.

## Why it fits well

Steno's notes are already Markdown files on disk - Obsidian's native format. There is no API to integrate against: an Obsidian vault is just a folder. That makes the MVP a thin, offline, privacy-preserving slice - no cloud, no new network surface, consistent with the on-device story.

## Proposed MVP

A "Sync to Obsidian" setting (off by default): pick a folder inside your vault; on note save/update, Steno writes/updates a Markdown copy there.

- One file per note, stable filename derived from the existing note stem so updates overwrite rather than duplicate.
- Frontmatter with date/title/folder (Obsidian-friendly properties), then the note body as it exists today.
- One-way push (Steno → vault). Edits made in Obsidian are NOT synced back and are overwritten on the next update from Steno - documented clearly. Two-way merge is a much bigger project and out of scope.
- Deletions: when a note is deleted in Steno, the vault copy is removed too (respecting the soft-delete/tombstone semantics from #234 once merged).

## Open questions

- Export scope: all notes vs. per-folder selection?
- Should transcripts be included (larger files, arguably the vault wants the notes only)?
- Filename/frontmatter conventions - worth matching what Granola-to-Obsidian users already expect?

## Non-goals (this issue)

Two-way sync, Obsidian plugin, wiki-link generation, backfilling edit history.

ruzin:
Adds **one-way Obsidian vault sync** — mirror your Steno notes into an Obsidian vault folder as Markdown. Off by default; opt in under **Settings → Integrations**. Closes #413.

## What it does
- **Sync to Obsidian** toggle + vault-folder picker (new **Integrations** settings tab).
- On enable, **backfills** all notes; thereafter mirrors every note save/update/delete.
- Notes only (no transcripts). Readable filenames `YYYY-MM-DD Title.md` under subfolders mirroring your Steno folders, with Obsidian-friendly frontmatter (`title`, `date`, `folder`, `participants`, `source: Steno`, `steno_stem`).
- **One-way, never clobbers your edits:** a vault file edited in Obsidian is preserved and flagged as a conflict (surfaced in the tab) rather than overwritten — on update *and* on delete.
- **Title/folder changes rename** the vault file via a stem-keyed index (`.obsidian-sync-state.json`) — no duplicates, no orphans.

## Design
Implemented in Electron main (new `app/obsidian-sync.js` engine + `app/obsidian-ipc.js`), hooked into the note lifecycle the same way enterprise auto-backup and soft-delete (#234) already are — not the one-shot Python CLI. Disk is the source of truth (hooks pass only a path; the engine re-reads and reconciles). **Never throws into a caller** — a failing vault write (unplugged drive, cloud folder, read-only) can't break a note save or a delete commit. Config via `src/config.py` + `simple_recorder.py` (default off), mirroring the `storage_path` pattern.

Hooked write paths: pipeline save, reprocess/generate-notes/retranscribe, the *silent* `update-meeting` (title/notes edit) and folder add/remove (via a new `folders-ipc` `onNoteFoldersChanged` seam), and delete-mirror at the single `commitPendingDelete` choke point. Startup loads cached config + reconciles index drift (async, yields — never janks launch).

## Testing
- **18 engine unit tests** (`app/obsidian-sync.test.js`, wired into `test:unit`): transform (transcript stripped, participants lifted), readable-filename + collision, rename-no-orphan, external-edit conflict (update + delete), reconcile, **never-clobber-a-pre-existing-file**, **no-mass-delete-on-empty-scan**.
- **Model-free T2 e2e** (`e2e/specs/obsidian-sync.t2.spec.ts`): drives the real app's settings + meeting bridges, asserts on-disk vault state (write / rename / conflict-preserve / delete). Skips gracefully without the backend bundle (CI builds it), per AGENTS.md.
- Gates green: `typecheck:renderer`, `lint:renderer` (0 errors), `test:unit` (316 node + vitest).

## Review
Ran a QA + staff-Engineer review pass over the diff, then an adversarial confirming pass. Fixed all HIGH findings (never-clobber-pre-existing-file; no-mass-delete on a failed scan; a control-byte encoding bug that made the engine file read as binary) and the cheap MEDIUMs (idempotent write / no cloud-sync churn / crash-window heal; rename ENOENT-only; backfill index-race). Confirming review: clean, no regressions.

## Known follow-ups (non-blocking)
- Folder **rename/delete** is the same silent class as add/remove — a subsequent re-sync tracks the new subfolder, but existing copies aren't auto-moved until then (documented; an `onFolderRenamed` seam is the fix).
- No "overwrite anyway" affordance for a flagged conflict yet (the tab shows the count).

## Verify
Runtime verify (full Electron app + PyInstaller backend) couldn't be run in the authoring environment — the core is proven by the unit + adversarial tests; the app-integration proof is the T2 e2e in CI. **Green CI is not the same as "watched it run"**, so a manual smoke (enable → pick a throwaway vault → confirm mirror/rename/conflict/delete) is worth doing before merge.


ruzin:
@Optic00 ready for review 🙏 — all checks green (incl. Windows T2).

**What it is:** one-way Steno → Obsidian-vault Markdown sync, off by default, under Settings → Integrations. Closes #413.

**Where to focus your review:**
- `app/obsidian-sync.js` — the engine (the meat). Disk-as-source-of-truth; readable filenames + a stem-keyed rename index (`.obsidian-sync-state.json`); external-edit conflict detection (never clobbers — on update *and* delete); backfill; reconcile; **never throws into a caller** (a vault write can't break a note save or a delete commit).
- Hook placement in `app/main.js` — every note-write path incl. the two "silent" ones (`update-meeting`, folder add/remove via the new `folders-ipc onNoteFoldersChanged` seam) and delete-mirror at the `commitPendingDelete` choke point.
- The new `folders-ipc` seam + `obsidian-ipc.js` + the Integrations tab.

**Testing:** 13 engine unit tests + a model-free T2 e2e (`obsidian-sync.t2`) asserting write/rename/conflict/delete on disk.

**Heads-up from the CI journey (all real fixes, not just test patches):**
- A control-byte in the filename regex made the engine read as *binary* in git — fixed (worth a fresh look now that it's diffable text).
- The Windows T2 lane surfaced a genuine file-lock fragility: `fs.renameSync`/`unlinkSync` can be blocked transiently on Windows (AV/indexer). The engine now unlink-then-writes (no rename), retries transient locks, and — the real fix — records a blocked unlink in a `stale` list that's drained on the next sync/reconcile, so a locked file is never permanently orphaned. `drainStale` + the retry helpers are the bits to sanity-check.

**Honest caveat:** runtime `/verify` was blocked in the authoring env (no Electron+PyInstaller there); correctness is proven by the unit + adversarial tests and the T2 e2e in CI — so a manual smoke (enable → throwaway vault → mirror/rename/conflict/delete) before merge is worth it.

**Known non-blocking follow-ups:** folder *rename/delete* is the same silent class as add/remove (needs an `onFolderRenamed` seam); no "overwrite anyway" affordance for a flagged conflict yet.

ruzin:
Thanks @cubic-dev-ai — strong review. Went through all 16; addressed in e52a135. Summary:

**P1 — all fixed**
- `obsidian-sync.js` **path traversal** (`..` folder & malformed `date`): `sanitizeFilename` now strips leading/trailing dots and never returns `.`/`..`/`''`; only a strict `YYYY-MM-DD` date enters a filename. Added a `..`-folder unit test.
- `reconcile` **partial-scan mass-delete**: `listSummaryFiles` returns a `complete` flag; deletes are trusted only when every base dir read cleanly **and** the scan found notes.
- `reconcile` **stale-but-existing dest not re-synced**: now re-syncs *every* present note (hash-aware idempotent skip), which also heals folder-rename subfolder drift.
- `obsidian-ipc` **vault switch mid-backfill**: `backfillAll` queues a re-run so the newly-selected vault is fully populated.
- `deriveFilename` **second collision**: keeps probing suffixed candidates until free.

**P2 — fixed**
- Reprocess mirrors under the canonical `realPath`, not the renderer alias.
- Reconcile delete now **records** an external-edit conflict (was silently skipped) and stashes a lock-blocked unlink in `stale` instead of dropping it.
- `useSettings`: `setObsidianSync` no longer wraps `mutate` (consistent with `mutateAsync`); the conflicts query polls (`refetchInterval`) so conflicts recorded by the async backfill surface in the tab.

**P3 — fixed**
- `Settings.tsx`: added `integrations` to `DEEP_LINK_IDS` (+ comment) — the tab was falling back to General (confirmed live). Good catch.
- `IntegrationsTab`: `aria-label` on the Switch.
- Test renamed + a real `backfillAll()` test added.

**Pushing back on one (P3, e2e re-nudge):** the rename `expect.poll` re-nudges a sync on purpose. On the Windows runner a file-lock can delay the old-file unlink past a single sync; the re-nudge is what drives the `stale`-drain retry until it clears. A one-shot re-sync + read-only poll would reintroduce the Windows flake we chased down. Kept intentionally; happy to add a comment making that explicit.

Suite 319/319; typecheck + lint clean. Also smoke-tested live in the app.

PR #470, opened 2026-08-02
feat(overview): show which notes still have their original audio

Optic00:
`keep_recordings` defaults off, so for most notes the source recording is gone. Everything that needs it is then quietly unavailable - re-transcribe, speaker listening samples, any future re-diarization - and nothing said so until you opened a note and found the action missing.

`list-meetings` now returns `has_audio`, and the overview row shows a small audio glyph beside the duration when it is true. Only when true: absence is the normal case, so a "missing" marker on most rows would be noise rather than information.

### Two implementation notes

Derived from **one** directory listing of `recordings/`, not a per-meeting existence check. This command is on the cold-start path and its own docstring calls it *optimized for fast loading*, so the cost must not scale with library size. A regression test counts the listings to hold it there.

Matching is on the stem alone, never an extension whitelist: the capture pipeline saves whatever the source produced (`.webm`, `.wav`, `.m4a`, `.mp3`), so a whitelist would silently mark imported meetings audio-less.

Two commits: the feature, and a follow-up fixing two stem-matching edge cases a review turned up - a directory named like a recording, and a note whose own name contains `_summary`.

### Verified

7 new backend tests, 1 new T1 spec, `typecheck:renderer` clean, `ruff` 29 - identical to `main`.

### DIFF
--- a/app/e2e-mock-ipc.js
+++ b/app/e2e-mock-ipc.js
@@ -165,6 +165,35 @@ function seedPriorSegments(rec) {
   ];
 }
 
+// Two notes differing ONLY in whether their original recording still exists,
+// for the overview's audio indicator (STENOAI_E2E_SEED_AUDIO_MEETINGS=1).
+// keep_recordings defaults off, so "no audio" is the normal case -- the pair
+// is what proves the icon tracks the flag rather than always rendering.
+const AUDIO_SEED_MEETINGS = [
+  {
+    session_info: {
+      name: 'With audio',
+      summary_file: 'with-audio_summary.md',
+      processed_at: '2026-08-02T12:00:00Z',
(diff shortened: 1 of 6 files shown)


Optic00:
Heads-up so this doesn't arrive silently: I merged these two commits into #473 (the speaker-review stack). I needed the audio indicator while testing that branch against real recordings, and one build was easier to reason about than two.

Consequence: if #473 lands on `feat/speaker-diarization` first, this feature comes along with it without this PR ever being reviewed. Git handles the overlap either way — same commits, common ancestor.

Either way works for me. Leave this open for its own review and I'll drop the merge from #473, or review it here and I'll close this once #473 lands. Your call.

PR #472, opened 2026-08-03
feat(speakers): human-in-the-loop cross-recording speaker identification

valentinweyer:
Part 2 of #359 (see that issue for full background/testing status). Builds on #455 (per-channel acoustic diarization) with human-confirmed cross-recording speaker identification.

**Note on the diff:** this branch already carries #455's diarization commits internally (re-applied while this branch was rebuilt), and `feat/speaker-diarization` doesn't have #455 merged yet, so this diff currently shows both. Once #455 merges, re-diffing/rebasing will shrink this down to just what's below.

## What's new here (on top of #455)

- **`PersonProfile`/`SpeakerPrototype` CRUD** (`src/config.py`), with hard-negative and context-aware (in-person vs. remote) evidence.
- **Suggestion engine** (`src/speaker_suggestions.py`): gates a candidate match on threshold, margin, stability across a meeting's segments, and hard-negative evidence from prior rejections, plus same-meeting fragment merging, transcript relabeling, and sample-audio/text extraction so a user can hear/read a clip before confirming.
- **Approval UI** (`SpeakerReviewPanel`, wired into the meeting detail view): approve / change / new-person / keep-generic, with duplicate-name prevention.
- **14 new CLI commands** for the full lifecycle (create/rename/delete person profiles, suggest-speakers, confirm-speaker, backfill-speaker-embeddings, speaker-suggestion-report, get-speaker-sample-audio, enroll-self-from-person, and more), plus a new `speakers` IPC group.
- **A setting to disable cross-recording identity matching** entirely, for users who only want per-meeting diarization without persistent voiceprints.
- **Verified deletion**: `delete_person_profile` now also strips the deleted person's voiceprint out of every other profile's mutual hard-negative list (previously it only removed the deleted person's own entry, so their sample kept living on inside everyone else's hard negatives indefinitely).

## Honest caveat

Cross-meeting matching quality isn't great yet — validating raw embedding-similarity against the AMI Meeting Corpus found that people sharing a room/mic score artificially similar regardless of true identity, and no threshold/margin tuning fixed that safely without also rejecting correct matches. This is why it's confirm-first, never a silent auto-assign. Within-meeting speaker separation (telling Speaker 1 from Speaker 2) is the reliable, validated part; cross-recording identity is a bonus when it works, not the thing carrying the feature.

## Testing

- Backend: 811 unittest cases pass (`python -m unittest discover tests`)
- Renderer: `tsc --noEmit` clean, `eslint` clean (no new warnings)
- Electron/JS: all `node --test` + vitest unit tests pass
- e2e: `speaker-diarization.t2`, `speaker-naming.t2`, `speaker-review.t1`

---
## Summary by cubic
Adds human‑confirmed speaker identification across recordings on top of per‑channel acoustic diarization, with a review UI, person profiles, and a macOS diarization/embedding sidecar. Improves processing feedback and adds a setting to disable identity matching.

- **New Features**
  - Speaker review panel (`SpeakerReviewPanel`) to approve/change/new-person/keep-generic, with sample audio preview and duplicate-name prevention.
  - Person profiles (CRUD) with context-aware prototypes and hard‑negatives, plus a suggestion engine that gates by threshold, margin, stability, and conflicts; exposed via a new `speakers` IPC group and CLI (suggest/confirm/backfill/enroll-self/report, etc.).
  - macOS `steno-diarize` sidecar (Swift/CoreML + embeddings) for mono and per‑channel diarization; up to four voices per channel, with transcript relabeling to confirmed names.
  - Processing screen adds an “Identifying speakers” stage and embedding sub‑progress; transcribe progress restored.
  - Setting to disable cross‑recording identity matching (`identity_matching_enabled`), respected by the pipeline and backfill.

- **Bug Fixes**
  - Deleting a person now removes their voiceprint from others’ hard‑negatives; evidence is channel‑scoped and “Change” reassigns prototypes cleanly.
  - `is_diarised` reflects real multi‑speaker output; long sentences split by word timings to avoid mislabels.
  - Windows spawns set `PYTHONUNBUFFERED=1` to keep progress alive; better surfacing of graceful CLI errors; processing-stages test runs with `fakeAudio` on CI and waits for the generation bump.

<sup>Written for commit fab3bbaef1e94f856a3d00479e7362df643bf508. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/stenolabs/stenoai/pull/472?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Optic00:
#455 is merged into `feat/speaker-diarization`, and I've pushed the merge of that branch into this one (`fab3bbae`) so you don't wake up to a conflicted PR and solve it in parallel with me. Revert it without asking if you'd rather do it your own way.

**Why it conflicted at all**, since it looked worse than it was: #455 was rebased onto the shared branch, while this branch was cut from the *pre-rebase* originals. So the same diarization changes arrived twice with different ancestry — 60 conflict hunks across 7 files, all of them that duplication. None of us spotted it when we agreed on the rebase; it's nobody's bug, just a consequence of rebasing a branch someone else was already building on.

**How I resolved it:** every conflicting hunk to *this* branch's side, since your identity work is the superset there (`logPipelineStdoutLine` is the clearest example — yours handles all `PROGRESS:` lines with throttling for the high-frequency ones, the other side only `PROGRESS:diarize:`). The merge then left a second copy of `formatElapsedSeconds` in `Processing.tsx`, which typecheck caught; that's removed again.

The net change against your previous head is only this:

```
 e2e/specs/processing-stages.t1.spec.ts | 11 ++++++++---
```

— the three `fakeAudio: true` calls, plus the settle-wait in `openProcessing`. The diff against the base dropped from +12424/-76 over 52 files to **+10577/-184 over 39**, which is the identity work on its own.

Verified before pushing: 828 Python tests OK, renderer typecheck clean, lint 37 warnings / 0 errors (unchanged), ruff unchanged — the merge touches no Python relative to your head. I did **not** run e2e locally; after last night I'd rather let this CI run be the signal than tell you a green macOS run means something.

Still yours to decide: `speaker-review.t1:177` expects a button labelled "Stop sample" and finds "Play sample". That's a statement about what your feature should do, not test plumbing, so I left it alone.

Optic00:
Merging this into `feat/speaker-diarization` now so the branch stops aging under you — it's the shared integration branch, not `main`, so nothing here reaches users, and the merge is revertible.

One thing said plainly before it disappears into the branch history: **the conflict resolution in `fab3bbae` has had no review but mine.** 60 hunks over 7 files, every one resolved to your side. For duplicate ancestry that is almost always right, but wherever the two sides had genuinely diverged it was my judgement, not yours, and it is much easier to inspect here in the PR diff than after the merge. The places my judgement actually entered:

- `app/main.js`, `logPipelineStdoutLine` — kept your version (all `PROGRESS:` lines with throttling for the high-frequency ones) over the other side's `PROGRESS:diarize:`-only branch.
- `Processing.tsx` — after the merge there were two identical `formatElapsedSeconds` definitions; typecheck caught it, I removed the second.
- `src/transcriber.py` (23 hunks) and `tests/test_transcriber_diarisation.py` (18) — all resolved to your side, no content of mine added.

If any of that reads wrong to you, `git revert` on the merge commit gets the branch back and I'll redo it your way.

Still open and yours: `speaker-review.t1:177`. It's flaky rather than failing — first attempt expects "Stop sample" and finds "Play sample", the retry passes — so it doesn't block anything, but it's both an intent question and an instability.

PR #491, opened 2026-08-24
feat(macOS): add Apple SpeechTranscriber backend

audreyt:
## Problem

Steno currently ships application-managed ASR weights on every platform. macOS 26 provides `SpeechAnalyzer` / `SpeechTranscriber`, a system-managed on-device engine designed for long-form and conversational audio. Using it can remove the first-run transcription-model download on supported Macs while keeping Parakeet and Whisper for other systems and locales.

This PR adds the native backend and packaging foundation only. It does **not** change the fresh-install default or expose a renderer control; that UI/default work is kept in a stacked follow-up.

## Architecture

- New pure-Swift `steno-transcribe` helper with four commands:
  - `status [locale]`
  - `prepare [locale]`
  - `transcribe-file <audio> [locale]`
  - `stream [locale]`
- Batch path: Python invokes `transcribe-file` and receives the existing normalized transcript/segment JSON shape.
- Live path contract: interleaved 16 kHz stereo Float32 enters stdin; the helper runs one SpeechTranscriber lane per channel and emits the existing `LIVE_READY` / `LIVE_SEG` / `LIVE_ERROR` protocol.
- The helper validates `supportedLocale(equivalentTo:)` against the actual supported-locale set, uses `AssetInventory` for system assets, finalizes input explicitly, deduplicates finals, and rejects punctuation-only results.
- PyInstaller places the helper under `_internal`, alongside the existing diarization helper.

## Platform behavior

| Platform | Behavior |
|---|---|
| macOS 26+ | Apple engine is available when explicitly selected |
| macOS 14.4–15 | Existing Parakeet/Whisper behavior remains the default |
| Windows | Existing Parakeet ONNX/Whisper behavior remains unchanged |

The application minimum remains macOS 14.4. The Swift helper itself targets macOS 26 and is invoked only after runtime/platform gating.

## Privacy

Audio stays on the Mac. The helper communicates only through stdin/stdout and system speech APIs. It creates no transcript temp files, sends no telemetry, and exposes no meeting text in errors.

## Verification

- `scripts/build-transcribe-sidecar.sh arm64` — arm64 Mach-O built
- `python -m unittest tests.test_apple_speech tests.test_sidecar_bundle_guard tests.test_config tests.test_transcriber` — 166 passed
- `bin/steno-transcribe status en` — available/supported/installed JSON returned
- `git diff --check upstream/main` — clean

## Follow-up

A stacked draft will add Setup/Settings exposure, fresh-install default selection on macOS 26+, and model-free coverage for switching the live locale during a recording.

---
## Summary by cubic
Adds an Apple SpeechTranscriber backend on macOS 26+ via a Swift `steno-transcribe` sidecar. On supported Macs, users can explicitly select Apple's on-device engine; Parakeet/Whisper behavior and defaults remain unchanged elsewhere.

- Introduces a pure-Swift sidecar with `status`, `prepare`, `transcribe-file`, and `stream` commands. Live mode reads 16 kHz stereo Float32 from stdin, runs one lane per channel, and emits the existing LIVE_READY/LIVE_SEG/LIVE_ERROR protocol. Only `prepare` downloads system assets; `transcribe-file` and `stream` require them already installed.
- Adds a Python bridge (`src/apple_speech.py`) that resolves the sidecar (overridable via `STENOAI_TRANSCRIBE_SIDECAR_PATH`) and normalizes JSON for batch and setup paths.
- Extends backend selection (`src/transcriber.py`): honors `apple`/`parakeet`/`whisper`; an explicit `apple` selection fails closed if the sidecar is missing; no silent fallback from Apple to Parakeet. Keeps shared transcript shape.
- Updates config (`src/config.py`): adds `apple` engine, keeps Parakeet as the default until UI changes, and gates Apple to macOS 26+. If a config set to `apple` is opened on an unsupported host, it resolves to Parakeet without mutating the stored value.
- Bundles the sidecar in `PyInstaller` under `_internal` and replaces the diarization-only guard with a generic macOS sidecar guard. CI builds and verifies the new binary and switches macOS jobs to `macos-26`. The build script only runs on macOS 26+; older hosts skip it and still build a Parakeet/Whisper bundle.
- Adds CLI helpers: `apple-speech-status [language]` and `prepare-apple-speech [language]`. Adds tests for the bridge, guards, selection, and native contract.

**Rollout**
- macOS release builds must run on `macos-26` and include the sidecar: run `scripts/build-transcribe-sidecar.sh` before `PyInstaller` (CI already does this). Builds fail if required sidecars are missing.
- No UI/default change in this PR; macOS users will opt into Apple in a follow-up. System assets are installed via `prepare`; no model downloads occur implicitly.

<sup>Written for commit cc81a7cb14e83d78eb2f31c53dd58656a9e7c2fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/stenolabs/stenoai/pull/491?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Optic00 (changes requested):
The current head still has unresolved correctness and compatibility issues: runtime asset downloads, silent live-audio buffer loss, missing macOS 26/helper gating, older-mac build failures, config and no-backend fallback errors, malformed heartbeat labels, mirror-only lexical tests, and missing sidecar executable and architecture validation. Please also rebase without reverting the interrupted-download fix and its tests already on main.

PR #505, opened 2026-08-28
feat(export): add readable transcript bundle

audreyt:
## Description

Adds a readable conversation section to transcript exports while retaining the original timestamped transcript.

Final ASR can split one utterance across several adjacent segments. The export now merges only consecutive segments with the same channel label when all three bounds hold:

- no more than 2.5 seconds between fragments
- no more than 20 seconds across the combined paragraph
- no more than 400 characters after merging

`You` renders as `Me`; other unconfirmed channel or speaker labels remain unchanged. Whenever merging changes the readable view, `## Timestamped transcript` preserves every original segment and timestamp.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [ ] Tested locally with `npm start`
- [x] Tested on macOS

Verified on current `upstream/main`:

- `bun run typecheck:renderer`
- `bun run lint:renderer` — existing warnings only; no new errors
- `bun run test:unit` — 338 Node tests and 203 renderer tests passed
- `bun run build:renderer`

## Additional Notes

The readable view never infers a named identity. The existing timestamped source remains intact and is included whenever coalescing occurs.

### DIFF
--- a/app/renderer/src/lib/transcriptBundle.test.ts
+++ b/app/renderer/src/lib/transcriptBundle.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, test } from 'vitest';
+import type { Meeting } from '@/lib/ipc';
+import { buildTranscriptBundle } from '@/lib/transcriptBundle';
+
+function meeting(overrides: Partial<Meeting> & { diarised_text?: string }): Meeting {
+  return {
+    session_info: {
+      name: 'Project sync',
+      duration_seconds: 45,
+      processed_at: '2026-01-15T12:00:00.000Z',
+    },
+    summary: '',
+    is_diarised: true,
+    transcript: '',
+    ...overrides,
+  } as Meeting;
+}
+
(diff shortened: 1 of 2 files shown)


Optic00 (changes requested):
Requesting changes because the readable export is incorrect for two stored transcript states:

- If a resumed recording restarts timestamps, a negative delta currently satisfies `<= 2.5` and merges the resumed fragment into the prior session. Require a non-negative gap.
- Older diarised transcripts may have no timestamps. They currently become one repeated speaker row per fragment and also get a `Timestamped transcript` section containing no timestamps, duplicating the source. Fall back to the original single transcript section unless usable timestamps exist.

The character bound should measure normalized fragment text, matching the whitespace collapse performed by the merge. Please also cover the exact and just-over boundaries for the 2.5-second gap, 20-second span, and 400-character cap. Current CI is green but does not exercise these states.

PR #519, opened 2026-09-01
fix: reconcile duplicate speakers across bleed channels

Hendrikc4:
## What users see

Steno records calls on two audio tracks:

- the microphone track, normally the person using the Mac;
- the system-audio track, normally everyone speaking through Zoom, Google Meet, and similar apps.

When someone uses laptop speakers instead of headphones, the microphone can also hear the voices coming from the computer.
Steno already removes most repeated words from the final transcript, but speaker identification still analyzes both original tracks.

The same real person can therefore be detected twice: once on the clean system-audio track and once through the quieter echo on the microphone track.
A transcript may switch between labels such as `Speaker 2` and `Speaker 5` even though both labels belong to the same person, and the Speakers panel can show duplicate people.

The original report reproduced this on a 4:55 meeting in Steno v0.7.0.
Steno reported three speakers on each track, with three clear cross-channel pairs at cosine distances 0.088, 0.065, and 0.092; every next-best match was at least 0.522 away.

No meeting content, names, audio, or voice embeddings are included in this PR.

## Why it happens

Text-level bleed correction runs before speaker identification and removes duplicate transcript segments.
Speaker identification then analyzes each complete channel WAV independently, including audio that produced removed segments.
Sortformer can therefore create two channel-local speaker IDs for the same leaked voice, and placeholder resolution treats those IDs as different people.

## What this changes

- Compares the per-meeting voiceprints produced by both diarization passes when identity matching is enabled.
- Merges only mutual-nearest mic/system pairs below the existing strict same-meeting distance threshold.
- Requires a clear runner-up margin on both sides.
- Never merges the dominant `You` and `Others` pair, and rejects any merge set that would erase an existing speaker split.
- Keeps the channel containing more post-bleed transcript text as the canonical copy.
- Updates transcript provenance and the Speakers panel to the same canonical speaker ID.
- Preserves the union of both clusters' time ranges for review clips without double-counting overlapping duration.
- Performs no cross-channel embedding comparison when identity matching is disabled.

## Result

Clear non-dominant echo pairs collapse to one canonical speaker.
Ambiguous pairs, the dominant `You`/`Others` pair, and meetings processed with identity matching disabled remain unchanged.
Transcript text is never deleted by reconciliation.

## Reviewer notes

- **No additional model run:** reconciliation reuses WeSpeaker embeddings already produced by the diarization sidecar.
- **Fails closed:** a pair must be mutual-nearest, be at or below the existing `0.10` same-meeting threshold, beat the runner-up by at least `0.05` on both sides, and pass the speaker-split guards.
- **Platform scope:** the path is active only when both stereo channels produced acoustic clusters and identity matching is enabled. Mono recordings and Windows/Linux fallback behavior are unchanged.
- **Privacy behavior:** when identity matching is disabled, this path neither collects clusters nor compares or persists embeddings.

## Validation

- `python -m unittest tests.test_transcriber_diarisation`: 151 passed
- `ruff check src/transcriber.py tests/test_transcriber_diarisation.py`: passed
- Full GitHub CI: passed on macOS, Windows, and Linux, including both T2 paths
- Independent architecture and code review: no remaining findings after two correction rounds

### DIFF
--- a/src/transcriber.py
+++ b/src/transcriber.py
@@ -45,6 +45,7 @@ from typing import Callable, Optional, Tuple
 
 from src._heartbeat import _emit_heartbeat
 from src.speaker_suggestions import (
+    SAME_MEETING_MERGE_DISTANCE_THRESHOLD,
     SUGGESTION_MIN_AVG_TURN_SECONDS,
     build_clusters_from_diarization,
     determine_recording_type,
@@ -129,6 +130,14 @@ RMS_MAX_WINDOWS = 60
 # misattribution in _assign_asr_segments_to_diar_segments. Matches the value
 # validated against real meeting audio in the research playground.
 STENO_DIARIZE_MERGE_GAP_S = 0.3
+
+# Cross-channel echo can make the same voice appear in both independent
+# diarization runs. Only merge mutual-nearest embedding pairs that are as
+# tight as the already-validated same-meeting fragment threshold, and whose
+# runner-up is clearly worse. The margin prevents two acoustically similar
+# speakers from being joined merely because both fall below the distance
+# cutoff.
+CROSS_CHANNEL_SPEAKER_MATCH_MARGIN = 0.05
 
 # Floor for the steno-diarize subprocess timeout. Real measured runtime
 # varies a lot with recording length: single-digit-to-tens-of-seconds for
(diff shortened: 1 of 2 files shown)


Optic00 (approved):
Reviewed the matching, fail-closed guards, identity opt-in, provenance rewrite, and regression coverage. The focused 151-test suite and Ruff pass locally, the full CI matrix is green, and an independent architecture and code review found no remaining issues.

PR #521, opened 2026-09-01
fix: recover quiet microphone speech in stereo recordings

Hendrikc4:
## The problem

In some Zoom and Google Meet recordings, Steno captures the local microphone much more quietly than the other people in the call.

The voice is present in the recording, but Steno sends the quiet audio directly to the transcription model. The model can then miss most of what the local user said, leaving the note with very few `[You]` turns.

The author reproduced this with a 30-minute recording:

- Before the fix: 7 `[You]` turns and 136 characters.
- After the fix: 61 `[You]` turns and 5,105 characters.

The microphone was about 10.6 dB quieter than the call audio. No private transcript text or recording data is included in this PR.

## The fix

Steno already has a step that adjusts quiet audio before transcription. This PR makes sure that step also runs for recordings with separate microphone and system-audio channels.

Steno still keeps the original channel volumes when it checks for echo and audio bleed. Only the copies sent to the transcription model are adjusted.

The preprocessing timeout now scales with recording duration, stereo splitting reports heartbeat progress, and the stereo path avoids applying the high-pass filter twice.

## Testing

- 193 focused audio, speaker, timeout, and integration tests pass locally.
- The backend CI suite passes: 1,271 tests, with 1 skipped.
- The complete CI matrix passes on macOS, Windows, and Linux, including packaged-backend builds and real-backend T2 pipelines.
- The author reports that a complete reprocess of the 30-minute recording finished successfully and recovered the missing local-speaker turns.

### DIFF
--- a/src/transcriber.py
+++ b/src/transcriber.py
@@ -296,6 +296,13 @@ def _parse_duration_from_ffmpeg_stderr(stderr: str) -> Optional[float]:
     return int(m.group(1)) * 3600 + int(m.group(2)) * 60 + float(m.group(3))
 
 
+def _duration_scaled_audio_timeout(duration_seconds: Optional[float], floor_s: int) -> int:
+    """Return a generous bounded timeout for a full-length audio pass."""
+    if duration_seconds and duration_seconds > 0:
+        return max(floor_s, int(duration_seconds * 2))
+    return floor_s
+
+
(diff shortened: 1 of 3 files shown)


Optic00 (approved):
Reviewed the stereo preprocessing, timeout, heartbeat, cleanup, and bleed-correction paths. The focused local suite and complete cross-platform CI matrix pass, and all review threads are resolved.

